PR #283, opened 2015-04-27
Added menu and ROM support to Brazilian clone "TK3000 //e".

wmoecke:
Added menu and ROM support to Brazilian clone "TK3000 //e".
I also included resources to future support Brazilian "Timemaster II H.O." clone board, "TKClock".


sicklittlemonkey:
I'm not looking into this work, but just to speed up the turnaround you
have a whole lot of files in that commit that should be excluded - obj,
tlog, pdb etc.

Cheers,
Nick.

wmoecke:
Thanks for looking into that, I will see to it that these are no longer included. I had the impression that the .gitignore file would take care of it.
I shall repost this commit once I've eliminated the garbage.

Issue #260, opened 2015-01-11
Save-state file v2


tomcw:
Spec:
* Support current .aws version (v1)
* New version (v2):
  * Change `SS_UNIT_HDR` to include a "type" field (this was the main extensibility limitation with v1)
  * Support adding new cards in future (ie. code should skip `SS_CARDTYPE`'s that are currently unsupported)
  * Support cards in any slot
  * Support restarting emulated system if (and only if) h/w config changes
  * Support current missing cards: eg. mouse, phasor, HDD, CP/M, Extended memory (RAMworks) etc

Update documentation to reflect this improved save-state (as currently it lists all the limitations, eg. no mouse etc).

Branch "savestate" will be used for development.

NB. This issue covers #255.

tomcw:
Work completed in  #315 / commit c798157. Closing.

Issue #254, opened 2014-12-15
(Graphics) HGR COLORs are not accurate


Michaelangel007:
The colors for the Hi-Resolution graphics screen are OK but could use improvement.

Running this AppleSoft Basic program on actual hardware and on AppleWin ...

```
10 HGR
20 FOR Y=0 TO 63:FOR X=0 TO 7
30 HCOLOR=INT(Y/8)
31 HPLOT X*32,Y*2 TO X*32+30,Y*2
32 HCOLOR=X
33 HPLOT X*32,Y*2+1 TO X*32+30,Y*2+1
40 NEXT:NEXT
50 ? "0=BLACK1  4=BLACK2"
60 ? "1=L.GREEN 5=ORANGE"
70 ? "2=PURPLE  6=MED.BLUE"
80 ? "3=WHITE1  7=WHITE2  8x8 HGR COLOR CHART";
90 GET A$:END
```

![apple_hgr_level_correct](https://cloud.githubusercontent.com/assets/7876796/5442853/a9fc639e-8451-11e4-882a-f6cf926bcdc8.png)

---
- Standard

![hgr_linards_tweaked_standard](https://cloud.githubusercontent.com/assets/7876796/5442877/d5c6f67e-8451-11e4-9ef2-f3fa1bd764e6.png)

---
- TV

![hgr_linards_tv](https://cloud.githubusercontent.com/assets/7876796/5442901/0cea3c92-8452-11e4-80e8-c7dea545f2cc.png)

Michaelangel007:
Rainbow Reference
https://groups.google.com/forum/m/#!topic/comp.sys.apple2.programmer/C70j36jSKvQ

    10 A = 9200 :B = 13168:C=100:X=279
    20  TEXT : HOME :GOTO 40
    30 HPLOT 0,Y TO X,Y:RETURN
    40 DATA 173,87,192,173,83,192,173,84,192,173,80,192
    50 DATA 208,251,173,86,192,160,22,136,208,253,234,173,87,192,76,9,3
    60  FOR I = 768 TO 796:READ N:POKE I,N:NEXT
    70 HGR : POKE  - 16302,0 
    80  HCOLOR= 4:HPLOT 0,0: CALL 62454 
    90 HCOLOR= 2:Y=63:GOSUB 30:Y=62:GOSUB 30
    100 HCOLOR= 6:Y=61:GOSUB 30:Y=60:GOSUB 30
[...]

PR #304, opened 2015-08-21
Save-state v2

tomcw:
Hi - a mega pull-request, I'm afraid!

The short summary is that it addresses #260, with this new format being extensible and supporting all the AppleWin cards (except Uthernet). Additionally the old v1 format is still supported (for loading).

I don't think it's worth anyone actually reviewing this pull-request. Maybe just give an opinion on the restoring of the configuration (see below).

This v2 format also saves/restores the configuration state. I'm in two minds about restoring the config though:
. On the one hand, you can save a game's state: eg. where you want to use cursor keys (or auto-fire) etc. for joystick emulation.
. But on the other hand, you don't know what state you'll be loading in. So it'll overwrite your current setup (eg. Apple II type, video mode, ConfirmReboot, ScrollLock toggle mode, volume settings, etc)
  - arguably you don't know what h/w state you are loading in anyway, so your current card setup will switch to the newly loaded state's

So a quick solution is to just disable the restore config state routine for now.
What do you think?

In terms of testing, I've got save-state files that test all cards. My plan is to put these into a separate git repo (eg. AppleWin-test). The reason for this is that they are binary files and if they get updated then all versions will accrue in the repo. OK, the size is only ~140K (or bigger for RAMworks saved state files), so perhaps having lots of these to download when you sync with a repo isn't such a big deal.

OTOH, the advantage of putting them directly into the AppleWin repo is that they'll be in sync with the commits to that repo (making testing specific versions easier).

Anyway, I think if it doesn't work having them in a separate git repo, then I can just move them across.

### DIFF
--- a/AppleWinExpress2012.vcxproj
+++ b/AppleWinExpress2012.vcxproj
@@ -59,6 +59,7 @@
     <ClInclude Include="source\Debugger\Debugger_Range.h" />
     <ClInclude Include="source\Debugger\Debugger_Symbols.h" />
     <ClInclude Include="source\Debugger\Debugger_Types.h" />
+    <ClInclude Include="source\Debugger\Util_MemoryTextFile.h" />
     <ClInclude Include="source\Debugger\Util_Text.h" />
     <ClInclude Include="source\Disk.h" />
     <ClInclude Include="source\DiskDefs.h" />
@@ -74,6 +75,7 @@
     <ClInclude Include="source\MouseInterface.h" />
     <ClInclude Include="source\NoSlotClock.h" />
     <ClInclude Include="source\ParallelPrinter.h" />
+    <ClInclude Include="source\Pravets.h" />
     <ClInclude Include="source\Registry.h" />
     <ClInclude Include="source\Riff.h" />
     <ClInclude Include="source\SaveState.h" />
@@ -94,7 +96,6 @@
     <ClInclude Include="source\Tfe\Tfearch.h" />
     <ClInclude Include="source\Tfe\Tfesupp.h" />
     <ClInclude Include="source\Tfe\Uilib.h" />
-    <ClInclude Include="source\Debugger\Util_MemoryTextFile.h" />
     <ClInclude Include="source\Video.h" />
     <ClInclude Include="source\z80emu.h" />
     <ClInclude Include="source\Z80VICE\daa.h" />
@@ -136,6 +137,7 @@
     <ClCompile Include="source\Debugger\Debugger_Parser.cpp" />
     <ClCompile Include="source\Debugger\Debugger_Range.cpp" />
     <ClCompile Include="source\Debugger\Debugger_Symbols.cpp" />
+    <ClCompile Include="source\Debugger\Util_MemoryTextFile.cpp" />
     <ClCompile Include="source\Disk.cpp" />
     <ClCompile Include="source\DiskImage.cpp" />
     <ClCompile Include="source\DiskImageHelper.cpp" />
@@ -149,6 +151,7 @@
     <ClCompile Include="source\MouseInterface.cpp" />
     <ClCompile Include="source\NoSlotClock.cpp" />
     <ClCompile Include="source\ParallelPrinter.cpp" />
+    <ClCompile Include="source\Pravets.cpp" />
     <ClCompile Include="source\Registry.cpp" />
     <ClCompile Include="source\Riff.cpp" />
     <ClCompile Include="source\SaveState.cpp" />
@@ -187,7 +190,6 @@
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">NotUsing</PrecompiledHeader>
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Release NoDX|Win32'">NotUsing</PrecompiledHeader>
     </ClCompile>
-    <ClCompile Include="source\Debugger\Util_MemoryTextFile.cpp" />
     <ClCompile Include="source\Video.cpp" />
     <ClCompile Include="source\z80emu.cpp" />
     <ClCompile Include="source\Z80VICE\daa.cpp">
--- a/AppleWinExpress2012.vcxproj.filters
+++ b/AppleWinExpress2012.vcxproj.filters
@@ -30,6 +30,9 @@
     <Filter Include="Source Files\Emulator">
       <UniqueIdentifier>{9b13bfc1-31ab-4f55-bb69-b8620ebdc2be}</UniqueIdentifier>
     </Filter>
+    <Filter Include="Source Files\Model">
+      <UniqueIdentifier>{15b450e4-f89f-4d80-9c44-48b32f33f3e3}</UniqueIdentifier>
+    </Filter>
     <Filter Include="Source Files\Uthernet">
       <UniqueIdentifier>{9bcc8097-f610-4843-bd76-a313aa54fbb0}</UniqueIdentifier>
     </Filter>
@@ -291,6 +294,9 @@
     <ClInclude Include="source\Debugger\Util_Text.h">
       <Filter>Source Files\Debugger</Filter>
     </ClInclude>
+    <ClInclude Include="source\Pravets.h">
+      <Filter>Source Files\Model</Filter>
+    </ClInclude>
   </ItemGroup>
   <ItemGroup>
     <Text Include="docs\CodingConventions.txt">
@@ -483,6 +489,9 @@
     </ClCompile>
     <ClCompile Include="source\Tfe\Uilib.cpp">
       <Filter>Source Files\Uthernet</Filter>
+    </ClCompile>
+    <ClCompile Include="source\Pravets.cpp">
+      <Filter>Source Files\Model</Filter>
     </ClCompile>
   </ItemGroup>
   <ItemGroup>
--- a/AppleWinExpress2013.vcxproj
+++ b/AppleWinExpress2013.vcxproj
@@ -75,6 +75,7 @@
     <ClInclude Include="source\MouseInterface.h" />
     <ClInclude Include="source\NoSlotClock.h" />
     <ClInclude Include="source\ParallelPrinter.h" />
+    <ClInclude Include="source\Pravets.h" />
     <ClInclude Include="source\Registry.h" />
     <ClInclude Include="source\Riff.h" />
     <ClInclude Include="source\SaveState.h" />
@@ -106,7 +107,7 @@
     <Text Include="docs\CodingConventions.txt" />
     <Text Include="docs\Debugger_Changelog.txt" />
     <Text Include="docs\FAQ.txt" />
-    <Text Include="docs\History.txt" />
+    <Text Include="bin\History.txt" />
     <Text Include="docs\ToDo.txt" />
     <Text Include="docs\Video_Cleanup.txt" />
     <Text Include="docs\Wishlist.txt" />
@@ -150,6 +151,7 @@
     <ClCompile Include="source\MouseInterface.cpp" />
     <ClCompile Include="source\NoSlotClock.cpp" />
     <ClCompile Include="source\ParallelPrinter.cpp" />
+    <ClCompile Include="source\Pravets.cpp" />
     <ClCompile Include="source\Registry.cpp" />
     <ClCompile Include="source\Riff.cpp" />
     <ClCompile Include="source\SaveState.cpp" />
--- a/AppleWinExpress2013.vcxproj.filters
+++ b/AppleWinExpress2013.vcxproj.filters
@@ -169,6 +169,9 @@
     <ClCompile Include="source\Debugger\Util_MemoryTextFile.cpp">
       <Filter>Source Files\Debugger</Filter>
     </ClCompile>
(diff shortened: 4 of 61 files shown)


tomcw:
FTR, I took a quick look at some other emulators and whether they saved any configuration state in the save-state file:

1. VICE (C64)
   * Just saves emulated machine h/w state.
    * Documentation is pretty comprehensive about what gets saved.
2. WinUAE (Amiga)
   * Appears to just saves emulated machine h/w state.
   * Documentation is poor (just 1 line!) so it's not clear at all what is really saved.
   * Has the concept of configuration files to define specific machine setups, and additionally true configuration things like key mappings etc. NB. There is no association (AFAICT) between save-state files and these configuration files.

So I'm leaning towards saving configuration state (since it's already done), but not restoring it.

Alternatively configuration state could only be restored by ticking a check-box (in the Load Save State dialog).

tomcw:
FYI Nick, your changes to Disk.cpp etc are missing since I've not merged master into this branch yet.

sicklittlemonkey:
Hi Tom.

> The short summary is that it addresses #260, with this new format being
extensible and supporting all the AppleWin cards (except Uthernet).
Additionally the old v1 format is still supported (for loading).

It's a shame we didn't discuss this before you started. I would have voted
for a text-based snapshot that is human-readable and editable.

I would love to have startup scripts. In fact you can combine the ideas:
the script configures the machine and loads the memory state etc. A friend
and I did this 20 years ago, and it worked well. That way when you load a
disk image (even for clean power-on, not restoring state) it can configure
the emulator hardware and for monochrome or joystick or speed etc.

This really bridges the gap between AppleWin being handy to just run any
image, and the power to customize it to run in different ways for power
users who use it for play, cracking, development etc.

Any chance of talking you into a text-based format at this late stage? ; -
) I see the reading and writing is centralized, so presumably it only
requires changing there.

BTW, what's the problem with Uthernet?

> I don't think it's worth anyone actually reviewing this pull-request.
Maybe just give an opinion on the restoring of the configuration (see
below).

Ok, so I vote for saving and restoring all hardware state including machine
type (anything that might affect the software) but not the user preferences
- e.g.ConfirmReboot, ScrollLock toggle mode, volume settings.

Video mode falls in between these two. Again, user-editable save state /
script would cover it.

Another way would be to have a list of options for things to save - either
in the config or a dialog during the save process. The latter might
actually be nice - have the options and path there with a browse button.

I realize this is more feedback than you wanted, so take what you will, and
of course thanks for taking on such a big task in any case. : - )

Cheers,
Nick.

tomcw:
Hi Nick,
Thanks for taking the time to look it over.

>Any chance of talking you into a text-based format at this late stage? ; -)

I'll certainly take a fresh look at it given your strong preference.

Of course I considered this, but I think at the time I felt that storing the RAM as ascii would just bloat out the (human-readable) file. Naively 1 byte inflates to ",0xNN" which is 5 chars... so a 5x inflation. This takes a regular 128KiB save-state file to 640KiB. Is this a problem?

Worst-case is the 8MiB RAMworks card which would inflate to a 40MiB text file.
But arguably disk space is cheap, so who cares?

Of course this could be "compressed" by having runs of hex-nibbles without the 0x prefix. EG. Rows of 64 bytes would inflate to 128 chars + newline. So roughly doubling the size of the ascii save-state file. Further compression could use base64 encoding, but then this isn't easily searchable / editable, which would be a requirement from my POV.

Anyway, I'm also keen to have human-readable save-states files, so will take a look at switching things over. Any preference for using a standard (eg. XML,JSON)?

tomcw:
>BTW, what's the problem with Uthernet?

TBH, I just never got round to having a wired connection to allow me to experiment with this.

I think (just like laptop hibernate or console system suspend) I'd need to signal to the Uthernet card that connection had been dropped and then save the h/w state. On restore perhaps do the reverse or rely on the s/w stack to be robust enough to allow reconnecting.

Actually I think other Apple II h/w devices (eg. Serial card) that interface with the PC's real h/w don't do this, so the saving/restoring of their state isn't particularly robust!

sicklittlemonkey:
> I'll certainly take a fresh look at it given your strong preference.

Thanks for considering it. It would be a huge win for the power user if an
editable config script was a free side-effect.

If you want a coding monkey to do some grunt work here, I can help.

> Anyway, I'm also keen to have human-readable save-states files, so will
take a look at switching things over. Any preference for using a standard
(eg. XML,JSON)?

If you don't like the idea of scripting, and want a C++-friendly format,
either would be OK. I'd tend towards XML myself, though I suppose there
must be C++ JSON libraries out there now.

But the best option IMO would be a custom format - simply a file that is a
list of one line commands that is interpreted in order. The following are
just suggestions:
CPU.Registers.PC FF59
Drive.1 "C:/Dev/Emulators/Images/WavyNavy.dsk"

> Of course this could be "compressed" by having runs of hex-nibbles
without the 0x prefix. EG. Rows of 64 bytes would inflate to 128 chars +
newline. So roughly doubling the size of the ascii save-state file. Further
compression could use base64 encoding, but then this isn't easily
searchable / editable, which would be a requirement from my POV.

I agree about having hex, so just simple commands with parameters:
Memory.Main 0000 00112233445566 ... // put 128 on a line by defaut?
Memory.Main 0080 00112233445566 ...

or you could make it multi-line
Memory.Main 0000
00112233445566 ... // put 128 on a line?
00112233445566 ...
. // "." being some symbol for end

Perhaps we could detect and assign large ranges of the same value somehow:
Memory.Aux 4000-BFFF 00* // "*" being some symbol for repeat

Another way, and actually what we did 20 years ago is just have a load
command, but the downside is you have multiple files. We actually made the
scripts look a lot like AppleSoft BASIC.

Finally, the size is not an issue if - at some stage - we add support for
zipped state files.

Cheers,
Nick.

Michaelangel007:
I've been revisiting #254 but thought I would throw in my comments:

:+1: for text based config.  Human readable is good.  Increases work for us devs (parsing, etc) but users would much prefer a human editable text format.  Disk space shouldn't be an issue these days.

Personally, I prefer JSON since it is dead simple to parse.  Avoid bloated crap like XML please. I don't see the point of base64 -- if you're going to do that, might as well just binary at that point.

Nick, I like your label naming!  `Memory.Main` `CPU.Registers.PC`  At some point as we clean up the code base this should map 1-to-1 to class/object.

The more scripting support we expose the better for users long-term.  At some point we really should expose an DLL / API so that a third party program (bot) could read / write into Apple memory and IO states. :-)

tomcw:
> But the best option IMO would be a custom format

The "problem" is parsing / validation of the input file. Why reinvent the wheel, when there are readily available OSS parsers? Some simple C parsers:
* cJSON: http://sourceforge.net/projects/cjson/
* TinyXML2: http://www.grinninglizard.com/tinyxml2/index.html

> Finally, the size is not an issue if - at some stage - we add support for zipped state files.

Good point! (And clearly we should support zipped state files, since we support zip-files for all other input files.)

> If you don't like the idea of scripting

I'm not completely clear on the benefits or perhaps the use-case. If we have a save-state file that we can load from the cmd-line at start-up then isn't this the same thing? Or is the script a way to specify preferences & load save-states files? (ie. you have both a start-up script file and a save-state file?)

It sounds like we need the (human-readable) save-state file regardless. So start-up script support would be something in addition.

@Michael:
I don't think avoiding XML because of bloat isn't a reason. You say yourself that disk space isn't an issue.

tomcw:
Another off-the-shelf human-readable data serialisation format is YAML (I think a super-set of JSON).

In terms of human readability, XML gets a lot of negative press (backing up Michael's stance).

sicklittlemonkey:
Thanks for the feedback, guys.

> The "problem" is parsing / validation of the input file. Why reinvent the
wheel, when there are readily available OSS parsers? Some simple C parsers:

Well, I think parsing a line-by-line command-driven script is pretty
trivial, but sure, I'm open to the alternatives.

> Good point! (And clearly we should support zipped state files, since we
support zip-files for all other input files.)

Yes, this should be pretty trivial since we're already using the lib.

> I'm not completely clear on the benefits or perhaps the use-case. If we
have a save-state file that we can load from the cmd-line at start-up then
isn't this the same thing? Or is the script a way to specify preferences &
load save-states files? (ie. you have both a start-up script file and a
save-state file?)

Let me start with some obvious add-on examples that scripting would provide
- not necessarily in the first cut.

Prompting the user: for colour or B&W, or prompt for a game's number of
lives to patch in etc. This might require simple variables.
Running external tools: to integrate a development environment.

Eventually a scripting use-case might involve loops, which I'm not sure how
we would fit into XML or JSON.

But I'm open to ideas.

Sorry - gotta run - moving offices today; will respond again tonight!

Cheers,
Nick.

Michaelangel007:
Digressing, there are 3 main problems with XML:

- lots of extra fluff / crap which really hurts readability -- major eye sorefest
- bloat
- fragments memory like crazy

I agree in this instance bloat shouldn't be an issue, but at one game studio I worked at we exported all our data from 3D Studio into XML which we then "cooked" into a native raw format for the various consoles.  Our file sizes were roughly 10x bigger. A 40 MB text file was not common.  When you are exporting things day-in-day, this really added up.

Regardless, the main issue is human readability.  XML feels like a step backwards.

I _really_ like JSON.  The format is brain dead simple.

I have no comment on [YAML](https://en.wikipedia.org/wiki/YAML). I've never used it, so can't say the pros or cons.

I recommend we _avoid_ inventing yet-another-format.  Even as bad as XML is would be a better choice then.

@sicklittlemonkey could you give a use case / example for the `I would love to have startup scripts. ` that you would like to (eventually) see?  This would help us determine if this needs to be addressed sooner or later.

This bug is getting a little long, so I would move to close this one and maybe we should spin off 2 new bugs?

(A) Save State is in binary format not text
(B) Save Configuration doesn't support scripting

tomcw:
I think we can close this pull-request.

The general summary being that it's rejected as-is, and it'll get reworked so that the save-state file is human-readable. Specifically I'm fine to prefer JSON over XML, so will start down that route.

In terms of start-up scripting, I think this is something extra that can be built on top of a human-readable save-state file.

sicklittlemonkey:
> The general summary being that it's rejected as-is, and it'll get
reworked so that the save-state file is human-readable. Specifically I'm
fine to prefer JSON over XML, so will start down that route.

I still think a DSL will be much easier to generate and parse than even
JSON. You have to ask yourself why you are writing the data in a format
no-one else will parse. ; - )

Anyway, human-readable is a great result either way.

Cheers,
Nick.

tomcw:
Hi Michael, Nick,

I've picked this up again, and have YAML support for save/load of the core Apple II system (ie. no cards). I want to share this with you before I go any further to get your buy-in and thoughts.

Here's an example of a save-state file:
(NB. I had to add a .txt extension to attach it to this comment, but for AppleWin I would prefer to keep the .yaml extension which is the standard way of doing things for YAML files.)

[TestSSv2.aws.yaml.txt](https://github.com/AppleWin/AppleWin/files/66689/TestSSv2.aws.yaml.txt)

Let me know what you think.

If you want to see some code then look at the "savestate_yaml" branch:
https://github.com/AppleWin/AppleWin/tree/savestate_yaml

sicklittlemonkey:
It looks great!

I can't believe you found some free time at this crazy time of year! ; - )

Cheers,
Nick.

tomcw:
Hi Michael, Nick,

I'm close to having the new v2 (YAML-based) save/load state complete.
Here's an example:
[Cybernoid.aws.yaml.txt](https://github.com/AppleWin/AppleWin/files/120970/Cybernoid.aws.yaml.txt)

(Again it is suffixed with ".txt" to allow attaching to this GitHub comment.)

The basic schema is:
. main Apple II unit (CPU, video, main memory, etc)
. aux memory
. slots 1-7

I'll finish tidying up, and then put it up as a new pull request.

Michaelangel007:
Looks good!  :+1:  for a simple human readable text file without all the bloat of XML, especially the memory +0x40. :-)

Two minor concerns:

    CPU:
      Type: 65C02 # NB. Currently ignored

1. I'm assuming CPU type `65C02` will be eventually used? We need the option to use the plain `6502` on the Apple ][, and down the road the `65D02` (my debugger-focused CPU).

2. Some of the advanced graphics demos change the video mode mid-scan line.  We don't need to solve this right now but how are we going to restore the frame buffer correctly?

tomcw:
Hi Michael,

Thanks for checking this.

1. Actually the 'Model' implicitly defines the CPU type (eg. a Model of 'Enhanced Apple//e' means CPU=65C02; and a Model of 'Apple II' means CPU=6502). This currently just reflects the way that the internals of AppleWin works. I'll look to change this to remove this small confusion (ie. to allow 'CPU' to override the Model's default CPU type).

2.  Video mode switching mid-scanline will now work fine. This new save-state file has a new 'Video' -> 'Cycles This Frame' attribute. This persists the `g_dwCyclesThisFrame` variable, which is used by the `VideoGetScannerAddress()` function. All this means that the running code remains in sync with the video across a save/load state operation. I have tested this with the [French Touch's](http://www.ctrl-pomme-reset.fr/french-touch/) Ansi Story end credits part.

sicklittlemonkey:
I haven't looked at the code, but your answers to those points sound good!

Cheers,
Nick.

PR #315, opened 2016-02-14
Save-state v2 (yaml: human-readable)

tomcw:
This extends the previous Save-state v2 pull request here: https://github.com/AppleWin/AppleWin/pull/304

As before: the short summary is that it addresses #260, with this new format being extensible and supporting all the AppleWin cards (except Uthernet & SAM(*)). Additionally the old v1 format is still supported (for loading).

I don't think it's worth anyone actually reviewing this pull-request (there are lots of changes affecting many files). But I have done a fair bit of testing for different h/w configs & machine types.

Unlike the previous Save-state v2, this new one doesn't save the configuration state.

The basic save-state file schema is:
. main Apple II unit (CPU, video, main memory, etc)
. aux memory
. slots 1-7

(*) re. lack of Uthernet & SAM support
I didn't get round to doing these in this initial round. Adding support should be easy, but for Uthernet I do wonder how robust the s/w TCP stack will be to cope with a save -> restore.


tomcw:
I've just noticed that GitHub will only partially serve up the "Files Changed" web-page because of the yaml doxygen files being so big.

So if you want to just take a look at a small (but meaningful) piece of this pull request then look at `DiskSaveSnapshot()` and `DiskLoadSnapshot()` [here](https://github.com/AppleWin/AppleWin/blob/savestate/source/Disk.cpp).

sicklittlemonkey:
Hi Tom.

A couple of comments, now that you point at one file I have time to look
at! : - )

<puts on asbestos underwear>

The YamlSaveHelper could perhaps be more helpful.

Compare this:
yamlSaveHelper.Save("%s: %s\n", SS_YAML_KEY_FILENAME,
yamlSaveHelper.GetSaveString(g_aFloppyDisk[unit].fullname).c_str());
yamlSaveHelper.Save("%s: %d\n", SS_YAML_KEY_TRACK,
g_aFloppyDisk[unit].track);
yamlSaveHelper.Save("%s: %d\n", SS_YAML_KEY_PHASE,
g_aFloppyDisk[unit].phase);
yamlSaveHelper.Save("%s: 0x%04X\n", SS_YAML_KEY_BYTE,
g_aFloppyDisk[unit].byte);

To this:
yamlSaveHelper.SaveString(SS_YAML_KEY_FILENAME,
yamlSaveHelper.GetSaveString(g_aFloppyDisk[unit].fullname).c_str());
yamlSaveHelper.SaveInt(SS_YAML_KEY_TRACK, g_aFloppyDisk[unit].track);
yamlSaveHelper.SaveInt(SS_YAML_KEY_PHASE, g_aFloppyDisk[unit].phase);
yamlSaveHelper.SaveHex4(SS_YAML_KEY_BYTE, g_aFloppyDisk[unit].byte);

That would help a lot with consistency and avoiding errors.

Another thing to consider: Is there anywhere better to put the string
constants, and should they be defines?

I guess the options are in each component file or one common file.
Actually, in the component file is probably easier to add/edit.

In fact C++ string constants are a pain. Defines work fine here. ; - )

Cheers,
Nick.

tomcw:
Hi Nick,

No need for the asbestos hat - I'm happy to have this critiqued! :)

Good idea about the YamlSaveHelper having type-specific save functions (in fact similar to the YamlLoadHelper). I'll take a look at making that modification.

re. string constants
I'd prefer to keep these strings local to each specific .cpp file, ie. if you think of these as classes (which a few of them are, eg. SerialComms.cpp) then keeping these string constants object-oriented.

Thanks.

sicklittlemonkey:
I see what you mean about YamlLoadHelper. Pity it doesn't stick to
camelCase, but I suppose it's better to be consistent. SaveUINT etc.

And yeah, as I concluded above, in the file is better. In C# and Java
string constants can be nicer, but in C++ they're not very elegant. At
least defines are dead simple.

Cheers,
Nick.

tomcw:
Hi Nick,

I've taken your good suggestion (above) on board and implemented this in 6b07fd6 (save-type funcs) and 11789c9 (load-type funcs). Or use this link just to see how [Disk.cpp](https://github.com/AppleWin/AppleWin/blob/savestate/source/Disk.cpp) now looks.

Take a look and let me know your thoughts.

sicklittlemonkey:
Hi Tom.

Sorry for the late reply. Busy weekend, including blueberry picking. Now
baby poo is blue. ; - )

Anyway, the code looks much nicer.

Final quibble about SaveHex8 etc ...
The number of bytes (or even digits) seems more logical to me than the
number of bits.
I guess you've done this because some print just one digit.

Anyway, either way is just a matter of taste. Well done for taking on such
a huge task!

Ship it! : - )

Cheers,
Nick.
​

tomcw:
Hi Nick,

re. SaveHex8() etc

I know what you mean, but I like the flexibility of bit-length (although I'm currently just using 4, 8, 12, 16, 32 and 64 bit-lengths). I'm sure if I look hard enough there's probably a 3-bit or 7-bit value being saved as a wider 4- or 8-bit value.

Perhaps `SaveUint8Hex()` feels better as there are similarities to the `uint8_t` and `UINT8` types?
This also fits with the existing, non-hex, general, `SaveUint()` function.

sicklittlemonkey:
I think the subtypes being the last thing in the name makes more sense.

So SaveHex4/8... or SaveUintHex4/8...

Cheers,
Nick.

tomcw:
I toyed with the idea of `SaveHexUint4()`, `SaveHexUint8()` etc. but ultimately this is just aesthetics (or matter of taste, as you said), so I think I'll probably just leave it - but I do think this "UintN" suffix more clearly conveys **bits** not nibbles... let me mull this a bit longer.

Anyway you've given the thumbs up with your "ship it" above, so I'll probably merge in the coming days (perhaps with the above rename).

Thanks!

sicklittlemonkey:
You're right when I think about int64 etc, so yeah, sounds good.

After all, naming is the hardest thing in programming. ; - )

Cheers,
Nick.

tomcw:
Hi Nick,

OK I made the change discussed above (+ fixed-up the VS2013 sln/proj), and have now merged the pull request into master. Thanks for your input on this.

In terms of testing, I've got save-state files that test all cards / machine types. My plan is to put these into a separate git repo (eg. AppleWin-test). The reason for this is that the images (.dsk / .hdv) are binary files and if they get updated then all versions will accrue in the repo.

OTOH, the advantage of putting them directly into the AppleWin repo is that they'll be in sync with the commits to that repo (making testing specific versions easier).

The current total file size (for this test repo) is ~13MB for 24 save-states and associated disk image (it's not one-to-one as some disk images are reused for multiple save-state). Not huge, but this will grow as more tests get added.

Anyway, it's easy to start with this separate git repo first (AppleWin-test). If we change our mind, then it's simple enough to move into the AppleWin git repo.

Issue #141, opened 2014-05-23
More than 2 disk drives


tomcw:
**BerliOS Feature Request #5465**

**Date:** 2011-November-06 17:50
**Submitted By:** tommyapple
**Category:** none
**Assigned To:** none
**Priority:** 5
# 

Date: 2011-Nov-06 17:50
Sender: tommyapple
Logged In: YES 
user_id=60767
Browser: Mozilla/5.0 (Windows NT 5.1; rv:7.0.1) Gecko/20100101 Firefox/7.0.1

Hi,

I don't know if this has been suggested or if it is
possible with the current version of AppleWin:

```
 I'd like to be able to use more than 2 disk drives.  
```

Is it possible to have a disk controller in slot 5 with
2 drives attached?

I am currently working on a large "project"\* of my own
using the Apple Pascal system, and having only 2 drives
is very frustrating.   Four drives (or even three)
would save me a lot of "disk-swapping".

Thanks,
Tommy
- The personal project I'm working on is disassembling
  the Wizardy 3, Legacy of Llylgamyn executables
  (Wizardy.Code and Wizutil.code), which were originally
  written in Pascal.   I am looking at the generated
  p-code on the commercial diskette and writing source
  code that when compiled produces the executable code. 
  It is more than just re-engineering in the sense that I
  will have exact source code to reproduce the
  executables.  To date I have successfully
  "re-engineered" the 43 subroutines that make up
  WizUtil.code.

PR #337, opened 2016-08-30
Tommy's 4 disk drive AppleWin feature

TommyGH:
This pull request is for Tommy's 4 disk drive AppleWin feature.

I've put a lot of information into the README.md.  You might want to just start reading it.
This code will likely never be integrated into AppleWin/AppleWin, but I did think I should share it.
Let me know if there are further actions you would like me to take.

Tommy


tomcw:
Hi Tommy,

> Note to AppleWin maintainers  
> ----------------------------  
> I'm new at this GIT and GITHUB stuff so please be patient with me.  
> 
> This "pull" request is not really a true pull request.  This code is nowhere near  
> correct or complete enough to integrate into the AppleWin master as it is.  I've  
> created this pull request only to make this code visible to others who might be  
> interested in an AppleWin with more than 2 disk drives.  I don't intend to improve  
> on it, but I feel it might be worth sharing.										  
> 
> Is there a better way of doing this?  Is it better to have a "branch" from AppleWin  
> that we know will never be pulled into master?  Is it worthwhile (or even desirable)  
> for AppleWin users to see and download the executable created as a result of this  
> code? and how best to do that?  In other words, if a user for AppleWin desires an  
> emulator with 4 disk drives (say they are using ApplePascal), how would they even  
> know that I've already created an "alpha" version?										  
> 

You could publish the executable from your repository (TommyGH/AppleWin).
You could announce it via the comp.emulators.apple2 news-group... but you'd have to keep pace with any new releases that we made.

We still have your original request (#141) from 2011. We are gradually creeping slowly towards a more configurable AppleWin where hopefully one day we can achieve what you have done in this pull request.

Thanks,
Tom

Michaelangel007:
Since this hijacks Slot 5 which is also used by the mockingboard we probably should a radiobutton to options so a user can can select:

`Slot 5`
* `( ) Mockingboard`
* `(x) Disk II Controller Card`

Michaelangel007:
I like where you are going with this Tommy.

But before this PR can be merged there needs to be some (major?) cleanup first. :-/

A single commit makes it hard to comment on individual changes (I guess I could annotate each line but I'm consolidating them all here so bear with me ... )  Remember with git: _commit early, git often._

* The (binary) file `AppleWinExpress2015.VC.db` should not be in the PR as it is a debug compiler generated file.
* The drive 1 & 2 resource BMP files in `AppleWinExpress2015.vcxproj` should NOT be renamed, Specifically [Lines 261-263](https://github.com/TommyGH/AppleWin/blob/69145f94ea4617b837f2e83ba25267a7969373ee/AppleWinExpress2015.vcxproj#L261-L263)
* All the changes in [README.MD](https://github.com/TommyGH/AppleWin/blob/69145f94ea4617b837f2e83ba25267a7969373ee/README.md) should be moved out into the comment into the PR. i.e. The list of files changed doesn't help users reading the readme -- it is intended for developers.
* The file version bump from `VALUE "FileVersion", "1, 25, 0, 4"` to `VALUE "FileVersion", "1, 25, 0, 41"` would normally be `.5` but it probably should be x.x.1.x -- `VALUE "FileVersion", "1, 25, 1, 0"`  as 4 drives isn't just a simple bug fix, it is a new feature.
* I would like see *all* the changes for 4 disks should be annotated with `// DISK4:` instead of nebulous `// Changes needed for "4 disk" (?)` -- so we can search for them in the future.
* The comment in Disk.cpp [Line #503](https://github.com/TommyGH/AppleWin/blob/69145f94ea4617b837f2e83ba25267a7969373ee/source/Disk.cpp#L503) should probably have a `// HACK:` prefix so we know that this is a kludge that might eventually need fixing.
* It looks like the snapshort support for Drives 3 and 4 is [incomplete on Lines 1398-1399](https://github.com/TommyGH/AppleWin/blob/69145f94ea4617b837f2e83ba25267a7969373ee/source/Disk.cpp#L1398-L1399) ?

This would be a good addition once the above issues are addressed.

tomcw:
I'm finally closing this PR, since at AppleWin 1.32.0, #853 is finally closed, which includes support for this PR.

PR #339, opened 2016-09-10
Miscellaneous smallish changes

kiyolee:


Michaelangel007:
@kiyolee Kelvin, this looks good!  Nice cleanup work!  There is only 1 minor change that would be nice to see:

* `g_restart` should be `g_bRestart`

@tomcw Tom, I took a look over this PR:

* Aside from the rename of `Applewin.exe` to `AppleWin.exe` (I'm indifferent on this)

I'm happy to accept this PR as-is.  The cleanup of `g_bRestart` would just be a bonus.

kiyolee:
Note that while the last set of commits (7e8b463 to f3f373e) does indeed bring VS2010, VS2013 and VS2015 all more in sync and VS2013 and VS2015 both build with no problem, VS2010 is still not building that ddraw.lib and dxguid.lib does not exist. However VS2010 from origin is also broken anyway but for other problems. VS2008 is also broken for other reasons, perhaps the code itself no longer support SDK that are too old. Cannot test VS2005 just yet.

kiyolee:
With the last set of commits (53e778a to cf65057), VS2008 and VS2010 can now build right out of the repository.

tomcw:
Thanks for the pull-request.

A few general comments:

1. I'll be switching to VS2008 for building AppleWin 1.26, so was planning on deprecating the VS2005 proj/sln. (NB. No need to change anything in this pull request - the changes can be accepted and then will just disappear when the VS2005 files go.)

2. In the `bin` folder I don't know why we are carrying an AppleWin.exe. It's not sync'ed with any tags or releases (if you want a specific release then get it from the AppleWin git project's `releases` tab). Also git doesn't work well with binary files. IMO, this can be deleted.

3. In 7e8b46, renaming a file & changing the same file's content looks to have resulted in deleting & re-adding these files (ie. 1000s of changed lines and hard to see what has really changed). From a (very!) quick google, the [advice](https://www.reddit.com/r/git/comments/3uu7bo/how_well_does_git_handle_mass_file_renames/) seems to be rename & change in separate commits.

4. I want to verify the removal of ddraw.lib before accepting this.

Michaelangel007:
> 3. In 7e8b463, renaming a file & changing the same file's content looks to have resulted in deleting & re-adding these files (ie. 1000s of changed lines and hard to see what has really changed). From a (very!) quick google, the advice seems to be rename & change in separate commits.

@kiyolee Echoing what Tom said. For the future, "Git best practices" are rename, commit, change, commit.

@tomcw I took a quick look at the changes in commit https://github.com/AppleWin/AppleWin/pull/339/commits/7e8b463aa3c89111a75035a1d26d1b852cb3212f and they look fine. Fortunately, GitHub highlights the changes but it is going to take a while to go through all of these and verify them.

kiyolee:
Thanks for the advice. I am relatively new to the git stuff. This is my first PR to any project. So may I assume the commits are all fine now? Or any suggestion to prove?

tomcw:
Hi Kelvin,

From the brief look at the commits, most seem fine. I think I might be able to selectively cherry-pick most of them (maybe avoid the applewin.exe rename & VS2005 fixes though).

btw. since none of these affect the imminent AppleWin 1.26, then I'll defer the cherry-picking until after the 1.26 release.

tomcw:
Cherry-picked the first batch of patches: ae336d6 to f7e7b54 (excluding the two relating to bin\AppleWin.exe: f779493, dbc6789).

NB. bin\AppleWin.exe was deleted at eb2c773.

tomcw:
re. 53e778a -  "ddraw.lib is not actually needed. (so it seems)"
We can remove ddraw.h and ddraw.lib, as we have now dropped the old DirectDraw (ddraw) for full-screen.

Also see: [Whither DirectDraw?](https://blogs.msdn.microsoft.com/chuckw/2010/06/16/whither-directdraw/)
>The DirectDraw API has been more or less deprecated for game developers since the release of DirectX 9.0 SDK back in 2002. ...

tomcw:
re. VS2010

I don't have VS2010 on any of my dev machines, so there's no way for me to test these VS2010 related commits. And I don't want to install VS2010 - I just recall it being terribly slow!

re. https:​//en.​wikipedia.​org/wiki/Microsoft_Visual_Stud​io

>Visual Studio 2008 is the last version to support targeting Windows 2000 for C++ applications.
>
>"Visual Studio 2012 Update 1" (Visual Studio 2012.1) was released in November 2012. This update added support for Windows XP targets and also added other new tools and features (e.g. improved diagnostics and testing support for Windows Store apps).

In fact I don't actually think VS2010 is needed, since VS2012-update1 offers the same (or better) than VS2010.

Michaelangel007:
I use VS2010 on my main Windows 7 dev box.  While VS2010 is _NOT mandatory_ it IS recommended.  I tend to get a little cranky when AppleWin fails to build on my box because someone forgot to "back-port" the 2015 solution/project configs back to 2010. ;-)

Us devs tend to "leap-frog" MSVC versions so we can spot potential config conflicts or other MSVC issues.

tomcw:
Hi Michael,

Can't you switch from VS2010  to VS2012 on your Win7 dev machine?
Also why do you say it is recommended?

My problem is that I can't really merge any of the VS2010 pull-requests without testing them. If you insist on sticking with VS2010, then can I ask you to review and merge these VS2010 changes?

There's also the VS2017 changes. Nick says auto_ptr is no longer supported, so that probably needs a dedicated github issue outside of this PR.

kiyolee:
I already have VS2017 working in my own fork.
auto_ptr is meant to be replaced by unique_ptr as part of C++11 or 14. (not sure which)
auto_ptr is still available in VS2017 and works.
The problem is you need to include <memory> for VS2017.
I added <memory> to StdAfx.h to fix the VS2017 build.
One technique to keeping old code while using C++11/14 unique_ptr is to typedef unique_ptr as auto_ptr if compiler is of C++11/14 compliant.

tomcw:
> Btw, I would like to comment that VS2012 was for targeting Windows 8 (NOT 8.1) and so along the dismiss of Windows 8, VS2012 is pretty much consider dead or simply non-existing.
> Moving forward from VS2010 should consider either VS2013/2015/2017.

I'm considering VS from the point of view of AppleWin supporting older Windows OS versions.
(Not that I do any testing on OS's older than Win7-64, but rely on end-user feedback to highlight issues.)

VS2005 - last VS version to support Win98/ME
VS2008 - last VS version to support Win2000: [ref](https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#2008)
VS2010 - supports WinXP and above
VS2012 - dropped support for WinXP
VS2012-update1 added support back in for WinXP
VS2013/15/17 - do these still include support for WinXP targets?

At AppleWin 1.26 we moved from VS2005 to VS2008, and so dropping support for Win98/ME.

I'm fine with sticking with VS2008 (for the official release build) for now, but it's useful to understand the implications of moving to a newer VS version.

If in the future we moved to a newer VS, then implicitly that means Win2000 is no longer supported by AppleWin (probably fine, but we should ask on the cea2 newsgroup). Now WinXP becomes the new base OS, so what is the minimum VS version that still supports WinXP?

kiyolee:
Yes, VS2013/15/17 does have XP support. (Never used myself though.)
The corresponding Platform toolset settings are v120_xp, v140_xp, and v141_xp.
For comparison, the default settings without XP support are v120, v140, and v141.

sicklittlemonkey:
Thanks Kelvin. I meant to say to Tom that I thought someone had mentioned
2017 recently

Tom, you asked:
> VS2013/15/17 - do these still include support for WinXP targets?

"Configuring Programs for Windows XP"
https://msdn.microsoft.com/en-us/library/jj851139.aspx

That's for 2015 but 2017 still has the same option.
BTW, they've optimized 2017 to load quicker than 2015.

> If in the future we moved to a newer VS, then implicitly that means
Win2000 is no longer supported by AppleWin (probably fine, but we should
ask on the cea2 newsgroup). Now WinXP becomes the new base OS, so what is
the minimum VS version that still supports WinXP?

I guess there's no reason to hurry.

It would be more interesting to support the web than XP. ; - )
https://medium.com/mozilla-tech/why-webassembly-is-a-game-changer-for-the-web-and-a-source-of-pride-for-mozilla-and-firefox-dda80e4c43cb

Cheers,
Nick.

tomcw:
Good to see involvement from Google's PNaCl team:
https://auth0.com/blog/7-things-you-should-know-about-web-assembly/

Michaelangel007:
Tom, could you clarify your last comment please? I'm not following how Google's Sodium Chloride team is related to AppleWin? Thanks.

sicklittlemonkey:
Michael, Tom is replying to my (half) joking comment about porting AppleWin
to the web using web assembly.

Cheers,
Nick.

Michaelangel007:
Thanks Nick -- guess I missed that *whoosh* :-)

The funny thing is Web Assembly is slowly gaining traction.  With these days one can compile C++ to Javascript it is hard to tell what is serious / joking anymore.

Michaelangel007:
For the VS2017 cleanup -- are we going to run into merge conflicts with https://github.com/AppleWin/AppleWin/pull/410 ?

tomcw:
Cherry-picked VS2017 support from 1ea5a5a and aabf0fa.

tomcw:
Outstanding in this PR:

1. 53e778a - "ddraw.lib is not actually needed. (so it seems)"
2. VS2010 related, see [above](https://github.com/AppleWin/AppleWin/pull/339#issuecomment-281809979)

kiyolee:
My master branch may have become a bit difficult to merge. Breaking down my changes into other sperate pull requests.

PR #364, opened 2016-10-19
Avoid double-counting extra cycles on IRQ boundary

mauiaaron:
Small patch that avoids double-counting uExtraCycles at IRQ boundaries.

Regards,

### DIFF
--- a/source/CPU.cpp
+++ b/source/CPU.cpp
@@ -433,7 +433,11 @@ static __forceinline void IRQ(ULONG& uExecutedCycles, UINT& uExtraCycles, BOOL& 
 		PUSH(regs.ps & ~AF_BREAK)
 		regs.ps = regs.ps | AF_INTERRUPT & ~AF_DECIMAL;
 		regs.pc = * (WORD*) (mem+0xFFFE);
-		CYC(7)
+
+		UINT extraCycles = uExtraCycles;
+		uExtraCycles = 0;
+		CYC(7);
+		uExtraCycles = extraCycles;
 	}
 }
 


source/CPU.cpp
```diff
@@ -433,7 +433,11 @@ static __forceinline void IRQ(ULONG& uExecutedCycles, UINT& uExtraCycles, BOOL&
 		PUSH(regs.ps & ~AF_BREAK)
 		regs.ps = regs.ps | AF_INTERRUPT & ~AF_DECIMAL;
 		regs.pc = * (WORD*) (mem+0xFFFE);
-		CYC(7)
+
+		UINT extraCycles = uExtraCycles;
+		uExtraCycles = 0;
+		CYC(7);
+		uExtraCycles = extraCycles;
```
tomcw: Why pass in  `uExtraCycles` at all to this `IRQ()` function? Just declare a local `uExtraCycles` set to zero.
Also you need to replicate for `NMI()`.

tomcw: Why pass in  `uExtraCycles` at all to this `IRQ()` function? Just declare a local `uExtraCycles` set to zero.
Also you need to replicate for `NMI()`.

tomcw: Why pass in  `uExtraCycles` at all to this `IRQ()` function? Just declare a local `uExtraCycles` set to zero.
Also you need to replicate for `NMI()`.

tomcw:
IMO this is best handled (and tracked) as a bug, not a pull request. See #366.

tomcw:
@mauiaaron: thanks for reporting this!

mauiaaron:
OK sounds good.  Figured you folks are alerted to this [minor] issue.  Will report any other issues I find in the future as bugs rather than PRs.  Regards,

PR #410, opened 2017-05-01
Allow compilation with VS2017

audetto:
1) neet to include memory explicitly.
2) auto_ptr has gone. replace with shared_ptr as it works everywhere


audetto:
What I meant is that std::shared_ptr<> works for VS2010 and afterwards.

If you are still targeting old compilers VS2008 and VS2005, then an #ifdef is probably required.

Something like this

#if _MSC_VER >= 1600 // VS2010
  std::shared_ptr<BYTE> pTrack;
#else
  std::auto_ptr<BYTE> pTrack;
#endif
  pTrack.reset(new BYTE[NIBBLES_PER_TRACK]);

tomcw:
I build the release (and still mainly develop) using VS2008 so will need the above #ifdef.
Can you fix this?

FYI, I discussed use of VS2008 in PR #339, and want to continue to use it. (But VS2005 isn't needed anymore, and I will  delete those sln/proj files soon.)

audetto:
I will submit an other one.

PR #412, opened 2017-05-02
Needed in VS2017.

audetto:
So I have regenerated the VS2017 solution and come to a different conclusion.

`#include <memory>`

is always needed.

Now, for auto_ptr<>, this actually depends on a compiler switch /std:c++14 vs /std:c++latest.
With the former (the default as well), auto_ptr still exists, which means that with the default conversion there is no need to change anything else.

So in this patch I do the bare minimum and include only `<memory>`.

I can only imagine that when I converted the solution a few week ago, I must have changed that switch, hence the need to replace auto_ptr.
Or VS2017 updated itself and changed behaviour, I remember installing it before it was fully released...

### DIFF
--- a/source/StdAfx.h
+++ b/source/StdAfx.h
@@ -46,6 +46,7 @@ typedef UINT32 uint32_t;
 #include <stack>
 #include <string>
 #include <vector>
+#include <memory>
 
 // SM_CXPADDEDBORDER is not supported on 2000 & XP:
 // http://msdn.microsoft.com/en-nz/library/windows/desktop/ms724385(v=vs.85).aspx


sicklittlemonkey:
I think which behaviour is the default may have changed, as I had the
auto_ptr errors a few weeks back.

Anyway, sounds good.

Cheers,
Nick.

tomcw:
This was cherry-picked from another PR, and is now in master (b02c7ff), so I will close this PR.

btw. interesting info about the compiler switches /std:c++14 and /std:c++latest.
I guess this will affect us at VS2019(?).

PR #414, opened 2017-05-04
Fix the following errors reported by VS2017 Code Analysis.

audetto:
applewin\source\applewin.cpp(761): warning C6385: Reading invalid data from '"Disk Image"':  the readable size is '11' bytes, but '21' bytes may be read.
applewin\source\debugger\debug.cpp(6624): warning C6386: Buffer overrun while writing to 'sName':  the writable size is '31' bytes, but '32' bytes might be written.
applewin\source\debugger\debugger_display.cpp(3715): warning C6053: The prior call to 'strncpy' might not zero-terminate string 'sText'.
applewin\source\log.cpp(42): warning C6053: The prior call to '_vsnprintf' might not zero-terminate string 'output'.
applewin\source\debugger\debug.cpp(2759): warning C6011: Dereferencing NULL pointer 'pFont'.
applewin\source\debugger\debugger_symbols.cpp(243): warning C6053: The prior call to '_tcsncpy' might not zero-terminate string 'pText'.
applewin\source\diskimagehelper.cpp(1132): warning C6053: The prior call to '_tcsncpy' might not zero-terminate string 'pszExt'.
applewin\source\diskimagehelper.cpp(1141): warning C6053: The prior call to '_tcsncpy' might not zero-terminate string 'szFilename'.
applewin\source\parallelprinter.cpp(242): warning C6053: The prior call to '_tcsncpy' might not zero-terminate string 'g_szPrintFilename'.

The one about RegSetValue(), according to Microsoft

https://msdn.microsoft.com/en-us/library/windows/desktop/ms724922(v=vs.85).aspx

the last argument cbData is ignored, so I set it to 0, as in some cases was anyway wrong (see "DiskImage").

### DIFF
--- a/source/Log.cpp
+++ b/source/Log.cpp
@@ -39,7 +39,7 @@ void LogOutput(LPCTSTR format, ...)
 	va_start(args, format);
 
 	_vsntprintf(output, sizeof(output) - 1, format, args);
-	output[255] = 0;
+	output[sizeof(output) - 1] = 0;
 	OutputDebugString(output);
 }
 
@@ -58,5 +58,6 @@ void LogFileOutput(LPCTSTR format, ...)
 	va_start(args, format);
 
 	_vsntprintf(output, sizeof(output) - 1, format, args);
+	output[sizeof(output) - 1] = 0;
 	fprintf(g_fh, "%s", output);
 }
--- a/source/Log.cpp
+++ b/source/Log.cpp
@@ -39,6 +39,7 @@ void LogOutput(LPCTSTR format, ...)
 	va_start(args, format);
 
 	_vsntprintf(output, sizeof(output) - 1, format, args);
+	output[255] = 0;
 	OutputDebugString(output);
 }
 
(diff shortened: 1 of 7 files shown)


source/Log.cpp
```diff
@@ -39,6 +39,7 @@ void LogOutput(LPCTSTR format, ...)
 	va_start(args, format);
 
 	_vsntprintf(output, sizeof(output) - 1, format, args);
+	output[255] = 0;
```
tomcw: 1. Should use `sizeof(output)-1` instead of 255.
2. Should apply the same fix to the next function, `LogFileOutput()` which has the same line of code.

audetto:
I've incorporated the feedbacks in Log.cpp.

tomcw:
I'm fine with the non-debugger patches.
@Michaelangel007 - can you take a look at the debugger patches?

Michaelangel007:
:+1 Debugger patches look good.

Michaelangel007:
That's neat feature -- request review ! Here is a link with some more information:

* https://help.github.com/articles/about-pull-request-reviews/

AH, OK, at the top of this page there is a `Add your review` button.

PR #428, opened 2017-06-04
Input option to use PC Joystick 1 Thumbstick 2 for emulated Joystick 2

sicklittlemonkey:
Only really useful for a controller with two thumbsticks (e.g. Logitech F310) and for Apple II software that uses both joysticks for one user.


tomcw:
Hi Nick,
Wow! This is great fun with this new controller mode!
I'm using my PS4 DualShock 4 controller, which works perfectly with your AppleWin code additions.

Anyway, LGTM, so merge away when you are ready.

sicklittlemonkey:
Glad you like it! Ok, in it goes.

PR #442, opened 2017-07-22
Fixed VS2010 to build.

kiyolee:
1. Removed reference to AppleWin.exe.intermediate.manifest. (the main problem that breaks the build)
2. SAM.cpp had multiple references.
3. Moved Pravets.\*, YamlHelper.\* to locations following other projects.
4. Renamed subfolder "Resources" to "Resource Files" and "Source" to "Source Files" to follow other projects.


Michaelangel007:
Go ahead and drop support for 2010 since I have MSVC 2019 installed now.

Issue #479, opened 2017-08-24
Screen partly corrupted using DEBUGGER. "111111111111" outside disassembly window.


TommyGH:
This is a bit of a weird one.  I was using the down arrow on my keyboard to scroll through some code in the disassembly pane.  All of a sudden 2 characters to the right of the disassembly pane (is it called the "info panel"?) were corrupted.  Each time I pressed the down arrow key, another " 1" appeared.  See attached screen captures.

It doesn't seem to be directly related to WM_PAINT, as I could move another application window over the corrupted window, then move it off again, and I could still see the corruption.  However, if I minimized AppleWin and then restored Applewin, the window was displayed without the corruption.

The problem is repeatable.  I could close Applewin, then restart it, and scroll through this same sequence of code and it would corrupt in the same manner.  There seems to be more than one place in the code that causes this corruption.

AppleWin:  1.26.3.0
Debugger:  2.9.0.8

I was using the enclosed DEBUG.TXT script and the enclosed A2_USER1.SYM.  I booted DOS(1983) (January) Master disk first before F7 and RUN DEBUG.TXT.

Corrupt Screen:
![corruptpaintjob](https://user-images.githubusercontent.com/7924807/29644516-672ac1e0-883c-11e7-8683-d18fa13e7c9b.jpg)

Good Screen:
![goodscreen](https://user-images.githubusercontent.com/7924807/29644545-9de1bb08-883c-11e7-92e9-75f7be3b6388.jpg)

First sign of corruption:
![firstsignofcorruption](https://user-images.githubusercontent.com/7924807/29644551-a68be0d0-883c-11e7-8c87-3e915fab61e8.jpg)

[debug.txt](https://github.com/AppleWin/AppleWin/files/1247425/debug.txt)

[A2_USER1.SYM.txt](https://github.com/AppleWin/AppleWin/files/1247426/A2_USER1.SYM.txt)

Michaelangel007:
Repro.:

1. Rename `a2_user1.sym.txt` to `a2_user1.sym` to auto-load symbols at boot
2. Start AppleWin
3. Boot: ` Apple DOS 3.3 January 1983.dsk`
4. `F7`
5. `run debug.txt`
6. `Down arrow` until `BD81`

Michaelangel007:
Trace:

* `CmdCursorLineDown()`

Michaelangel007:
Also, $BD7F and $BD81 are showing a few pixels of the background outside the disassembly pane.

[...]

PR #480, opened 2017-08-24
Bug 479 fixes

Michaelangel007:
See #479

### DIFF
--- a/source/Debugger/Debugger_Display.cpp
+++ b/source/Debugger/Debugger_Display.cpp
@@ -2129,12 +2129,53 @@ WORD DrawDisassemblyLine ( int iLine, const WORD nBaseAddress )
 		if ( !g_bConfigDisasmOpcodesView )
 		    nMaxLen += (MAX_OPCODES*3);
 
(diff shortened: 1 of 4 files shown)


source/Debugger/Debugger_Display.cpp
```diff
@@ -2129,12 +2129,53 @@ WORD DrawDisassemblyLine ( int iLine, const WORD nBaseAddress )
...
+			nOverflow += 2;                             //      '##'
+			nOverflow ++  ;                             //         ' '
+		}
+
+		if (bDisasmFormatFlags & DISASM_FORMAT_CHAR)
+		{
+			nOverflow += strlen( line.sImmediate );
+		}
+
+		if (nLen >=  (nMaxLen - nOverflow))
```
tomcw: @Michaelangel007: can nMaxLen-nOverflow ever go negative?
I'm just a little worried about the array index on pTarget (below) if it does - ie. an out of bounds access.

tomcw:
@Michaelangel007 - other than the one comment/question, then this LGTM.

Issue #125, opened 2014-05-23
Cannot format a ProDOS discs


tomcw:
**BerliOS Bug #17707**

**Date:** 2010-November-10 19:20
**Submitted By:** georgejh
**Category:** none
**Bug Group:** none
**Assigned To:** warthurton
**Priority:** 4
# 

Hi,

Error "DISK II DRIVE TOO SLOW" appears when attempting to format a ProDOS disk.

I am using ProDOS 1.1.1 Utility disk on S6 D1 and a copy of Master.dsk (DOS) on S6 D2. Error appears when attempting to format new volume on S6 D2.

Best Regards

tomcw:
Reproduced with AppleWin 1.26.3.0:
- Boot prodos_utility.dsk (which is ProDOS 1.1.1)
- 'F' (ProDOS Filer)
- 'V' (Volume commands)
- 'F' (Format a volume)
- Create a blank (empty, zero byte) blank.dsk image into drive-2
- Slot-6, Drive-2, any volume name and get a "DISK II DRIVE TOO SLOW" error

tomcw:
Here's my (long!) analysis and proposed solution for this issue...

ProDOS's FORMATTER writes tracks as:

16 sectors of:
. Gap1(sector0) or Gap3(sector1-15)
. Address Field
. Gap2
. Data Field

Gap1 and Gap2 are fixed sizes of 129 and 5 lots of 40-cycle 0xFF sync bytes.
Gap3 starts as a large size of 30 x 40-cycle 0xFF sync bytes.
If the track just written can't be read back successfully (eg. due to Gap3 being too big and the track overwriting itself), then Gap3 is reduced by 1 and the the track is re-written.
NB. There can be a small bit of track overwriting as up to ~0x7C bytes of Gap1 could be overwritten leaving ~5 0xFF sync bytes before sector0's Address Field.

If the track just written can be read back successfully, but Gap3 >= 0x1B+1(*1) then the FORMATTER will return the error: "DISK II DRIVE TOO SLOW"

```
; Succeeded writing track, now check gap3_count tolerance
79B5-   A4 D4       LDY   $D4		; w_Gap3Count
79B7-   CC 1F 7D    CPY   $7D1F		; kMinGap3 (=0x0E)
79BA-   B0 05       BCS   $79C1
79BC-   A9 04       LDA   #$04		; 
79BE-   4C F9 79    JMP   $79F9		; error code = 4 (Drive too FAST)
;
[...]

Issue #338, opened 2016-09-05
Ultima I Disk Drive speed error


BPaden:
I'm not hugely experienced with the Apple II, so this could be a user error. I'm also running this with wine, so this could be a bug in wine. But when I boot up Ultima I - The First Age of Darkness and try to format a player disk, I get a `Disk drive speed error`. I've tried with the Apple II+ and the Apple IIe (the game doesn't seem to work at all with the original apple II), and I have tried using the authentic disk access speed, all to no effect.

TommyGH:
This is similar to:
  Cannot format Pascal Disk #196

In #196 the error message is:  
  Drive speed is too slow.

Fyi,
Tommy

tomcw:
From PR #507:
 .dsk & .nib images now format OK for both authentic/enhanced disk access speeds.
(.nib needs a zero-length image to be initialised with alternating values, eg. 0x80,0x81.)

Just like #196, before calling the formatter routine, it will try to determine if the disk is already formatted and has a subroutine (albeit slightly different to #196) that reads the drive latch until the nibble changes:

```
D4DA:A6 3E                LDX $3E    
D4DC:A0 00                LDY #$00   
D4DE:BD 8C C0             LDA $C08C,X
D4E1:20 F0 D4             JSR $D4F0   ;  12cy / RTS
D4E4:48                   PHA         ;  3
D4E5:68                   PLA         ;  4
D4E6:DD 8C C0             CMP $C08C,X ;  4
                                      ;  = 23 cycles
D4E9:D0 05                BNE $D4F0  
D4EB:A9 28                LDA #$28   
D4ED:88                   DEY        
D4EE:D0 EE                BNE $D4DE  
D4F0:60                   RTS        
```

(Again the delay between LDA and CMP is 23 cycles.)

NB. The formatter routine is just the standard ProDOS formatter routine (see my analysis in #125), with just this difference:
- code org'ed at $A800-$AC1E; data $AC1F-$AC39

tomcw:
From the file: "Beneath Apple ProDOS 8 - Supplement for Version 1.1.1.pdf", page 67:

![prodos-datachanging](https://user-images.githubusercontent.com/6696896/32752992-251fff00-c8c3-11e7-842d-87b48a85ede2.PNG)

tomcw:
Closing as work is complete.

Issue #196, opened 2014-06-04
Cannot format Pascal Disk


tomcw:
**BerliOS Bug #19368**

**Date:** 2014-Jun-03 17:15
**Submitted By:** tbraun3
**Category:** 1.24.0
**Bug Group:** none
**Assigned To:** none
**Priority:** 5
# 

While attempting to format a new disk in Apple Pascal (1.1 or 1.2) an error occurs: "DRIVE SPEED IS TOO SLOW". It doesn't matter if it is already formatted in another OS or not. 

I can get the same message if I try to format a new disk in ProDOS.

I also ran the speed test on Central Point Software's "Filer Utility (1982). It reported the speed of the drives to be 125.07 ms. whether they are set to authentic or enhanced speed. The recommended speed is between 197.00 and 203.00 ms.

tomcw:
From PR #507:
.dsk images now format OK for both authentic/enhanced disk access speeds.
.nib images get reported as: "UNIT NUMBER 5 IS NOT ONLINE" (both authentic/enhanced).

tomcw:
>.nib images get reported as: "UNIT NUMBER 5 IS NOT ONLINE" (both authentic/enhanced).

Before calling the formatter routine, it will try to determine if the disk is already formatted and eg. ask if you want to "DESTROY DIRECTORY OF BLANK ?" or "DESTROY DIRECTORY OF &LT;volume name&GT; ?"

This "_determine if the disk is already formatted_" code has a subroutine that reads the drive latch until the nibble changes:
```
D382:A0 00                LDY #$00   
D384:BD 8C C0             LDA $C08C,X
D387:20 94 D3             JSR $D394	; 12cy / RTS  
D38A:48                   PHA        	;  3
D38B:68                   PLA        	;  4
D38C:DD 8C C0             CMP $C08C,X	;  4
					;  = 23 cycles
D38F:D0 03                BNE $D394  
D391:88                   DEY        
D392:D0 F0                BNE $D384  
D394:60                   RTS        
```

(I don't know why the delay between LDA and CMP is 23 cycles.)

For this reason, the zero-length .nib can't just be `memset()` to a fixed value (eg. 0x00 or 0x80). But setting to alternating values (0x80, 0x81) works.

NB. The formatter routine is just the standard ProDOS formatter routine (see my analysis in #125), with these differences:
- code org'ed at $3D00-$411F (data $411F-$4139)
- gap3 min is 0x10 (instead of 0x0E)
[...]

PR #507, opened 2017-11-08
ProDOS disk format

tomcw:
Proposed fix for #125. Support:
- .dsk / .nib images
- ProDOS format / DOS 3.3 init
- authentic / enhanced disk access speed


tomcw:
Hi @sicklittlemonkey and @Michaelangel007:
Nick, I'm mainly after your opinion on this PR, as it's all about the disk ][ emulation.

Initially I started this to fix the ProDOS format issue (#125), but extended the work to cover DOS 3.3 INIT... for .dsk & .nib images; and for both authentic & enhanced drive emulation.

The TL;DR story is [here](https://github.com/AppleWin/AppleWin/issues/125#issuecomment-342977995).
And the short version is:
> My current working solution is to monitor the written bytes, detect when all 16 sectors have been written (ie. a full track), then to clamp the track length to 0x18B0. I do this by keeping the last 0x18B0 bytes, and zeroing the remaining bytes.

What's the risk? (ie. could this introduce regressions?)
Well I don't think there's any, and it will only affect Apple II code that formats tracks.
NB. Writing 16 sectors to a pre-formatted track is **_not_** the same thing, as the sector's address field is only written during track formatting.

Testing has been done with DOS 3.3 INIT, ProDOS formatting, Pascal formatting (#196), Ultima I formatting (#338), Copy II+ v9.1 formatted (DOS 3.3 & ProDOS).

All the new code has been moved to a new DiskFormatTrack.cpp and corresponding FormatTrack class.
Disk.cpp has been slightly extended by adding a static instance of FormatTrack, called g_formatTrack. NB. There are just 9 places in Disk.cpp where there are calls to methods in g_formatTrack.

I've moved the LOG_DISK macros to DiskLog.h (as it's used by both Disk.cpp and DiskFormatTrack.cpp). And I've extended your logging, so you get nice extras like:

On a sector read:
```
read D5AA96 detected - Vol:FE Trk:00 Sec:01 Chk:FF
```

On a sector write (during a track write):
```
write D5AA96 detected - Vol:FE Trk:00 Sec:07 Chk:F9
```

On a track write:
```
Gap1 = 0x102, Gap2 = 0x08, Gap3 = 0x0B (11), TrackSize = 0x18E7
```

So if you get time can you take a quick look? And let me have any comments.
btw. when I merge, I will just do a squash-merge, so it becomes just a single commit.

Thanks.

tomcw:
re. LOG_DISK macros

I forgot this, when logging individual disk nibble writes you now get the cycle count (useful to see sync FF's):
```
write 1676 = FF (cy=+40) sync #125
write 1677 = FF (cy=+40) sync #126
write 1678 = FF (cy=+40) sync #127
write 1679 = FF (cy=+40) sync #128
write 167A = D5 (cy=+32)
write 167B = AA (cy=+32)
write 167C = 96 (cy=+32)
```

tomcw:
NB. Only the VS2008 AppleWin.vcproj is updated with the new files. So you'll have to add DiskFormatTrack.cpp if you are using VS >2008.
If/when the PR is OK'd then I'll fix-up all the other VS proj files.

tomcw:
Merged to master at 4a69ba8. Closing.

PR #517, opened 2017-12-13
Comments after static analysis

qweassaewq:


tomcw:
@qweassaewq - which static analysis tool did you use?

NB. I won't be fixing external libs (eg. zip_lib), as we'd have local patches that we'd need to maintain when we upgrade to newer version of that open-source lib.

source/Memory.cpp
```diff
@@ -1260,6 +1260,7 @@ void MemInitialize()
 	MemInitializeCustomF8ROM();
 	MemInitializeIO();
 	MemReset();
+	//The function was exited without releasing the pointer / handle  'newloc'.A memory / resource leak is possible.
```
tomcw: There's no leak here.
Initially the pages are just reserved:
```
memimage = (LPBYTE)VirtualAlloc(NULL,_6502_MEM_END+1,MEM_RESERVE,PAGE_NOACCESS);
```

Then later they are committed, with a test that the addresses match:
```
LPVOID newloc = VirtualAlloc(memimage,_6502_MEM_END+1,MEM_COMMIT,PAGE_READWRITE);
if (newloc != memimage)
{
    // ...
```

source/Tfe/Tfe.cpp
```diff
@@ -597,7 +597,7 @@ void tfe_init(void)
 
 void tfe_shutdown(void)
 {
-    assert( (tfe && tfe_packetpage) || (!tfe && !tfe_packetpage));
+    assert( (tfe && tfe_packetpage) || (!tfe && !tfe_packetpage)); // An excessive check can be simplified. The '(A && B) || (!A && !B)' expression is equivalent to the 'bool(A) == bool(B)' expression.
```
tomcw: I'm not concerned about this, since it's harmless and it's external library code.

source/Tfe/Tfearch.cpp
```diff
@@ -267,8 +267,8 @@ BOOL TfePcapOpenAdapter(const char *interface_name)
             TfePcapDevice = TfePcapAlldevs;
         }
     }
-
-    TfePcapFP = (*p_pcap_open_live)(TfePcapDevice->name, 1700, 1, 20, TfePcapErrbuf);
+	
+    TfePcapFP = (*p_pcap_open_live)(TfePcapDevice->name, 1700, 1, 20, TfePcapErrbuf); // There might be dereferencing of a potential null pointer 'TfePcapDevice'.
```
tomcw: TfePcapOpenAdapter() will set `TfePcapDevice = TfePcapAlldevs` if specified adaptor is not found (or NULL is passed).

tfe_arch_enumadapter_open() inits TfePcapAlldevs. If no device is found then it returns an error.

So I'm not going to do anything about this.

source/Tfe/Tfesupp.cpp
```diff
@@ -106,7 +106,7 @@ char *lib_stralloc(const char *str)
 
     size = strlen(str) + 1;
     ptr = (char *)lib_malloc(size);
-
+	//The potential null pointer is passed into 'memcpy' function.Inspect the first argument.
```
tomcw: The first arg to memcpy is 'ptr', which gets initialised by lib_malloc().
lib_malloc() calls `exit(-1)` if malloc() fails.

This looks OK to me.

source/Z80VICE/z80.cpp
```diff
@@ -6429,7 +6429,8 @@ void z80_WRMEM(WORD Addr, BYTE Value)
 		case 0xE000: addr = laddr+0xC000; break;
 		case 0xF000: addr = laddr+0x0000; break;
 	}
-	CpuWrite( addr, Value, ConvertZ80TStatesTo6502Cycles(maincpu_clk) );
+	//Potentially uninitialized variable 'addr' used.Consider checking the first actual argument of the 'CpuWrite' function.
+	CpuWrite(addr, Value, ConvertZ80TStatesTo6502Cycles(maincpu_clk));
```
tomcw: This is OK - visually you can see that the switch statement covers all cases.

zip_lib/unzip.c
```diff
@@ -492,6 +492,7 @@ extern unzFile ZEXPORT unzOpen2 (path, pzlib_filefunc_def)
 
     s=(unz_s*)ALLOC(sizeof(unz_s));
     *s=us;
+	// There might be dereferencing of a potential null pointer 's'. Check lines: 494, 493.
```
tomcw: As stated, I'm not changing any external libs.

zip_lib/zip.c
```diff
@@ -778,7 +778,7 @@ extern int ZEXPORT zipOpenNewFileInZip3 (file, filename, zipfi,
 
     ziplocal_putValue_inmemory(zi->ci.central_header,(uLong)CENTRALHEADERMAGIC,4);
     /* version info */
-    ziplocal_putValue_inmemory(zi->ci.central_header+4,(uLong)VERSIONMADEBY,2);
+    ziplocal_putValue_inmemory(zi->ci.central_header+4,(uLong)VERSIONMADEBY,2); //The 'zi->ci.central_header' pointer in the 'zi->ci.central_header + 4' expression could be nullptr. In such case, resulting value will be senseless and it should not be used. Check lines: 781, 777.
```
tomcw: As stated, I'm not changing any external libs.

tomcw:
Thanks for raising this PR.

I've fixed all the issues that I agree with, so I'm closing this PR.

@qweassaewq - which static analysis tool did you use?

PR #524, opened 2017-12-28
Stereo audio chain in Speaker.cpp, emulate 4 DAC SAM card in SAM.cpp

rmacri:
Refactored Speaker.cpp so the audio chain is stereo. The Apple speaker remains mono and centred. This enabled me to emulate the 4 DAC MEGAAudio S.A.M. clone card and progress my Drum8 software to support mixing 8 samples at once, on both a normal S.A.M. or a multi-DAC stereo configuration.
I recognise this is a very specific case for a card not around in the Apple II
era (AFAIK), but thought you might find the stereo sample code useful if you ever
consider converting the Mockingboard code to be clock locked.
I could progress with cleanups, eg: some optimisation, adding a UI option for the SAM DAC type instead of an #ifdef in SAM.cpp, if there is interest to merge this. The Speaker changes are independent of SAM (mostly) and included a STEREO #ifdef for fallback to mono, this should/can be removed if the new code is adopted.


tomcw:
Hi Riccardo,

Is the "4 DAC MEGAAudio S.A.M. clone card" something you or someone else created? Is there any reference info about it?

(Perhaps I should create an issue to reference this PR, so that if someone searches the issues they can find this PR.)

I'd say this is very specific, and as such currently I'm not so keen to merge this in. But let's keep this PR open.

Thanks.

tomcw:
NB, commit 71c67cf supports inserting SAM cards into both slot 4 and 5 at the same time.
Not the same as this PR, but maybe sufficient?

PR #539, opened 2018-02-13
Changes to ease code compilation in gcc.

audetto:
Disk_t and HDD contain a std::string and for this reason they need a proper constructor, ZeroMemory is not guaranteed to work.
Moreover the operator= does not seem to be required as the compile generated one is good.

The only less than trivial change is exactly in Disk_t and HDD. I've tested inserting and swapping FloppyDisks and it all works well (gcc and VS2017).
I've realised one thing as well. If I swap while one drive is spinning, then which drive should be spinning after the swap? Is spinning a property of the disk or the drive?


tomcw:
>I've realised one thing as well. If I swap while one drive is spinning, then which drive should be spinning after the swap? Is spinning a property of the disk or the drive?

It's a property of the drive, so this is a (current) bug in AppleWin.

tomcw:
I agree that ZeroMemory() shouldn't be used on these structs, but I think all that's needed is a call to the dtor and then reconstruct by calling the ctor via placement new for the 2 array elements. I'd like to explore this instead of using std::vector.

Why? Well the only issue is the ZeroMemory() in the LoadSnapshot() - for both disk.cpp and hdd.cpp. So it seems simpler to just fix these single ZeroMemory() calls, than the more invasive replacement with std::vector.

tomcw:
OK, I tried replacing the ZeroMemory() with dtor+placement new, eg:
```
		ZeroMemory(&g_aFloppyDisk[i], sizeof(Disk_t));
```
changed to:
```
		g_aFloppyDisk[i].~Disk_t();
		new (&g_aFloppyDisk[i]) Disk_t();
```

But as you have discovered there are 4 places where ZeroMemory() is called on this non-POD struct. So these 2 lines (ctor+new) should be moved to a dedicated function.

btw. for your call to std::vector<type>::operator[].clear()
- doesn't this erase the elements of the whole vector? (ie. not just that element)
- the capacity may remain at 2, but could shrink (implementation dependent)
- so maybe accessing again with operator[] could lead to undefined behaviour?
- NB. it appears that operator[] is not bounds-checked (see [here](url))

audetto:
clear() is a function on the struct Disk_t, it does not touch the vector, which is resized once and for all to NUM_DRIVES (exactly like the c array used to be).

It could have been called resetDisk or zeroDisk to avoid confusion.

To change the size of the vector (i.e. shrink) one needs to call something like, resize, erase, clear on the vector itself (or reassign it), which is never done.
The vector will be destroyed at shutdown when the "finaliser" of the module is invoked and the vector destructor called.

I don't think that replacing ZeroMemory with destructor+constructor would work, unless the class has been already created (which does not happen in a c array). The first time, one must not call the destructor.

audetto:
#include <cstring>
#include <string>
#include <iostream>

struct Data
{
  std::string str;
};

int main()
{
  Data d;

  memset(&d, 0, sizeof(Data));  // BAD
  //new (&d.str) std::string();   // GOOD
  d.str = "";
  std::cout << d.str << std::endl;
}

If I run this code in gcc, it with crash in the assignment = "".
If I use the placement new, then it works, but I think this is an order of magnitude more complex that using the std::vector which deals with all of this internally.

If one is worried that std::vector could be resized, maybe using a std::array<Disk_T, NUM_DRIVES> is an alternative.
This is similar to both and it will take care of properly initialising the C++ classes of Disk_t, but then it cannot be resized.

tomcw:
>clear() is a function on the struct Disk_t

Ah! So it is. I was eye-balling the code, so just read it as std::vector<>::clear()!
Yes, better to change to some better named function as you suggest.

>I don't think that replacing ZeroMemory with destructor+constructor would work, unless the class has been already created (which does not happen in a c array). The first time, one must not call the destructor.

Actually the ctor for std::string is called on each array element for g_aFloppyDisk[] at start-up (ie. before main() is called). You can see this by setting a breakpoint on the global variable and then stepping the assembly. I've done this in both VC++ (VS2015) and clang (v5.0.0). So it _is_ OK to explicitly call the dtor on these array elements.

>If I use the placement new, then it works, but I think this is an order of magnitude more complex that using the std::vector which deals with all of this internally.

Well, probably not that more complex(!), but I certainly agree with you that calling an explicity dtor + placement new is less readable/maintainable than just using a good old std::vector.

tomcw:
btw. since AppleWin is still using VS2008, then we can't use std::array, as this is C++11.
But as I said above, std::vector is fine.

tomcw:
Closing as superseded by #541.

PR #541, opened 2018-02-21
Changes to ease code compilation in gcc. V4

audetto:
Disk_t and HDD contain a std::string and for this reason they need a proper constructor, ZeroMemory is not guaranteed to work.

In this version there is no need to have a std::vector, as a c array actually initialises all items.


tomcw:
Great! Thanks for identifying this simplification (remaining with C-style arrays).

This LGTM, but I will give it some runtime testing before merging it into master.

tomcw:
A minor thing: why have you added HD_Insert() to the external interface in Harddisk.h?

audetto:
It is committed now.
But HD_Insert is needed by the Qt frontend, strictly speaking not to compile in gcc.
Problem is that there is no way, other than calling HD_Select() to insert an image.

HD_select is OS specific and open a dialog window.

It is one of these cases where there is emulator code mixed with frontend - os specific code.

Andrea

tomcw:
>But HD_Insert is needed by the Qt frontend

I see... OK, I'll add a comment so that it's not removed from Harddisk.h.

NB. Configuration's Property Sheet Page ("Disk") calls HD_SelectImage() to pick the HDD image, and then internally calls HD_Insert().

Issue #555, opened 2018-05-12
One pixel shifted to the right?


landloafer:
I just saw the title screen screen of prince of persia (DHGR) 
![apple_000](https://user-images.githubusercontent.com/39210397/39954334-3f8f5e0c-55f0-11e8-851c-990f0125051b.png)
with APPLEPC2.52 (dos). I noticed there is an extra line at the right most. But can't verify which one is correct.

landloafer:
![applewin_screenshot_000000002](https://user-images.githubusercontent.com/39210397/39954467-a1e26ef8-55f2-11e8-9f76-bddeda0446ad.png)
1.27.2.0

landloafer:
![applewin_screenshot_000000001](https://user-images.githubusercontent.com/39210397/39954479-cd212794-55f2-11e8-817b-a9e00ed7173a.png)
1.27.2.0

tomcw:
Thanks for reporting this. Yes, it certainly looks like a bug introduced after 1.25.04 (ie. 1.26+).
I'll take a look at a fix.

tomcw:
>While investigating this, I noticed there is a bug in AppleWin's NTSC.cpp, `updateVideoScannerHorzEOL()` ...

Fixed in 3e7cc36.

landloafer:
Thanks!
Is it the whole screen shifted one pixel?

tomcw:
>Is it the whole screen shifted one pixel?

Yes... but only DHGR, so in mixed mode there's a 1 pixel misalignment between DHGR (top 160 lines) & the bottom 32 lines of TEXT/TEXT80/GR/DGR etc.

Here's a zoomed display with the dash-line being the 159th DHGR line, and then TEXT80 below. It demonstrates the misalignment: (NB. screenshot from Sheldon's AppleWin-NTSC-05 build, but AppleWin 1.26/1.27 will be the same.)
![ntsc05-dhgr-mixed-zoomed](https://user-images.githubusercontent.com/6696896/40140608-48c3b348-594b-11e8-9fc5-038b4ea2892b.png)

landloafer:
Interesting read! So anticipated to next release.

Michaelangel007:
Tom, thanks for the fixing that. I guess I should have done more testing -- I didn't even think to try DHGR monochrome!

tomcw:
Here's the HCOLOR vertical  line before the fix:
[...]

PR #577, opened 2018-08-18
Extend video by 1 pixel for 14M video modes

tomcw:
This is for #555.


tomcw:
Testing shows this is not ready:
- Mixed GR (or HGR) with 80COL isn't clearing the RHS (561th) pixel for the GR (or HGR) top part.
- TEXT (in 'Monitor Color' mode) for eg. INVERSE ']' (col-39) draws an extra pixel on the RHS
  - I suspect that the StretchBlt() operation that is slightly different for the different video modes (eg. 'Monitor Color', etc) is the cause of this...
  - Perhaps start the rendering earlier (eg. for Monitor/TV Color) so that the StretchBlt() doesn't need special cases.

tomcw:
Manually did a merge --squash to master at 7d100a3

Issue #408, opened 2017-04-27
Memory: Support Saturn 128K Language Card


Michaelangel007:
I've stubbed in the bare essentials for the Saturn 128K Language Card. I've also added a new global to track the type of extended language card:

There is now [g_eMemType](https://github.com/AppleWin/AppleWin/blob/master/source/Memory.h#L49) of type `MemoryType_e` in [Memory.h](https://github.com/AppleWin/AppleWin/blob/master/source/Memory.h#L42)

* `MEM_TYPE_NATIVE`, set by default
* `MEM_TYPE_RAMWORKS`, set if `-r` command line is used
* `MEM_TYPE_SATURN`, set if `-saturn` command line is used

The hard part is the actual memory paging.

What's the correct way to go about updating the memory paging tables?  Add `#ifdef SATURN` to `UpdatePaging()` and copy memory to/from the Saturn 16K memory banks in the $D0 and $E0 loops?

```c
	for (loop = 0xD0; loop < 0xE0; loop++)
	{
		int bankoffset = (SW_BANK2 ? 0 : 0x1000);
		memshadow[loop] = SW_HIGHRAM ? SW_ALTZP	? memaux+(loop << 8)-bankoffset
												: memmain+(loop << 8)-bankoffset
									: memrom+((loop-0xD0) * 0x100);

		memwrite[loop]  = SW_WRITERAM	? SW_HIGHRAM	? mem+(loop << 8)
														: SW_ALTZP	? memaux+(loop << 8)-bankoffset
																	: memmain+(loop << 8)-bankoffset
										: NULL;
	}

	for (loop = 0xE0; loop < 0x100; loop++)
	{
		memshadow[loop] = SW_HIGHRAM	? SW_ALTZP	? memaux+(loop << 8)
													: memmain+(loop << 8)
										: memrom+((loop-0xD0) * 0x100);

		memwrite[loop]  = SW_WRITERAM	? SW_HIGHRAM	? mem+(loop << 8)
														: SW_ALTZP	? memaux+(loop << 8)
																	: memmain+(loop << 8)
										: NULL;
	}

```

tomcw:
Michael: commit 248c14165520177341bf7f3282738cb58b081654 introduced this warning:
source\memory.cpp(1709) : warning C4102: '_done_saturn' : unreferenced label

tomcw:
>What's the correct way to go about updating the memory paging tables?

Did you figure this our, or is this question still outstanding?

[...]

PR #589, opened 2018-10-12
Apple II: Language Card and Saturn support for slot-0

tomcw:
PR for #408.

Apple II and Apple II+ support for a Language Card or Saturn 128K in slot-0.
NB. Not for Apple //e and above.

Includes full save-state support.
Implementation has been moved to new classes in a new LanguageCard.cpp.

### DIFF
--- a/AppleWinExpress2008.vcproj
+++ b/AppleWinExpress2008.vcproj
@@ -626,6 +626,14 @@
 					>
 				</File>
 				<File
+					RelativePath=".\source\LanguageCard.cpp"
+					>
+				</File>
+				<File
+					RelativePath=".\source\LanguageCard.h"
+					>
+				</File>
+				<File
 					RelativePath=".\source\Log.cpp"
 					>
 				</File>
--- a/AppleWinExpress2013.vcxproj
+++ b/AppleWinExpress2013.vcxproj
@@ -71,6 +71,7 @@
     <ClInclude Include="source\Harddisk.h" />
     <ClInclude Include="source\Joystick.h" />
     <ClInclude Include="source\Keyboard.h" />
+    <ClInclude Include="source\LanguageCard.h" />
     <ClInclude Include="source\Log.h" />
     <ClInclude Include="source\Memory.h" />
     <ClInclude Include="source\Mockingboard.h" />
@@ -154,6 +155,7 @@
     <ClCompile Include="source\Harddisk.cpp" />
     <ClCompile Include="source\Joystick.cpp" />
     <ClCompile Include="source\Keyboard.cpp" />
+    <ClCompile Include="source\LanguageCard.cpp" />
     <ClCompile Include="source\Log.cpp" />
     <ClCompile Include="source\Memory.cpp" />
     <ClCompile Include="source\Mockingboard.cpp" />
--- a/AppleWinExpress2013.vcxproj.filters
+++ b/AppleWinExpress2013.vcxproj.filters
@@ -76,6 +76,9 @@
     <ClCompile Include="source\Keyboard.cpp">
       <Filter>Source Files\Emulator</Filter>
     </ClCompile>
+    <ClCompile Include="source\LanguageCard.cpp">
+      <Filter>Source Files\Emulator</Filter>
+    </ClCompile>
     <ClCompile Include="source\Log.cpp">
       <Filter>Source Files\Emulator</Filter>
     </ClCompile>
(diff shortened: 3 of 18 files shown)


tomcw:
Hi Michael - if you get a chance then take a quick look at LanguageCard.cpp.

Basically I've taken the work you did in branch "saturn_langcard", moved it to this new file and added save-state support.

Thanks.

tomcw:
Hi Michael,
In the end I settled on this class hierarchy:

```
LanguageCardUnit
 ^
 |
LanguageCardSlot0
 ^
 |
Saturn128K
```

IE. 
- LanguageCardSlot0 _IS A_ LanguageCardUnit
- Saturn128K _IS A_ LanguageCardSlot0

AFAIAC this is good enough, but LC & Saturn are fixed to slot-0; and neither can be used in a //e (or above).

The "no support for a Saturn in a //e" is because (a) there's no slot-0; but also the Saturn's 16K RAM would compete with the //e's high 16K RAM/ROM on the bus. So just the same as a real //e, the Saturn card should not be used in one.

I'll do a bit more testing, and probably squash & merge in the next few days.

tomcw:
>The "no support for a Saturn in a //e" is because (a) there's no slot-0; but also (b) the Saturn's 16K RAM would compete with the //e's high 16K RAM/ROM on the bus. So just the same as a real //e, the Saturn card should not be used in one.

This (b) part is probably not correct.

From Sather (UTAIIe 5-24), "When INHIBIT' is low in the Apple IIe, all mobo RAM, aux RAM and mobo ROM is disabled, including high RAM."

This differs from the A][, in that the (slot-0) "16K RAM card will not release the $F800-$FFFF address range under any circumstances..."

NB. Peripheral slot connection pin 32 is INHIBIT'.

So a Saturn in a //e (slots 1-7, since n slot-0) can be used, so long as it pulls INHIBIT' low.
Anyway, currently not a problem as I'm only supporting slot-0 for now.

tomcw:
How does a Saturn respond to RESET?

EG. Here ([Saturn 128k Problems](https://groups.google.com/forum/#!topic/comp.sys.apple2/ypcKjYnHEQY)), the poster says RAMTEST128K fails and then "Interestingly, the machine does not respond to a Ctrl-Reset (!) and must be power-cycled."

From my experience(*), after a RAMTEST128K failure, it leaves HIGHRAM enabled, and a RESET doesn't reset the Saturn's state, so the 6502 vectors to ($FFFC) which will be same RAMTEST value... ie. causing a crash/hang.

(*) My Saturn emulation behaves like the 16K LC and RESET doesn't disable RAM (ie. enable ROM).

Michaelangel007:
Sorry, I'm late.  Looks good !

Now I'm curious what changes were made to get Fantavision to work ! :-)

tomcw:
>Now I'm curious what changes were made to get Fantavision to work ! :-)

In Memory.cpp, UpdatePaging(), you are missing a few lines of code to set the memwrite[] for the $E000-$FFFF range, ie, the code below between "MISSING" and "MISSING-END":

```
	for (loop = 0xE0; loop < 0x100; loop++)
	{
		memshadow[loop] = SW_HIGHRAM	? SW_ALTZP	? memaux+(loop << 8)
#ifdef SATURN
													: g_uSaturnTotalBanks
														// Sat.  Mem.          Saturn     L.C.
														// 0000  D000  Slot N  Bank A  =  Bank 2
														// 1000  D000  Slot N  Bank B  =  Bank 1
														// 2000  E000  Slot N
														// 3000  F000  Slot N
														? g_aSaturnPages[ g_uSaturnActivePage ] + ((loop - 0xC0) << 8)
#endif // SATURN
														: memmain+(loop << 8)
										: memrom+((loop-0xD0) * 0x100);

		memwrite[loop]  = SW_WRITERAM	? SW_HIGHRAM	? mem+(loop << 8)
														: SW_ALTZP	? memaux+(loop << 8)
#ifdef SATURN // MISSING
																	: g_uSaturnTotalBanks
																	? g_aSaturnPages[g_uSaturnActivePage] + ((loop - 0xC0) << 8)
#endif // SATURN // MISSING-END
																	: memmain+(loop << 8)
										: NULL;
	}
```

Because this was missing, then any writes to RAM above $E000 would go to memmain (I think!) and not your g_aSaturnPages[0] memory. This may only be an issue when read & write are different (eg. read ROM / write RAM). The Saturn System RAMTEST128K presumably pass as read & write are the same (eg. read RAM / write RAM).

PR #623, opened 2019-02-22
Do not reset on minor state load errors

rzumer:
The state load function resets the machine too aggressively in my opinion. Occasionally I press F12 instead of F11 to save a state, and after specifying an invalid file name get an error and the system resets, despite no configuration being set. For those minor errors that occur at the start of the state load process, instead of throwing an exception, this displays the error message and cancels the state load.

### DIFF
--- a/source/SaveState.cpp
+++ b/source/SaveState.cpp
@@ -358,13 +358,28 @@ static void Snapshot_LoadState_v2(void)
 {
 	try
 	{
+		std::string err_msg = "";
 		int res = yamlHelper.InitParser( g_strSaveStatePathname.c_str() );
 		if (!res)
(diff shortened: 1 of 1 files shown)


tomcw:
Thanks for the PR.

tomcw:
I've changed the patch to something simpler and IMO more consistent.

Details: `InitParser()` and `ParseFileHdr()` can both throw. So we had this odd mix of throw (and restart) or no-throw (and no restart). The restart doesn't need to happen in either case.

So I've reverted to nearly what the code was before, but added a bool to control whether to restart or not.

rzumer:
Seems good, thanks.

Issue #559, opened 2018-06-05
Drop support for VS2010 and VS2012?


tomcw:
In PR #339, [here](https://github.com/AppleWin/AppleWin/pull/339#issuecomment-285855059), I considered which versions of VS support which OS's:

---

I'm considering VS from the point of view of AppleWin supporting older Windows OS versions.
(Not that I do any testing on OS's older than Win7-64, but rely on end-user feedback to highlight issues.)

VS2005 - last VS version to support Win98/ME
VS2008 - last VS version to support Win2000: [ref](https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#2008)
VS2010 - supports WinXP and above
VS2012 - dropped support for WinXP
VS2012-update1 added support back in for WinXP
VS2013/15/17 - do these still include support for WinXP targets?

At AppleWin 1.26 we moved from VS2005 to VS2008, and so dropping support for Win98/ME.

I'm fine with sticking with VS2008 (for the official release build) for now, but it's useful to understand the implications of moving to a newer VS version.

If in the future we moved to a newer VS, then implicitly that means Win2000 is no longer supported by AppleWin (probably fine, but we should ask on the cea2 newsgroup). Now WinXP becomes the new base OS, so what is the minimum VS version that still supports WinXP?

---

Yes, VS2013/15/17 does have XP support. (Never used myself though.)
The corresponding Platform toolset settings are v120_xp, v140_xp, and v141_xp.
For comparison, the default settings without XP support are v120, v140, and v141.

---

"Configuring Programs for Windows XP" https://msdn.microsoft.com/en-us/library/jj851139.aspx

That's for 2015 but 2017 still has the same option. BTW, they've optimized 2017 to load quicker than 2015.

tomcw:
Anyway, VS2012 hasn't been updated for 3 years (May 2015).

But I can see that Michael still updates VS2010 periodically. NB. I don't since I don't have it installed (and have no desire to install it!).

I propose we drop VS2012 from the repo.
And also VS2010 too... but this is up to Michael, who looks to still be using it.

tomcw:
@Michaelangel007 - can we drop VS2010 support yet?

...and close PR #442 too (which is for V2010).

tomcw:
NB. I've moved all VS2005 & VS2012 proj & sln files.
Just VS2010 still to decide on. . .

tomcw:
VS2019 was released earlier this week, but I don't know if it has a Platform toolset for XP.

[...]

PR #634, opened 2019-04-02
Use default initializers for local variables

rzumer:
Fixes [type.6 type safety warnings](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-always) (C++ Core Guidelines), generated via code analysis in Visual Studio 2019.

Some of the changes are redundant because instead of default constructors, `Clear()` functions or similar are used to initialize new instances. However the performance cost of using default initializers should be very small, and this ensures that local variables are initialized on instantiation even when they are not handled by the aforementioned clean-up functions.

It is probably OK to remove the calls to the clearing functions from the default constructors that employ them with these changes, but I preferred to err on the side of caution. If all looks good I can amend this PR to remove the calls.


tomcw:
>Fixes type.6 type safety warnings (C++ Core Guidelines), generated via code analysis in Visual Studio 2019.

That was quick - VS2019 was only released on Tuesday!

This syntax works on VS2015 & VS2017... but AppleWin is still officially built with VS2008, and it doesn't work with this old compiler. So first we need a discussion about moving to a more recent version of VS.

rzumer:
Thanks for checking. That's worth discussing if VS2008 is not critical for supporting certain platforms, but I will try a different approach that is backwards-compatible instead.

tomcw:
>That's worth discussing if VS2008 is not critical for supporting certain platforms

See #559

tomcw:
I don't see the point of adding this C++11 initializer list feature suffix to these areas of the code. IMO, this C++11 syntax is obscure and therefore confusing. So I am rejecting this PR, and will close it.

PR #637, opened 2019-04-10
Refactor Disk.cpp

tomcw:
Refactored Disk.cpp into a class.

### DIFF
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -105,7 +105,7 @@ int			g_nMemoryClearType = MIP_FF_FF_00_00; // Note: -1 = random MIP in Memory.c
 IPropertySheet&		sg_PropertySheet = * new CPropertySheet;
 CSuperSerialCard	sg_SSC;
 CMouseInterface		sg_Mouse;
-DiskIIInterfaceCard sg_DiskIICard;
+Disk2InterfaceCard sg_DiskIICard;
 
 SS_CARDTYPE g_Slot0 = CT_LanguageCard;	// Just for Apple II or II+ or similar clones
 SS_CARDTYPE g_Slot4 = CT_Empty;
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -105,7 +105,7 @@ int			g_nMemoryClearType = MIP_FF_FF_00_00; // Note: -1 = random MIP in Memory.c
 IPropertySheet&		sg_PropertySheet = * new CPropertySheet;
 CSuperSerialCard	sg_SSC;
 CMouseInterface		sg_Mouse;
-Disk2InterfaceCard sg_DiskIICard;
+Disk2InterfaceCard sg_Disk2Card;
 
 SS_CARDTYPE g_Slot0 = CT_LanguageCard;	// Just for Apple II or II+ or similar clones
 SS_CARDTYPE g_Slot4 = CT_Empty;
@@ -270,7 +270,7 @@ static void ContinueExecution(void)
 	const bool bWasFullSpeed = g_bFullSpeed;
 	g_bFullSpeed =	 (g_dwSpeed == SPEED_MAX) || 
 					 bScrollLock_FullSpeed ||
-					 (sg_DiskIICard.IsConditionForFullSpeed() && !Spkr_IsActive() && !MB_IsActive()) ||
+					 (sg_Disk2Card.IsConditionForFullSpeed() && !Spkr_IsActive() && !MB_IsActive()) ||
 					 IsDebugSteppingAtFullSpeed();
 
 	if (g_bFullSpeed)
@@ -317,7 +317,7 @@ static void ContinueExecution(void)
 	const DWORD uActualCyclesExecuted = CpuExecute(uCyclesToExecute, bVideoUpdate);
 	g_dwCyclesThisFrame += uActualCyclesExecuted;
 
-	sg_DiskIICard.UpdateDriveState(uActualCyclesExecuted);
+	sg_Disk2Card.UpdateDriveState(uActualCyclesExecuted);
 	JoyUpdateButtonLatch(nExecutionPeriodUsec);	// Button latch time is independent of CPU clock frequency
 	PrintUpdate(uActualCyclesExecuted);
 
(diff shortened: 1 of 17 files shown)


tomcw:
Hi @sicklittlemonkey / Nick,

I've been wanting to do this for a while, ie. create a class for the Disk.cpp functions & variables. It's kind of step-1 for supporting .WOZ.

As a quick test, it was suddenly very easy to add a 2nd DiskII interface card to slot-5 :-)
(But there's no UI to support this yet!)
```
	if (g_Slot5 == CT_EMPTY)
	{
		DiskIIInterfaceCard* pDiskIISlot5 = new DiskIIInterfaceCard;
		pDiskIISlot5->Initialize(pCxRomPeripheral, 5);
		pDiskIISlot5->InsertDisk(0, "C:\\Work\\Personal\\Images\\lady tut PRODOS (san inc pack).dsk", false, false);
	}
```

The local class hierarchy becomes:
`DiskIIInterfaceCard` HAS_A `Drive_t` HAS_A `Disk_t`

Anyway, can you critique this pull-request?

Michaelangel007:
Moving at the end of the month so real busy packing but had time to review about half of this. Looks good so far! 👍 Excited to see .WOZ support coming along.

* [x] e7d6eac
* [x] bbe0e675
* [x] f755db85
* [x] 4235f08f
* [7c04c2c5]

Few quick comments.

1. I would like to see is a comment detailing the parameter `enhancedisk`.  Does an enhanced disk have more tracks? Different timing / encoding support? It isn't obvious from the name alone.

e.g.
```
virtual void Read(ImageInfo* pImageInfo, int nTrack, int nQuarterTrack, LPBYTE pTrackImageBuffer, int* pNibbles, bool enhancedisk)
```

2. Do we still need these constants `DRIVE_1` and `DRIVE_2` ?

in one place we do:

```
if (strlen(sg_DiskIICard.GetFullName(DRIVE_1)) > 0)
```

in another place we just pass it.

```
		DWORD duplicated = (DWORD)SendDlgItemMessage(hWnd, comboOther, CB_FINDSTRINGEXACT, -1, (LPARAM)sg_DiskIICard.GetFullName(driveSelected));
```

Making an enum is probably overkill but might want to consider being more consistent and just get rid of DRIVE_1 all together ?

Thanks for the cleanup of `iDrive` parameter.  We probably should review our Coding Style for arguments vs variables.

tomcw:
>(1) I would like to see is a comment detailing the parameter `enhancedisk`.

Sure, I'll add some comments fully documenting this.

>(2) Do we still need these constants `DRIVE_1` and `DRIVE_2` ?

I prefer some kind of constant, otherwise it could be easy to accidentally use the wrong magic number, eg. is `1` drive-1 or drive-2?. And they are used a lot: 31 instances of DRIVE_1 and 25 of DRIVE_2.

>Making an enum is probably overkill

So you prefer this...?
```
const int DRIVE_1 = 0;
const int DRIVE_2 = 1;
const int NUM_DRIVES = 2;
```

I'm fine to change it.

>(3) We probably should review our Coding Style for arguments vs variables.

Yep. These days I've ditched a lot of the Hungarian stuff.

Michaelangel007:
Re. request for enhancedisk.

> Sure, I'll add some comments fully documenting this.

Awesome, thanks!

Re. DRIVE_1, etc.

> I prefer some kind of constant

OK, no prob.

Ah, I see you already have a enum for DRIVE_1 in [Disk.h](https://github.com/AppleWin/AppleWin/blob/eea8cb3cdf21e8532c23c82cb65fa4459308a01e/source/Disk.h#L30).

Yeah, that's fine to keep the enum.  No sense in making busy work!  I should have looked at where it was defined first. =P  Just need to remember that DRIVE_1 is the numeric zero and not one.  Plus we can search for it as you mentioned.

> These days I've ditched a lot of the Hungarian stuff.

:-(  C'est la vie.

Are we still keeping `s` and `g` for static and globals, respectively?

Michaelangel007:
Reviewed:

* [x] aa1e4ca -- cleanup looks great!
* [x] 249b15b0
* [x] c6d45122
* [x] 971eff78 -- thanks for documenting enhancedisk!

Q, Is it worth renaming `DiskIIInterfaceCard` to `Disk2InteraceCard` ?  Those three I's make things a little hard to read.  It's not a big deal just thought I would mention it.

Q.  This is tangential to this patch -- I probably should log this as a Feature Request.  I was thinking about this yesterday -- is there any interest in allowing mounting a disk drive as "temporary"?  One of the problem is that AppleWin "locks" access to a disk when it is mounted and it is a bit of a PITA if you want to manipulate it with other utilities constantly mounting / ejecting.

Aside: Not sure why `Read/Write` and `Read` are grayed/greyed out for the 1st floppy in the right-click tool tip?

Michaelangel007:
I'll give Nick a chance to share his feedback but :+1: on Merge Request.

tomcw:
>Are we still keeping s and g for static and globals, respectively?

`g_` is useful, but as we turn more code into classes (like this Disk.cpp) then these will disappear (to an extent). btw. I couldn't see any use of `s`.

>Q, Is it worth renaming DiskIIInterfaceCard to Disk2InterfaceCard ?

Yes, I agree - I will make this change.

>Q. This is tangential to this patch -- I probably should log this as a Feature Request. I was thinking about this yesterday -- is there any interest in allowing mounting a disk drive as "temporary"?

Raised new issue here: #641 

>Aside: Not sure why Read/Write and Read are grayed/greyed out for the 1st floppy in the right-click tool tip?

I don't see this. Can you provide a repro?

>Do we want to add m_ prefix for fields in Drive_t?

Yes, I'll do this too.

sicklittlemonkey:
Looking good!

Depending on your plans, a presumably imminent step would be to try to push more logic down into FloppyDrive and FloppyDisk ... So may as well make them classes now rather than structs.

tomcw:
>a presumably imminent step would be to try to push more logic down into FloppyDrive and FloppyDisk

Yes, eventually / gradually.

>... So may as well make them classes now rather than structs.

Done!

Michaelangel007:
>> Aside: Not sure why Read/Write and Read are grayed/greyed out for the 1st floppy in the right-click tool tip?
> I don't see this. Can you provide a repro?

I'm not sure if this is a bug/feature but you can't select `Read/Write` or `Read only` until AFTER you mount a disk (as they are greyed/grayed out until that point.)

Is there a use case for selecting this BEFORE you mount a floppy?

tomcw:
I'd argue it's a property of just the disk (and not the drive), so I don't see a use-case for this.

Issue #544, opened 2018-02-24
Support WOZ disk image format


tomcw:
WOZ format v1.0 officially released on 17th Jan 2018:
http://evolutioninteractive.com/applesauce/woz_reference.pdf

>The WOZ format, the mainstay of the Applesauce project by John Morris, has been officially released. This new format for reading and writing Apple II floppy disks allows emulators to more accurately reflect the actual data on an Apple II floppy disk and disk image.

Ref: http://www.callapple.org/vintage-apple-computers/apple-ii/woz-format-updated-to-version-1-0-and-officially-released/

tomcw:
Here's a short program that demonstrates my understanding of the .woz file format:
- It consumes the first 2048 bits (ie. 256*8 bits) from a .woz file
- It ignores the extra sync zero-bits (or generates random bits)
- And dumps the resulting hex nibbles

[main.txt](https://github.com/AppleWin/AppleWin/files/1939958/main.txt)

If I get some time, I may have a stab at supporting (initially read-only) .woz files.
(I'm just stating this upfront in case @sicklittlemonkey or @Michaelangel007 want to take this instead.)

sicklittlemonkey:
We're still looking for a place to live, so I'm not biting off anything big
for the next month at least.

But you've been doing a lot recently Tom - great stuff!

Cheers,
Nick.

tomcw:
Just a though on .woz testing...

In 4am's notes on cracking "Gertrude's Puzzles" ([here](https://ia800507.us.archive.org/18/items/GertrudesPuzzles124amCrack/Gertrude%27s%20Puzzles%201.2%20%284am%20crack%29.txt)), it might be worthwhile to write a short bit of disk][ code to replicate this "E7 bitstream" (+ a dedicated .woz image).

sicklittlemonkey:
Yes, there will be a few interesting cases to test.

BTW, your program sums up the basic logic of the sequencer and transition
detection amplifier well. Shift bits, and generate spurious ones if there
are too many zeroes.

But the real-time specifics get a bit tricky eg:
- When the high bit is set then the sequencer loops while the latch holds
its value for seven cycles ... and longer if 0's come along (this is how
bit copy programs detect "sync bytes/nibbles")
- C08D resets the sequencer - used in the bit-slip (E7) protection
- Some of the sequencer code is redundant - namely that to detect zeroes,
since the bitstream is provided for us.

Anyway, the real question is how to shove bit-stream disk access into
[...]

PR #653, opened 2019-06-07
Support (read-only) WOZ1/WOZ2 images

tomcw:
PR for #544.

Supports:
- all "woz test images" v1.3 (WOZ1, WOZ2) are working, except 3.5"
- additionally: Frogger (spiradisc), Choplifter (not Enhanced //e!), Lode Runner, Marble Madness, Skyfox.
- woz images can be .gz or .zip compressed (ie. same as other supported images)
- save-state

Limitations:
- read-only, so WOZ images are forced to be write-protected
- 5.25" only (not 3.5")


tomcw:
Hi Michael, Nick (@Michaelangel007, @sicklittlemonkey):

OK, here's a PR up for review for WOZ support.

It's read-only WOZ support atm, but I'd prefer to release a read-only WOZ AppleWin version now, rather than delay waiting for a full r/w WOZ version, as I don't know how big a job adding write support will be.

The main changes are to Disk.cpp, where I've added a new ReadWriteWOZ() for $C0EC accesses. This complements the current ReadWrite() for .dsk/nib images, which still exists. And ControlStepper() has been extended to support quarter tracks.

If you want to build and test some WOZ images, then you can use [this zip](http://evolutioninteractive.com/applesauce/woz_images.zip) or grab an image from [woz-a-day](https://archive.org/details/wozaday)... and let me know what's not working!

sicklittlemonkey:
Hi Tom.

Wow, that looks great! I can't test it today, but I will as soon as I can.
I've had a look through the diffs, and it's good - surprisingly simple.

Looking forward to the WOZ-specific code being pushed out into another
class, and then it will be easier to play with.
YAE had a table-based step routine that included half-tracks, which might
make the code a bit tidier and perhaps faster.

Cheers,
Nick.

tomcw:
>YAE had a table-based step routine that included half-tracks, which might make the code a bit tidier and perhaps faster.

It's mainly your code in ControlStepper(), and I think it's pretty tight & efficient in the way you wrote it. 

One way to tidy it would be to consolidate `m_track`, `m_phase` & `m_quarter` into a single member, eg. `float m_phase`, which has an accuracy of 0.5 (ie. half a phase is a quarter track position).

Another thing I'm not convinced about (or don't understand) is your code for calculating `direction`. I think you rely on the fact that the DiskII firmware will seek to track-0 and turn off the magnets. Then to seek it'll turn on magnets from phase-0. But surely I could seek to track-0 stopping at another phase (1, 2 or 3), then that would be the track-0 base phase/magnet? Anyway it obviously works well as you have implemented it, so perhaps not worth worrying about.

sicklittlemonkey:
> YAE had a table-based step routine that included half-tracks, which might
make the code a bit tidier and perhaps faster.

Oops, I meant quarter tracks of course.

sicklittlemonkey:
> Another thing I'm not convinced about (or don't understand) is your code
for calculating direction. I think you rely on the fact that the DiskII
firmware will seek to track-0 and turn off the magnets. Then to seek it'll
turn on magnets from phase-0. But surely I could seek to track-0 stopping
at another phase (1, 2 or 3), then that would be the track-0 base
phase/magnet? Anyway it obviously works well as you have implemented it, so
perhaps not worth worrying about.

I'm sure you have already, but review Sather's //e book, pages 9-7:
"To rotate the rotor one step to the right in Figure
9.4, energize the phase-2 magnet, then deenergize
the phase-1 magnet. Magnetic attraction will then
align the cog with the phase-2 magnet."

That's how I calculate direction - just check for an adjacent magnet on.

And 9-14.
"The head assembly has a
mechanical stop (an electrical stop in some alter-
nate drives) at track 0. One method to absolutely
ascertain the head position is to step outward 80
phases as the bootstrap program does. The head will
run against the track stop, and you wiII be at track
0. The head/stepper motor linkage is aligned so that
the motor will be phase-0 aligned at track 0, so
from track 0 it is known that stepping inward must
be begun by turning phase-1 on by a reference to
$C083,X."

Track 0 is defined as the outermost track at phase 0.

Cheers,
Nick.

tomcw:
>The head/stepper motor linkage is aligned so that the motor will be phase-0 aligned at track 0, so from track 0 it is known that stepping inward must be begun by turning phase-1 on by a reference to $C083,X.

Ah, that's the bit I was missing!
(Of course, it's mechanical, so the cog won't turn anymore at track-0)

Thanks.

tomcw:
btw, some timings: NIB vs WOZ

This is for Rescue Raiders (Disk Access Speed = Enhanced Speed)
- Rescue_Raiders-(1_3).nib
  - Time from emulation reboot until first $C000 access: 515 msec
  - Time from emulation reboot until first $C000 access: 515 msec
  - Time from emulation reboot until first $C000 access: 499 msec
- Rescue Raiders - Disk 1, Side B.woz (WOZ v2, RRv1.5)
  - Time from emulation reboot until first $C000 access: 546 msec
  - Time from emulation reboot until first $C000 access: 546 msec
  - Time from emulation reboot until first $C000 access: 530 msec

Not quite the same image, since it's RR1.3(NIB) vs RR1.5(WOZ), but WOZ is only ~30ms (or 6%) slower than NIB.

sicklittlemonkey:
> Ah, that's the bit I was missing!
> (Of course, it's mechanical, so the cog won't turn anymore at track-0)

Indeed. The sound you hear when booting is the head assembly bouncing off
the stop:
"The head assembly has a mechanical stop (an electrical stop in some
alternate drives) at track 0"

tomcw:
re. my **simplified** logic state sequencer (LSS)

I have a `m_floppyLatch` and an internal `m_shiftReg`. This isn't accurate as the real interface card only has the `m_floppyLatch`, with extra state being held in the LSS itself (eg. the ability to hold the value of the latch for 7-cycle LDA $C0EC/BNE loop to read it).

I did quickly look at doing an implementation that more accurately matched the real LSS (see 21d3ff7), but reverted as my simplified version was, well... simpler!

sicklittlemonkey:
Simulating the sequencer is probably the toughest part. Does it run the E7
(slipstream) protected titles?

tomcw:
>Simulating the sequencer is probably the toughest part. 

Yeah - well I just mentioned this in case you look at the code and think "that's not how it's implemented in hardware".

>Does it run the E7 (slipstream) protected titles?

Yes (ie. LSS reset when reading mid-bitstream)... and other protection methods like: cross-track sync, half-tracks, spiradisc (quarter tracks), 3.5us bit-cell timing, fake bits etc.

See the readme.txt in John Morris' [test images zip](http://evolutioninteractive.com/applesauce/woz_images.zip), which has a great set of tests which cover different copy-protection methods.

sicklittlemonkey:
> Yes (ie. LSS reset when reading mid-bitstream)... and other protection
methods like: cross-track sync, half-tracks, spiradisc (quarter tracks),
3.5us bit-cell timing, fake bits etc.

Then this is awesome work. Ship it! Like you I'm surprised that more
sequencer "state" emulation isn't needed. (Yet!)

The reason I think YAE's table approach probably makes sense now that we
use quarter tracks is that we can replace this ...
if (m_phases & (1 << ((pDrive->m_phase + 1) & 3)))
    direction += 1;
if (m_phases & (1 << ((pDrive->m_phase + 3) & 3)))
    direction -= 1;

... and this ...
if ((m_phases == 0xC || // 1100
    m_phases == 0x6 || // 0110
    m_phases == 0x3 || // 0011
    m_phases == 0x9)) // 1001

... with a table like this and a simple lookup:
static int stepper_movement_table[16][NO_OF_PHASES] = {
{  0,  0,  0,  0,  0,  0,  0,  0 }, /* all electromagnets off */
{  0, -1, -2, -3,  0,  3,  2,  1 }, /* EM 1 on */
{  2,  1,  0, -1, -2, -3,  0,  3 }, /* EM 2 on */
{  1,  0, -1, -2, -3,  0,  3,  2 }, /* EMs 1 & 2 on */
{  0,  3,  2,  1,  0, -1, -2, -3 }, /* EM 3 on */
{  0, -1,  0,  1,  0, -1,  0,  1 }, /* EMs 1 & 3 on */
{  3,  2,  1,  0, -1, -2, -3,  0 }, /* EMs 2 & 3 on */
{  2,  1,  0, -1, -2, -3,  0,  3 }, /* EMs 1, 2 & 3 on */
{ -2, -3,  0,  3,  2,  1,  0, -1 }, /* EM 4 on */
{ -1, -2, -3,  0,  3,  2,  1,  0 }, /* EMs 1 & 4 on */
{  0,  1,  0, -1,  0,  1,  0, -1 }, /* EMs 2 & 4 */
{  0, -1, -2, -3,  0,  3,  2,  1 }, /* EMs 1, 2 & 4 on */
{ -3,  0,  3,  2,  1,  0, -1, -2 }, /* EMs 3 & 4 on */
{ -2, -3,  0,  3,  2,  1,  0, -1 }, /* EMs 1, 3 & 4 on */
{  0,  3,  2,  1,  0, -1, -2, -3 }, /* EMs 2, 3 & 4 on */
{  0,  0,  0,  0,  0,  0,  0,  0 } }; /* all electromagnets on */

So we get the direction and phase adjustment in one step.

I asked Doug Kwan to GPL it back in 2007, and he kindly obliged. (I hoped
to use it on the Sharp Zaurus at the time)
https://sourceforge.net/projects/yae-gpl/

Of course, we don't have to use his code. We can generate a table from
using the logic we're replacing.

PR #674, opened 2019-08-09
Prevent uninitialized value bugs and improve string safety.

beevik:
This change does two things:

**1. Updates the registry APIs to reduce the likelihood of uninitialized
variables.**

The code wasn't always checking the return value of registry load operations.
In some cases, this led to uninitialized memory being used, and crashes could
result. For example, LoadConfiguration in Applewin.cpp was using an
uninitialized value for the computer type if no registry variable for the
"Apple 2 type" was set.

New registry reading methods and macros have also been introduced, allowing
default value fallbacks for the cases where a registry variable is not found.
This makes registry access simpler and safer when a default value is known in
advance.

The registry code's style has also been updated to conform with the rest of
the code base (tabs instead of spaces, naming conventions, etc.)

**2. Introduces string safety improvements.**

A number of code paths have been modified to use safe-string functions instead
of their unsafe counterparts (e.g., strcpy, sprintf).  In the process, some
strings were converted from "char" to "TCHAR". This was done mostly for
consistency with the rest of the code-base.

### DIFF
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -60,9 +60,11 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #include "Configuration/PropertySheet.h"
 #include "Tfe/Tfe.h"
 
+#define VERSIONSTRING_SIZE 16
+
 static UINT16 g_AppleWinVersion[4] = {0};
-char VERSIONSTRING[16] = "xx.yy.zz.ww";
 static UINT16 g_OldAppleWinVersion[4] = {0};
+TCHAR VERSIONSTRING[VERSIONSTRING_SIZE] = "xx.yy.zz.ww";
 
 const TCHAR *g_pAppTitle = NULL;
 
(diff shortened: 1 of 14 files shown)


tomcw:
Some quick feedback:
- in general both of these two things are good improvements
- we should try to get this reviewed and committed soon before it becomes stale (ie. as the mainline moves forwards)
- I will be rejecting the switch from new/delete to malloc/free - so these changes need to be reverted
  - There's also 1 change from `VirtualAlloc()` to `malloc()`. Just use `new`.
- Minor point: (my) newer code tends not be use Hungarian notation _so much_ (there was a case where you added a "dw" prefix). Consistency with the original function should take precedence, so in this specific case (`halfScanLines` -> `dwHalfScanLines` in video.cpp) it's OK.

Thanks.

beevik:
Oops, that second commit (the one that included malloc/free) wasn't meant for this pull request.  I'll remove it.

tomcw:
For VS2008, I was getting lots of warning 4995's when compiling:
'function': name was marked as #pragma deprecated

...due to this addition to stdafx.h:
#include <strsafe.h>

So for VS2008 only (Debug & Release) I just included this in the Disable Specific Warnings.

tomcw:
See this comment: https://github.com/AppleWin/AppleWin/issues/559#issuecomment-480484253

Basically there are AppleWin users on very old versions of Windows, so I tend to stick with old versions of VS (from the History.txt, I switched to VS2008 in 2016).

I'd be good to move to a more current VS version, but of all the things to do, it doesn't feel particularly pressing.

Issue #678, opened 2019-08-23
Alternate key(s) to toggle max speed


Michaelangel007:
Currently a user can use `ScrollLock` to toggle full speed **after** they enable it in Options > Input > [x] `Scroll Lock acts as toggle for full-speed CPU.`

= Problems =

There are a few issues with this:

* `ScrollLock` isn't convenient to press (being on the far right of the keyboard)
* Macs don't have a dedicated `ScrollLock` key. :-/ (You _might_ be able to use Fn+Shift+F12 and/or F14 but people seem to report various issues of that working / not working.  Excel users on a MBP are SOL IIRC.)
* It is in the process of being phased out.  My Logitech Illuminated Keyboard requires `FN + Pause/Break` to send this ⤓ key.
* Virtual ][ uses `Command-3` for full speed. (MacOSX Command == Windows Ctrl.) (See http://www.virtualii.com/)

= Solution =

I propose using these Ctrl-# keys:

* `Ctrl-0` Speed fastest
* `Ctrl-1` Speed 100%
* `Ctrl-3` Toggle speed fastest / 100%

Taking a quick look at "Ye olde" Beagle Bros Peeks and Pokes scancodes chart ...

* http://www.lazilong.com/apple_II/bbros/ascii.jpg 

.... I see we can take advantage of ASCII's crappy design with its omission of no scancodes for Ctrl-0 .. Ctrl-9.

i.e. The Apple 2 can't detect Ctrl-0.  We don't have to worry about passing these keys into the virtual machine.

```
CALL-151
300:AD 00 C0 10 FB 8D 00 04 8D 10 C0 60
FC58G N 300G
```

I hacked in ~20 lines of code and added this feature this morning.

Pull Request coming up next.

tomcw:
@Michaelangel007:
>I propose using these Ctrl-# keys: (Ctrl-0,1,3)

What about Ctrl-2? Why did you skip this?

6502-workshop:
Thanks @Michaelangel007! This feature would definitely be a boost to my productivity. My primary workstation is a Mac and I RDP into a Windows machine when I need windows specific apps, thus I don't have a scroll lock key available.

Michaelangel007:
Hey Tom

Two reasons (pardon the pun):

* By putting a 1 key "spacer" between `Ctrl-1` and `Ctrl-3` it makes it harder for a user to accidentally press the wrong key
[...]

PR #679, opened 2019-08-23
Bug 678 toggle speed

Michaelangel007:
Sorry for the extra "junk" / files in the PR. Thankfully they are harmless (debugger) comments.

I created a new branch but didn't realize they were included for #678


tomcw:
Hi Michael,

I have a minor concern with Scroll-Lock behaviour and Ctrl-0/1 not taking affect:
- Config->Input: "Scroll Lock acts as toggle for full-speed CPU" = **checked**
- Tap Scroll Lock On = Full Speed
- In the main `ContinueExecution()` loop this will set local var `bScrollLock_FullSpeed=true`
- So regardless of what you do with Ctrl-0/1 it'll always run at full-speed

Similarly for:
- Config->Input: "Scroll Lock acts as toggle for full-speed CPU" = **unchecked**
- Press & holding Scroll Lock = Full Speed
- Ctrl-0/1 will have no affect
- NB. This is much less of an issue, since you'd have to holding Scroll Lock and also pressing Ctrl-0/1!

Anyway, as stated, this is a minor concern. I don't think it's worth extending the (already) fairly complicated Full-Speed logic in the main `ContinueExecution()` loop to improve this edge case.

tomcw:
Another thing: if I've set Config's CPU freq slide to eg. 2 MHz (btw. which also gets persisted to the Registry), then Ctrl-0 (speed toggle) will set to Full-Speed, then to 1 MHz.

I would expect that Ctrl-0 would toggle between Full-Speed and 2 MHz.

IMO, we should fix this one.

tomcw:
NB. Since it's saved to the Registry, then a quick fix could be to just load the value from the Registry. Not efficient, but this isn't speed critical, and it should keep the code simple.

tomcw:
@Michaelangel007 - I've updated the code to match my "spec" in #678.
From my point of view, it's good to merge!

PR #682, opened 2019-08-24
Use GENie to generate Visual Studio project files.

beevik:
This change adds a "Make_Solution.bat" batch file that uses the
GENie tool to create a Visual Studio solution and project files
for any desired version of Visual Studio.

See https://github.com/bkaradzic/GENie for a description of
GENie.

With this change, it will no longer be necessary to distribute or
maintain Visual Studio project and solution files for AppleWin.
You can simply build them with Make_Solution.bat instead. You
can build the project files for whichever supported version of
Visual Studio you like.  GENie currently supports Visual Studio
versions 2010 through 2019.

Using GENie also makes it easy to add new configurations for
AppleWin.  For instance, it is a one-line change to add a 64-bit
configuration.

### DIFF
--- a/Make_Solution.bat
+++ b/Make_Solution.bat
@@ -1,0 +1,21 @@
+@echo off
+
+set version=%1
+if "%1"=="" (
+    set version=vs2019
+)
+
+set geniepath=%~dp0genie
+pushd %geniepath%
+genie.exe %version%
+popd
+if ERRORLEVEL 1 goto FAIL
+
+echo Visual Studio solution created: %geniepath%\AppleWin.sln
+exit /b
+
+:FAIL
+echo.
+echo The GENie tool failed to run. Please download the genie.exe file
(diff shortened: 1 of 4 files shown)


beevik:
I should add that I designed the GENie configuration to generate a solution that is organized nearly identically to the VS2017 solution currently checked into the repo.

tomcw:
In the past I've considered cmake to generate the VS sln/proj files.
Do you know the pros/cons of GENie vs cmake?

Does GENie have a dependency on Lua being installed?

audetto:
cmake claims to support up to VS2019
https://cmake.org/cmake/help/v3.15/manual/cmake-generators.7.html

And it is a well established / supported tool.

Example of cmake files (for linux) are present in the linux port.
https://github.com/audetto/AppleWin

One could just add a few of the remaining files protected by some "ifdefs" based on the platform.

PR #683, opened 2019-08-25
Add support for a 64-bit version of AppleWin

beevik:
This change adds 64-bit support to AppleWin.
1. It adds a 64-bit Visual Studio configuration for AppleWin.
2. It corrects several lines of code that were incompatible with 64-bit compiles.
3. It corrects a larger number of warnings that were generated by the 64-bit compiler.

This change builds on top of an earlier pull request, https://github.com/AppleWin/AppleWin/pull/682, which added the ability to generate Visual Studio solutions and projects.

### DIFF
--- a/HookFilter/HookFilter.cpp
+++ b/HookFilter/HookFilter.cpp
@@ -6,13 +6,15 @@ static bool g_bHookAltGrControl = false;
 
 // NB. __stdcall (or WINAPI) and extern "C":
 // . symbol is decorated as _<symbol>@bytes
-// . so use the #pragma to create an undecorated alias for our symbol
+// . so use the #pragma to create an undecorated alias for our symbol (only in 32-bit mode)
 extern "C" __declspec(dllexport) LRESULT CALLBACK LowLevelKeyboardProc(
   _In_ int    nCode,
   _In_ WPARAM wParam,
   _In_ LPARAM lParam)
 {
+#ifndef _WIN64
 	#pragma comment(linker, "/EXPORT:" __FUNCTION__ "=" __FUNCDNAME__)
+#endif
 
 	if (nCode == HC_ACTION)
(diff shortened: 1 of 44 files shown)


tomcw:
Do the x64 builds work with the Visual Studio Community versions?

audetto:
Instead of adding casts for every other return value, would it not be better to change the variable type to the natural type returned by the function?

kiyolee:
By coincident, I have also done the x64 build (see branch target-x64 of my clone). The changes are most the same except I like to comment when changed from DWL_MSGRESULT to DWLP_MSGRESULT, the Windows API used should be changed from SetWindowLong to SetWindowLongPtr. I did build with VS2019 Community and ran Load Runner for an hour, if this is good enough as a test, the x64 build works. I didn't run faster though. :)
I was about to file a pull request, now I guess I can save that.

kiyolee:
> Instead of adding casts for every other return value, would it not be better to change the variable type to the natural type returned by the functions?

I think this is a common annoyance when building x64 Windows programs where size_t is 64 bit (used more often in C++ standard libraries) whereas int (long and DWORD as well) is 32-bit (which is used then more often in Windows API). You usually can't simply call Windows API with arguments taking results of functions like strlen() without warning. To shut up all warnings, you will need tons of type castings that I try to avoid as that may hide truncation problems. I know lot of people (including myself) don't like the kind of simply warning suppressing type castings and I usually don't bother to fill pull request for just that purpose.

tomcw:
@beevik - PR #747 also adds 64-bit build support.
It appears similar, but less changes than yours.
Can you take a look and comment? (As we now need to pick one of them to merge into master)

tomcw:
I am going to close this PR since I've now merged the similar PR #747 into master.

PR #687, opened 2019-09-07
Replace char * with std::string

audetto:


audetto:
This is roughly what the replacement would look like.
It is not meant to be merged right now, I still need to look at every single line and validate.

source/ParallelPrinter.cpp
```diff
@@ -228,32 +228,20 @@ static BYTE __stdcall PrintTransmit(WORD, WORD, BYTE, BYTE value, ULONG)
...
 {
-	if (*prtFilename)
+	if (!prtFilename.empty())
 	{
-		strcpy(g_szPrintFilename, (const char *) prtFilename);
+		g_szPrintFilename = prtFilename;
 	}
 	else  //No registry entry is available
 	{
-		_tcsncpy(g_szPrintFilename, g_sProgramDir, MAX_PATH);
```
audetto: This seems to be the only part of the code checking max path depth.
I have removed it. As long as the code can deal with file opening errors, I would let the OS deal with what is a valid path or not.

audetto:
It is in a better state now.
Please try it, have a look and comment.
I have not tested every single line, some of them I don't even know how to get to.

tomcw:
@audetto - I will discuss with Michael (@Michaelangel007) and Nick (@sicklittlemonkey).

source/Debugger/Debug.cpp
```diff
@@ -4779,14 +4768,16 @@ Update_t CmdMemorySave (int nArgs)
 		{
 			if (! bHaveFileName)
 			{
+				TCHAR temp[MAX_PATH];
 				if (! bBankSpecified)
-					sprintf( g_sMemoryLoadSaveFileName, "%04X.%04X.bin", nAddressStart, nAddressLen );
+					sprintf( temp, "%04X.%04X.bin", nAddressStart, nAddressLen );
```
Michaelangel007: Would like to see this `g_sMemoryLoadSaveFileName` -> `temp` change retain the descriptive name instead of the non-descript `temp`.
i.e. `sMemoryLoadSaveFileName`.

Michaelangel007:
Have reviewed commits:

* [x] [d5890e83](https://github.com/AppleWin/AppleWin/pull/687/commits/d5890e835632d01bbdd6c5f0c057c065a6c40260)
* [x] [f5f60310](https://github.com/AppleWin/AppleWin/pull/687/commits/f5f60310c1d5fcfe6cb82b80d44fd0dc5fdd0f9a)
* [x] [dced4793](https://github.com/AppleWin/AppleWin/pull/687/commits/dced4793b71f685f834586bc288b257236387eb2)

Aside from the non-descript `temp` change they look good.  Will review remaining commits later.

source/Debugger/Debug.cpp
audetto: It would be desirable to find an elegant pattern to accommodate std::string and a safe version of sprintf.

sicklittlemonkey: If using C++11 then I like and use the following solution:
https://stackoverflow.com/questions/2342162/stdstring-formatting-like-sprintf/26221725#26221725

If also using C++17 then the following gist also converts std::strings to const char* using constexpr where required, which is a nice safety feature:
https://gist.github.com/Zitrax/a2e0040d301bf4b8ef8101c0b1e3f1d5

audetto: Interesting, waiting for C++20's std::format, I am more than happy to try the first proposal after this.

tomcw:
Thanks @audetto - I've now merged this.

I've only done some basic testing (for both Debug & Release builds), but I expect some more edge cases will need fixing. We'll just deal with these as and when they are found.

tomcw:
Nick (@sicklittlemonkey) : perhaps we add your C++11 or C++17 code in a CPP conditional macro, eg:

```
#if _MSC_VER >= 1900
// VS2015 stuff
#elif _MSC_VER >= 1910
// VS2017 stuff
#endif
```

Just an idea to get it into the codebase now.

sicklittlemonkey:
Sure, though by the time it's active there might be better solutions in
C++23 or whatever. ; - )

PR #710, opened 2019-10-22
Debugger: BPMR & BPMW

tomcw:
Extended BPM to also support BPMR and BPMW
Fixed BPM[R|W] 0:FFFF (ie. support a range of 0x10000)


Michaelangel007:
Nice work Tom and Arnaud!

PR #715, opened 2019-11-02
Memory fix

audetto:
ImageInfo must be initialised by calling its c++ constructor (as it contains std::string).

In VS it seems to work (by chance?), in g++ it fails.
C++ standard requires the constructor anyway.

### DIFF
--- a/source/DiskImage.cpp
+++ b/source/DiskImage.cpp
@@ -57,7 +57,7 @@ ImageError_e ImageOpen(	const std::string & pszImageFilename,
 	if (*ppImageInfo == NULL)
 		return eIMAGE_ERROR_BAD_POINTER;
 
-	ZeroMemory(*ppImageInfo, sizeof(ImageInfo));
+	new (*ppImageInfo) ImageInfo();
 	ImageInfo* pImageInfo = *ppImageInfo;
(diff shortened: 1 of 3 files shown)


source/DiskImage.cpp
```diff
@@ -57,7 +57,7 @@ ImageError_e ImageOpen(	const std::string & pszImageFilename,
 	if (*ppImageInfo == NULL)
 		return eIMAGE_ERROR_BAD_POINTER;
 
-	ZeroMemory(*ppImageInfo, sizeof(ImageInfo));
+	new (*ppImageInfo) ImageInfo();
```
audetto: Is it really required a call to VirtualAlloc followed by a placement new? Can one not just use normal memory allocation with new?

audetto:
I think there is a more elegant way to handle this but it requires a newer c++.
```
struct AStruct
{
    int a_number = {};
    double another = {};
};
```

source/DiskImage.cpp
tomcw: Yes, I agree. I don't know why it's using VirtualAlloc() for this object!
Changing to "new" seems sensible.

audetto:
I've removed VirtualAlloc and simplified the code. Nothing else to push on this branch.

IMHO, it is ready to be merged.

PR #723, opened 2019-11-11
Correct way to clear a std::string.

audetto:

### DIFF
--- a/source/Harddisk.cpp
+++ b/source/Harddisk.cpp
@@ -189,9 +189,9 @@ static void HD_CleanupDrive(const int iDrive)
 
 	g_HardDisk[iDrive].hd_imageloaded = false;
 
-	g_HardDisk[iDrive].imagename[0] = 0;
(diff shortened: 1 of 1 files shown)


audetto:
This might look like an irrelevant fix, but without it, one cannot rely on the fact that 

HDD::imagename.empty()  (using .empty() from from std::string)

has the correct meaning.

Because it could be a buffer with a 0 at the begin, which is what this change avoids.

It is better to keep the string empty to signal an empty string, rather than mix some c and c++ emptiness for the same string. I had forgot it in the previous string changes.

kiyolee:
This is NOT irrelevent. The wrong code does not create an empty string, instead it create a std::string containing an initial NUL character and may be other junk following that.
Certainly wrong C++ code.

audetto:
I think more is true,

g_HardDisk[iDrive].imagename[0] = 0;

could well generate an access violation on an empty string.
Or an exception if debug stl is enabled.

.clear() is the only correct way of doing it.

Issue #726, opened 2019-11-19
More than two disk drives would be awesome


vromtramchick:
Hi!

I love playing Wasteland on your emulator.  

Back in the day, I had 4 FDDs hooked up to my Apple IIe and would play Wasteland.  It would be awesome if you added the option to have more than 2 FDDs.

Thanks.

tomcw:
It looks like Wasteland has 4 disks.
asimov link: https://mirrors.apple2.org.za/ftp.apple.asimov.net/images/games/rpg/wasteland/
But perhaps there are extra disk(s) for saving progress etc?

If I provided a command line way to put a 2nd DiskII controller in slot-5, and then (again from the command line) insert disks into slot-5 drive 1&2; would that work for you?

NB. You would not be able to remove or change the disks in slot-5 drive 1&2. From AppleWin's GUI you'd only be able to manipulate the disks in slot-6 drive 1&2.

Adding a configuration GUI to support drives in a slot-5 DiskII card would be more work.

vromtramchick:
I can't believe you responded so quickly.  I can't even believe you
responded at all!  I appreciate the prompt response, but I don't want you
to do any extra work.  At some point more than two drives would be great,
but please, I don't want you to put in extra effort.

Wasteland was the best game ever.  Being a fan of The Terminator (watched
it so much the tape in the VHS cassette snapped), Wasteland allowed you to
be exist in an apocalyptic world where the machines had taken over.  There
had been games like it, but none were as in depth or complicated.  Plus,
their anti-piracy technique was brilliant.  They used cutscenes.  Now,
cutscenes couldn't be visually presented, I don't think the concept of a
cutscene even existed at the time, but that's what they did - low tech.
They included a collection of paragraphs which you were instructed to read
as you progressed through the game.  Sure you could just read all the
paragraphs, but there were more fake paragraphs than paragraphs which
pertained to the game!  Plus if you didn't have the book of paragraphs, you
couldn't play the game.  You would miss out on too much information.

Wasteland's 4 disks were confusing at first because you didn't use them
like in a normal game.  First, you had make copies of all four.  You would
need the disk one master to boot the game (just as you are required to do
in your emulator), but you would play the game on the four copies you made,
swapping them out as your progressed through the different towns and
facilities.  When I was 14, I purchased my own apple IIe, and one summer, I
took the 2 drives off my dad's computer and attached them to mine, dropping
the second floppy drive card into, I believe, slot 7.  The game was
[...]

PR #741, opened 2019-12-17
Support 2nd Disk][ card and improved card management

tomcw:
PR for #726.

Support 2nd Disk][ in slot-5, via command line:
- -s5 diskii
- -s5d1 \<imagefile\>
- -s5d2 \<imagefile\>

NB. there's currently no Configuration UI support, except the Drive icons' tooltips show what's in slot-5 & slot-6 (for drive-n). So there's no way to eject the disks or insert new disks. The use-case I'm supporting it Wasteland which just has the 4 disks in the 4 drives.

Improved card management:
- Added `class Card` (in Card.h) which all other cards (that exist as classes) derive from (eg. LC,SSC,Mouse,Disk2).
- Added `class CardManager` (in CardManager.cpp\h) which now manages the 8 slots (and aux slot).
- Added `class Disk2CardManager` (in Disk2CardManager.cpp\h) which provides methods for operations that act on all Disk2 instances at the same time.
- Currently limited to just 1x SSC and 1x Mouse card (why would you need more?). This simplifies things, meaning there's no need to have dedicated SSCManager / MouseCardManager objects.
- Currently the 2nd Disk2 card can only be put into slot-5. This limitation is just due to the complexity of the Configuration UI. Having a more general drop-down per slot UI would remove this limitation.

I'm putting this up for review, but there are a lot of changes, so if you are curious then maybe just look at the new classes I've added.


tomcw:
>Does this eliminate the need for Tommy's multiple drives branch? I forget how many drives he added!

Not entirely. (He also added a 2nd disk controller to slot-5, so 2 extra drives.)

But he had extra GUI for:
- 10 buttons (was 8) on the right hand side; where 2 drive buttons became 4 half-height drive buttons. This meant you could insert/eject images for all 4 drives.
- Configuration UI: 2 extra combo drop-downs for the 2 extra drives.

PR #747, opened 2020-01-07
X64 enabling

kiyolee:
Here are a few changes to make 64-bit build possible.
The changes are mainly to use windows api in the correct way that works for both 32-bit and 64-bit builds.

### DIFF
--- a/HookFilter/HookFilter.cpp
+++ b/HookFilter/HookFilter.cpp
@@ -12,7 +12,9 @@ extern "C" __declspec(dllexport) LRESULT CALLBACK LowLevelKeyboardProc(
   _In_ WPARAM wParam,
   _In_ LPARAM lParam)
 {
+#ifndef _WIN64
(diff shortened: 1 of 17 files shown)


kiyolee:
I submit this pull-request because I think the source code changes required for x64 should be separated from the projects/solutions changes. That makes thing clearer. See there is no conflicts to the base branch.
The changes here not specific to x64 either. It uses more correct windows apis and makes code more generic across platforms.

kiyolee:
Btw, if you are interested to see how I build my x64 binaries, you are welcome to look at my `target-x64` branch. I have been maintaining a different set of VS projects/solutions for some time now.
The advantage of my set of VS projects/solutions is that there is a cleaner structure to support all (most) various versions of VS. They all can build inplace without conflicting with each other. The base branch only supports using one version of VS at a time.
I have just cleaned up the git history (a bunch of git rebase) a bit. If you like that, it should be just one pull-requiest away.

tomcw:
Thanks @kiyolee - finally merged this into master.

Issue #709, opened 2019-10-22
AppleWin Settings


Harbinger-365:
Hi all,

I am working on a project to allow Apple II games to be played in a portable fashion setup so that the end user doesnt need to know how to use the emulator itself, just click a bat and it will play the game. The idea is to make these games more accessible for people who are curious but dont know much about computers in general. To that effect I have an install script which sets everything up for the user including default settings for AppleWin based on a few questions asked during the script.

My issue is that as far as I know these settings need to be inserted into the registry for AppleWin to use. I have done this using a reg file, however this triggers Windows UAC which might scare some users away. I was wondering if you had ever considered using an .ini file to store the settings instead of the registry, therefore making it easier to change a setting for more casual users, making it more portable with pre configured settings, and removing the need to edit the registry?

I apologise if this sort of suggestion bothers anyone on the project, I truly do appreciate the work you have all done. And thanks again for fixing the HDV dismount issue.

tomcw:
What configuration are you needing to set in the Registry?
EG. perhaps there are already command line options for these settings?

NB. similar request in #446

Harbinger-365:
Similar idea to #446 for a different reason, yes. In my case the user very likely will have never run AppleWin without one of my game launch bat files so there will be nothing set in the registry at all. One result would be no sound enabled as you need to enable sound within the settings of AppleWin (to my understanding) on an initial setup. I also default to Full-screen show subunit status as off, and Mockingboard on, for the users registry. There is also some questions asked to the user via the setup.bat whether they would like RGB, NTSC, Color TV, Scanlines, and whether to use joystick or keyboard for controllers. All of this feeds the reg file which I then insert into the registry at the end of the bat. It would be much simpler via an .ini which is why I ask.

tomcw:
OK, this all makes sense - and I agree that things would get very complicated on the command line!

Switching from the Registry to an ini file (or better a yaml file) seems like a big job. But I've not given it any thought at all, so perhaps I'm wrong. So no promises on if/when this can be achieved.

Harbinger-365:
I appreciate your looking into it.

Keatah:
I would also love to see config.ini support.

Michaelangel007:
Tom, 

As a short-term compromise would it be useful to extend the debugger to have a command to load/save ini settings?

[...]

Issue #446, opened 2017-07-26
Make Applewin 'portable'.


Keatah:
Is there a way to make AppleWin behave like a 'portable' app and have it save its config to an ini file instead of the registry? And is there a way to disable the "I accept" & "I decline" dialog - which always pops up when running a different version..?

When doing curating and archiving activities I often switch back and forth between 2 versions. 1 for better video modes and speed, the other for better compatibility.

TommyGH:
"And is there a way to disable the "I accept" & "I decline" dialog - which always pops up when running a different version..?"

Interesting, I just made a comment about this "I accept" and "I decline" in another issue.  Until now I was unaware that it was due to my switching back and forth between versions!  And I second the motion to disable the pop up (if possible).

tomcw:
>And is there a way to disable the "I accept" & "I decline" dialog - which always pops up when running a different version..?

This is the prompt showing the GPL license. I guess we could change from:
- show prompt if version changes (since the last accepted version)

To
- show prompt if version is newer than the last accepted version

NB. If you want to suppress this, then you'd need to set the correct `Version` in the Registry before starting that version of AppleWin:
HKEY_CURRENT_USER\Software\AppleWin\CurrentVersion\Configuration\Version

tomcw:
>Is there a way to make AppleWin behave like a 'portable' app and have it save its config to an ini file instead of the registry?

Not currently.

I guess you could write a script to export the sub-branch:
HKEY_CURRENT_USER\Software\AppleWin\CurrentVersion\Configuration
Then import the other version's sub-branch.

This would also "solve" the above issue with the prompt about accepting the GPL each time.

tomcw:
But really you want #357 done, which is the underlying problem.

Keatah:
Yes I primarily switch because of video modes. The second reason is speed. the older the version the faster it runs. more or less I suppose.

To eliminate the dialog I think I'll make a tiny batch file that plugs the version # I'm using into the registry.

tomcw:
>The second reason is speed. the older the version the faster it runs. more or less I suppose.

You must be talking about the full-speed mode. But specifically is this during **regular** emulation or debugger stepping emulation? If it's the latter, then there is a `gg` command to "Go (unthrottled)" in single-stepping mode (eg. with breakpoints active). This was added at experimental v1.26.2.0.

Keatah:
I would be referring to standard/regular emulation. Nothing in debugger.

[...]

PR #757, opened 2020-02-10
Allow use of an INI-file instead of the Registry

taeber:
This change adds the `-conf` switch which, if specified, will use an INI-file instead of the Windows Registry.

This affects AppleWin/AppleWin#709 and AppleWin/AppleWin#446.

## Testing
I tested this using VS2019 on Windows 10.

1. Open PowerShell and navigate to the folder containing `AppleWin.exe`.
2. Run `AppleWin.exe -conf config.ini`, where `config.ini` is a non-existing file.
3. Close AppleWin.
4. Note the creation of `config.ini`. Also note that no values in the Registry should be changed (this is easy to test if you remove the AppleWin group first so you can see that nothing is created).
5. Run `AppleWin.exe` normally (without `-conf`). Change some settings. Notice that the Registry is used and the `config.ini` is unaffected.

### DIFF
--- a/help/CommandLine.html
+++ b/help/CommandLine.html
@@ -11,6 +11,8 @@
 		<p style="FONT-WEIGHT: bold">AppleWin can be driven from the command line as 
 			follows:
 		</p>
+		-conf &lt;pathname&gt;<br>
+		Use an INI file for configuration instead of the Registry.<br>
 		-d1 &lt;pathname&gt;<br>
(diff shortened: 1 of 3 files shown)


tomcw:
@taeber - this PR looks simple and neat, so I don't see any issues in accepting it(*). Before I merge it in, I want to give it some testing. I'll try to find some time this week. Thanks!

(*) MSDN says that GetPrivateProfileString() and WritePrivateProfileString() are intended for Win16 compatiblity... so I wonder for how much longer Microsoft will keep them for?

Issue #734, opened 2019-12-02
DiskII controller card: support 13-sector firmware


tomcw:
There have been a number of recent issues/comments about 13-sector .woz images.

DOS 3.3 System Master's START13/BOOT13 can be used but it's not a great workflow (and not obvious to all users). Also at least one title (Micro Chess 2.0) depends on the real firmware. (See #732)

So as a convenience, have the card auto-select the firmware based on the .woz image properties.

Only change the firmware at reset/reboot to avoid changing whilst running in $C6xx space!

NB. No plans to support the rare .d13 format. And no plans to auto-detect whether a .nib is 13-sector. For dual 13/16-sector images (eg. Taipan.woz) then prefer 16-sector firmware.

PR #761, opened 2020-02-21
Auto-switch DiskII firmware to 13 or 16 sector depending on disk in drive-1

tomcw:
PR for #734.

Auto-switch is done on each reset.
Only .WOZ v2 (or higher) images supported, as these indicate whether the boot sector is 13 or 16 sector.


tomcw:
I toyed with the idea of having the drive LEDs different colours for a DiskII card w/13-sector firmware (eg. blue for read, light-blue for read from write-protected image), but instead took the simpler route of just adding 'S6-13' to the window's title.

PR #775, opened 2020-03-29
Fixed CmdCursorSetPC command

ThorstenBr:
Hitting "=" in the debugger is supposed to set PC to the _current cursor address_.
Setting PC to the _number_ of arguments made no sense anyway.

### DIFF
--- a/source/Debugger/Debug.cpp
+++ b/source/Debugger/Debug.cpp
@@ -3661,7 +3661,7 @@ Update_t CmdCursorPageUp4K (int nArgs)
 //===========================================================================
 Update_t CmdCursorSetPC( int nArgs) // TODO rename
 {
-	regs.pc = nArgs; // HACK:
+	regs.pc = g_nDisasmCurAddress; // set PC to current cursor address
 	return UPDATE_DISASM;
 }
 


tomcw:
In `DebuggerProcessKey(int)`, CTRL+RIGHT_ARROW also calls `CmdCursorSetPC(g_nDisasmCurAddress)`.

I'll change this to pass 0 instead, and set the PC directly as you do.

tomcw:
Thanks for this PR. I manually checked-out your branch, tweaked it a bit, then squashed it (to reduce to a single commit), but in so doing, I lost you as a contributor for this PR. Sorry about that!

Anyway, I'm closing this, since it's merged to master now.

ThorstenBr:
Ok, no problem, thanks.
I did a complete Linux port of the debugger for Linapple and am currently testing on Linux. Meanwhile, I noticed a few more issues with some debugger commands and documentation, also affecting AppleWin. When I'm done, I'll provide another pull request with a set of patches for "upstream" AppleWin.

tomcw:
Sounds very interesting!

The debugger's code is a classic example of lots of code with many corner cases, but no unit tests! So as things got extended, the corner cases were left untested, so things broke.

In the VS2008 vcproj, I did start on a few micro-unit tests for the debugger. eg here:
https://github.com/AppleWin/AppleWin/blob/master/test/TestDebugger/TestDebugger.cpp
...but this is just a drop in the ocean.

I think a better way to do it, would be to add a gdb/lldb server to the debugger (exposing all the functionality), and then properly exercise it eg. using a python client. But even this only tests the internals, and not the debugger's CUI.

Issue #773, opened 2020-03-28
Add Apple II J-Plus support


tomcw:
Now that custom ROMs can be loaded (#771), then I thought that J-Plus support (from the command line) was done, eg:

```
AppleWin.exe -model apple2p -rom IIJPLUS.ROM -videorom JVIDEO.ROM
```

But @sicklittlemonkey told me that the video rom (in AppleWin 1.29.11.0) isn't quite correct...

>I did give it a go, though I wonder where that ROM
collection came from. I wrote the text files, but those ones are
different from the ones I remember releasing, unless there were
multiple versions. The full versions are here:
https://mirrors.apple2.org.za/Apple%20II%20Documentation%20Project/Computers/Apple%20II/Apple%20II%20j-plus/Documentation/
>
>Searching around I did find a problem with what I've written, and the
j-plus video ROM doesn't behave exactly as I would expect in AppleWin:
https://mametesters.org/view.php?id=6175
>
>I don't have my original notes handy, but from memory I was going by
screenshots like the one at the link above (in fact possibly that
exact one) ...
>
>It would be nice to get this working as it should.

Here's what AppleWin 1.29.11.0 shows...
![IIj+_videorom](https://user-images.githubusercontent.com/6696896/77821653-c6000a80-70e3-11ea-995b-b0c6f43e47a3.png)

From Nick's notes:
```
Characters in ROM:
00-3F Katakana } 00-7F activated by access to $C05C (annunciator 2 off)
40-7F ASCII    }
80-BF ASCII    ] 80-FF activated by access to $C05D (annunciator 2 on, set after reset)
C0-FF ASCII    ]

Characters displayed onscreen with annunciator 2 off:
00-3F Katakana - normal
40-7F Katakana - flashing
80-BF ASCII    - normal
C0-FF ASCII    - normal

Characters displayed onscreen with annunciator 2 on:
00-3F ASCII    - inverse
40-7F ASCII    - flashing
80-BF ASCII    - normal
C0-FF ASCII    - normal
```

So AN2 controls what video rom chars are displayed for 00-7F.
**EDIT:** NB. there's no way to get INVERSE Katakana.

To do:
- [x] Add a new Apple II J-Plus model (apple2jp) to the code base
- [x] Support AN2 just for this apple2jp model
- [x] Expose this via cmd line & Config UI

sicklittlemonkey:
Just a note that we verified the flashing behaviour. Here's a photo where it's caught changing.
[...]

PR #776, opened 2020-04-01
Support Apple II J-Plus

tomcw:
PR for #773.


tomcw:
> Looks good. That commit message says AN3 rather than AN2, but those messages are a pain to change! ; - )

Actually that was just a general bug for all the Apple II/II+ models (nothing to do with the J-Plus support) - AN3 was being "stolen" by DHGR, and not properly set for these pre-//e models.

PR #785, opened 2020-05-19
Bookmark cleanup

Michaelangel007:
There were numerous minor issues with bookmarks:

* Having multiple bookmarks (`Ctrl-Shift-#`) made it impossible to tell exactly which one you would jump to via the `Ctrl-#`.  They now have their own indicator as a QoL improvement.  (Debug_Font.bmp was updated to included "enclosed numerics".)
* Replacing an existing bookmark would incorrectly increase the total.  You would quickly hit the maximum making the entire bookmark system essentially worthless.
* Even worse (!) resetting bookmarks `bmc *` didn't actually clear the total back to zero.  DOH!

Old bookmarks:

* bookmarks_old

New bookmarks:

* bookmarks_new

With the new glyphs we could also visually show the enclosed numerics in red for breakpoints.

NOTE: Technically we support 10 bookmarks however `Ctrl-Shift-0` is NOT being seen by AppleWin due to Windows _Input Language_ shenanigans. I tried various settings but couldn't get Microsoft's broken keyboard settings to actually work.

Reference(s):

* https://en.wikipedia.org/wiki/Enclosed_Alphanumerics
* https://superuser.com/questions/1367843/how-to-disable-ctrlshift-keyboard-layout-switch-for-the-same-input-language-i


tomcw:
Hi @Michaelangel007 - basically LGTM! So merge it in.

btw, I wasn't even aware of bookmarks before this... please add to the debugger HTML docs (https://github.com/AppleWin/AppleWin/tree/master/help) if you can, or at least point me at where to add some info (and provide a "spec" for them). Then I can add some quick info to the docs.

re. Ctrl-Shift-0
I don't know if it'll help, but there is now a keyboard hook filter, so you can suppress this key-combos being pass on to Windows.
See this project: https://github.com/AppleWin/AppleWin/blob/master/HookFilter/HookFilter.cpp

Perhaps this can be used to hook Ctrl-Shift-0? Should be easy to test.

Tom

Michaelangel007:
Quick Spec:

---

One can navigate around the disassembly window using the arrow keys.  For example pressing right arrow `->` will take you to the target address of a `JSR`.  (*)  You can press the `TAB` to return to where the PC (Program Counter) is but what if you want to inspect several different functions?

Enter bookmarks.  They allow you to "tag" an address and consequently jump back to them.

* Set a bookmark use `Ctrl-Shift-#`, that is:
  *  `Ctrl-Shift-1` set bookmark 1
  * `Ctrl-Shift-2` set bookmark 2
  * `Ctrl-Shift-3` set bookmark 3
  * `Ctrl-Shift-4` set bookmark 4
  * `Ctrl-Shift-5` set bookmark 5
  * `Ctrl-Shift-6` set bookmark 6
  * `Ctrl-Shift-7` set bookmark 7
  * `Ctrl-Shift-8` set bookmark 8
  * `Ctrl-Shift-9` set bookmark 9

* To jump to a existing bookmark, if it exists, press `Ctrl-#`, that is:
  * `Ctrl-1` to jump to bookmark 1
  * `Ctrl-2` to jump to bookmark 2
  * `Ctrl-3` to jump to bookmark 3
  * `Ctrl-4` to jump to bookmark 4
  * `Ctrl-5` to jump to bookmark 5
  * `Ctrl-6` to jump to bookmark 6
  * `Ctrl-7` to jump to bookmark 7
  * `Ctrl-8` to jump to bookmark 8
  * `Ctrl-9` to jump to bookmark 9

* To move to the previous bookmark press `Ctrl -`
* To move to the next bookmark press `Ctrl +`

Prior to AppleWin 1.30 bookmarks would show up as a cyan background color on the address in the left column.  In AppleWin 1.30 they now show up as a number with a circle around them where the `:` separator is between address and opcodes are listed.

For example in this screenshot:

* bookmarks_new

We see that 9 bookmarks have been set.

```
    4233①ED 46
    4237②6C 48
    423B③02 50
    423D④81 53
    4243⑤92 55
    4247⑥F6 55
    424B⑦2F 57
    424F⑧11 58
    4253⑨D1 58
```

That is, address $4233 has bookmark 1 set to it, address $4237 has bookmark 2 set, etc.

Bookmarks can appear in any order that you set -- not just contiguous.

If you scroll away you can quickly revisit these locations by pressing `Ctrl-1`, etc.

An address can only have ONE bookmark assigned to it.  If you try setting a new bookmark over an existing one, the old one will become free for use.

Here is a table of native bookmarks commands from inside the debugger

|Command | What it does|
|:--|:--|
| `help bookmarks` | Lists all bookmark related commands |
| `help bc` | Lists specific help about the `bc` command |
| `bma` | Add a bookmark |
| `bmc #` | Clear a specific bookmarks, for example `bmc 1` clears bookmark 1 |
| `bmc *` | Clears all bookmarks |
| `bml` | Lists all boommarks |
| `bmsave` | Saves all bookmarks  -- currently broken |

---

(*) NOTE: This was broken on commit [c35b863b](https://github.com/AppleWin/AppleWin/commit/c35b863bd9286ee2658a1a944d02b02a13b1ad6c).  I have a patch that will be included in #786

```
```

Michaelangel007:
@tomcw Thanks for the feedback and the info about `HookFilter.cpp` !  Merged.

I'm going to punt on fixing `Ctrl-Shift-0` and `Ctrl-0` for now but I'll make a mental note of it to investigate later.

I have a ton of mouse button functionality currently in the works that really _streamlines_ navigating around.  I'll write-up all the new stuff on #786 once it is ready to be merged.

tomcw:
Hmm, I quickly tried adding a hook filter for `Ctrl-Shift-0`, but it didn't work.
```
		// Suppress ctrl-shift-0
		if (pKbdLlHookStruct->vkCode == '0')
		{
			if (GetKeyState(VK_CONTROL) < 0 && GetKeyState(VK_SHIFT) < 0)
				suppress = true;
		}
```

NB. If I set a breakpoint on the `suppress` line, then it does get hit when I type `Ctrl-Shift-0`. But back in `DebuggerProcessKey( int keycode )` the code to handle 0...9 does NOT get hit.

tomcw:
Added new CHM help section on bookmarks (a332e58f1)

Issue #795, opened 2020-06-03
Memory management


fenarinarsa:
Hello,

I'm currently working on extending the debugger's features and I need to be able to access the real contents of the RAM banks (main & aux) at any given time.

I checked Memory.h and Memory.cpp and I found out that there is many layers of memory buffers:
mem
memmain, memaux
memwrite
memdirty
memshadow
It seems the various buffers are refreshed by UpdatePaging() when the CPU flips memory-management soft-switches.

Currently in order to "see" the actual contents of both banks, I need to rebuild them on the fly by using mem+memdirty (which acts like a cache), memmain and memaux.

I'm wondering what was the reason behind this design choice. Maybe it's performance-related?

Because the way I see it, it can be rewritten to be simplified a lot. For instance, a memshadow_RD and a memshadow_WR that contains pointers to memmain and memaux's pages and accessed that way  (excluding IOs of course):
`value = *(memshadow_RD[addr>>8] + (addr & 0xFF))`
would work without a lot of performance loss on today's computers.

Then both memshadow could be updated in UpdatePaging(). Maybe even by copying a bunch of pointers for large memory zones like $60xx-$BFxx. It would be a maximum of 512*2 bytes to copy instead of a lot of cached data as it's done now.

I'm asking because I know AW handles a lot of things so maybe I'm missing something important there, like a feature that won't work this way?

tomcw:
Yes, it is a little complex! The `UpdatePaging()` design was done by the original AppleWin author, and it's (more or less) remained like that since.

At a high level, the 64KiB `mem` is a consistent *virtual memory* view from a *read* perspective. It gets updated whenever the virtual mem read-view changes. Writes are done indirectly via pointers.

So it's efficient if there's little bank switching (either language card or main/aux)... but if there's lots of bank switching then your full indirect approach for both read & write is likely more efficient.

I'm reluctant to change it because: (a) it works (if it ain't broke don't fix it), and on modern PCs it's fast; (b) Apple II bank switching is complex anyway, so changing it will introduce bugs; (c) I've got plenty of other things to work on ;-)

If it's *physical memory* you want to r/w, then take a look at `CmdMemorySave()` in debug.cpp. This uses `MemGetBankPtr()` to flush the cache and get a pointer to the physical 64KiB bank.

fenarinarsa:
Thank for the answer, so it's indeed a legacy reason.

[...]

Issue #788, opened 2020-05-23
Development: Add support for logging the time for emulation sub-systems


tomcw:
In cea2 ([here](url)), Andy McFadden asks about time spent in emulation sub-systems.

>Do you have stats on how much of emulation performance is instruction fetch/decode/execute vs. other emulation tasks? 

This is interesting from the POV of AppleWin, so worthwhile doing some profiling to answer this question.

tomcw:
My comment on cea2:

Here are some quick numbers for the latest AppleWin 1.29.12 (release build, w/out Visual Studio attached, RGB video) running French Touch's MadEffect2 demo:

```
Perf breakdown:
. CPU %     = 18.217025.2
. Video %   = 68.534746.2
... NTSC %    = 61.095756.2
... refresh % = 7.438990.2
. Audio %   = 11.952201.2
... Speaker % = 2.079797.2
... MB %      = 9.872404.2
. Other %   = 1.296029.2
. TOTAL %   = 100.000000.2
```

I just put markers in the main emu loop, where it does useful work; but didn't include the time spent in the big "sleep" which is used to peg emulation speed to 1MHz (or whatever MHz speed you selected).

NB. CPU% can vary a lot from title-to-title depending on things like heavy bank-switching (which is just memcpy) but can quickly impact performance if the 6502 code is written *NOT* efficiently for emulators! :-)

Heavy bank-switching (memcpy) will surely dominate even recompiled code, meaning no benefit for compiled over interpreted 6502, for these cases.

AppleWin is single-threaded and in release-build it only consumes ~2% of an AMD64 CPU core (3GHz). So I could offload some tasks to other compute units (eg. offload video to the GPU, and probably audio to another CPU or sound card), but there's no need for this added complexity when it runs fast enough already.

Additionally: since video dominates CPU (here by a factor of ~4) then the focus should be in speeding up the video part, not the CPU part... assuming you want opcode-accurate video, eg. for beam-racing demos.

btw. in AppleWin's "full-speed" mode, then video isn't updated per opcode, instead just once a video frame. Also audio is muted. So CPU% tends to ~95%, video is just refresh (~5%).

tomcw:
### Timings for Willy Byte (game, which has Mockingboard music during the intro)

--- DEBUG, RGB: (Intro until WB turns switch on) / using VS to capture OutputDebugString()
```
Perf breakdown:
. CPU %     = 16.555035.2
. Video %   = 73.397352.2
[...]

Issue #572, opened 2018-07-24
4x window mode for HiDPI displays


hexnut:
Running on a Dell XPS 13 9630 with 3200x1800 resolution display via wine on Fedora Linux. 2x windowed mode is a bit small and fonts would be unreadable except for font DPI scaling with winecfg. Thinking a 4x mode would be about right.

tomcw:
From issues #358, #488:
AppleWin will try to integer scale the base resolution of 560x384.
(Currently it only attempts to scale by 2.)

So this means:
- x4: 2240 x 1536

As AppleWin scales using GDI's StretchBlt(), then there's no smoothing (anti-aliasing) so the resulting image will just get chunkier.

Anyway, it's probably worth supporting this for now.

sicklittlemonkey:
How does fullscreen look at 3200x1800?

Cheers,
Nick

rzumer:
Making it configurable would also be nice.

Issue #538, opened 2018-02-10
Linux, AppleWin & Qt


audetto:
Hi,

I would like to share a little project on which I have worked in the last months.
It is a port of AppleWin to a native linux application based on Qt.

It is available here https://github.com/audetto/AppleWin

It is actually 2 ports, one to ncurses and one to Qt as described in https://github.com/audetto/AppleWin/blob/master/linux.md.

I have written 2 frontends to better understand what is emulator code and what is frontend.

I have reached a point where "normal" emulator features are in a usable state and don't feel embarrassed any longer to share it.

The main goal of this exercise is the following: keep merging from AppleWin trivial.
I do not want to spend ages to merge and for this reason very few changes to the AppleWin source files have been allowed. Mostly to fix unavoidable compilation errors in gcc or when the cost of making the code compile was too high (see the #ifdef at the end of MouseInterface.cpp).

What about graphics? It is very similar to linapple (maybe like AppleWin 1.25), lo res in color, hi res in BW. No options, and I don't think I am making good usage of OpenGL.

What's next? Good question!

I don't really know myself. I would definitely like to commit to AppleWin all the "pure c++" changes to make merging even easier, and I will create a branch for this if anyone is interested.

Why Qt? Why not SDL? To be honest I have chosen Qt because it provides so many UI widgets and features that I had no idea how to create in SDL. But at the same time I have realised that timers in Qt are nowhere as precise as in Windows.

More features? Audio? libretro? debugger (there is too much Win32 code in the debugger files to be used as they are)? Android? Depends on time.

Anyway, guess I will stop when the fun stops.

audetto:
Here is the branch that includes the minimum set of gcc required changes.

https://github.com/audetto/AppleWin/tree/gcc

It is just for information for the time being. It needs a little work to ensure it works in VS as well.

AppleJackOverlord:
Thank you!

sicklittlemonkey:
> It is a port of AppleWin to a native linux application based on Qt.

Great.

> I have written 2 frontends to better understand what is emulator code and
what is frontend.

It would be awesome to have a truly cross-platform AppleWin and abstract
away the OS dependent parts.

> The main goal of this exercise is the following: keep merging from
AppleWin trivial.

Fair enough.

> I don't really know myself. I would definitely like to commit to AppleWin
all the "pure c++" changes to make merging even easier, and I will create a
branch for this if anyone is interested.

Sounds good! I hope Tom agrees.

[...]

PR #799, opened 2020-06-10
Extended debugger

fenarinarsa:
Improved debugger with realtime follow of the running emulation.

- Resized window when entering debugger
- Display emulator output and memory heatmap (activity) next to the debugger console
- Emulator output follows the video counter (CRT beam) after tracing an opcode
- Improved graphical perfs in debugger so it can follow the emulator live in "go" (stepping) mode.
- F7 still used to enter/break/exit the debugger

New debugger commands:
G - Run, debugger following PC
GD - Run, debugger NOT following PC
GC - Run "cycle": Emulator live output refresh after each opcode.
xxxxD - Shortcut to D xxxx (hex dump)

Heatmap colors
Blue: Read
Red: Write
Green: Exec
White: Current displayed video page

I/O and ROM pages are not shown, only the two 64k active banks of RAM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ AppleWin
 Apple II emulator for Windows
 
 
-Download latest (stable) release: [AppleWin v1.29.13.0](https://github.com/AppleWin/AppleWin/releases/download/v1.29.13.0/AppleWin1.29.13.0.zip)
+Download latest (stable) release: [AppleWin v1.29.15.100FT](https://github.com/Fr3nchT0uch/AppleWin/releases/download/v1.29.15v100FT/AppleWin1.29.15.100FT.zip)
 
-Release Notes: [v1.29.13.0](https://github.com/AppleWin/AppleWin/releases/tag/v1.29.13.0)
+Release Notes: [v1.29.15.100FT](https://github.com/Fr3nchT0uch/AppleWin/releases/tag/v1.29.15v100FT)
 
 
 Building
@@ -22,12 +22,3 @@ Experimental build: pending
 
 Please report [new issues](https://github.com/AppleWin/AppleWin/issues/new)
 
-
-Previous Versions
-=================
-
-* [AppleWin v1.28.8.0](https://github.com/AppleWin/AppleWin/releases/tag/v1.28.8.0)
-
-Last pre-NTSC change:
-
-* [AppleWin v1.25.0.4](https://github.com/AppleWin/AppleWin/releases/tag/v1.25.0.4)


fenarinarsa:
Done.
Also, I forgot but I changed the debugger version to 2.10.0.0. I don't know if it's okay to increment the minor version this way.

Michaelangel007:
This is a major feature so I would say it is fine to bump it to 3.0 if we accept + merge this. (Still being discussed ATM.)

fenarinarsa:
If you want a bit more information, the next features I plan to add to the debugger are the following:
- heatmap: zoom on a memory area
- heatmap: adding ROM & IO pages
- configurable layout*

* For instance, in the current 3-buffers display layout, being able to replace the heatmap with any video framebuffer (example: HGR1), and swapping the debugger (>small) with the extra display (>big). The only framebuffers that would be always displayed are the emulator output and the debugger.
And likely, adding a 6-buffers display layout.
To implement this, a rewrite of the video generators is needed to take a framebuffer as parameter.

sicklittlemonkey:
Nice work.

I like that you have a roadmap for making this more than eye candy.

Configuration of the layout is crucial so that current users can add what
they need and not have it forced upon them. ; - )

Cheers,
Nick.

tomcw:
Hi @fenarinarsa,

Some of us from the AppleWin team have been discussing this offline. We really like your enthusiasm for working on this feature, so please continue to develop it in your branch.

That said, we'd like to provide some steer to this feature, so that it's a good fit with the direction that we want to take AppleWin in.

Given this, we want to define a specification for this feature (probably in #326) of which there will certainly be overlap with your work. We'd encourage you to get involved in that discussion too. There will be some features & implementation from your work that we'd like to include (via a PR), but perhaps some parts that we'd decide weren't a good fit for AppleWin. NB. Don't take this as a criticism: everyone has their own opinions, but we just can't take everything, otherwise the product would be overwhelming!

As a general point, please bear in mind that AppleWin's debugger has many users, eg. some currently developing games, apps, demos, etc. So we need to take care how we introduce new features such that there isn't a big surprise for them... so small increments are usually the approach we take.

Let me layout some specifics here:
- the heatmap is a debugging aid, but it should be optionally enabled; not enabled by default
- to be really useful we need "pick an address" functionality with precise address granularity
  - picking this address, should be give some info about that address (see Nick's YouTube video on #326)
  - but we should be able to improve on Nick's data
- don't need to show the debugger screen tracing in realtime... but IMO there is a use-case where you are debugging (single-stepping) and want to see the heatmap
- implementation detail: performance of the 6502/65C02 emulation is important, so we should move the heatmap code to the "65D02" debugger-specific emulation (which AppleWin currently doesn't use)
- I don't like the re-sizing of the window when switching from regular emulation to debugger view... so should we have separate "floating windows"? or resize the main window only when enabling heatmap?
  - NB. a separate window allows the user to scale independently to the main debugger window (important for those with larger 2K, 4K displays).
- As Nick says above, configuration of the layout is crucial: so this need discussion & defining
- etc... 

Basically with so many changes we want a good discussion first to get some agreement on what "spec" we are aiming for.

I hope this feedback doesn't come across as negative. As you can see from Michael and Nick's comments above, there is real interest in this feature, but as I've stated, we just need to take care to deliver the right feature for AppleWin.

Thanks - Tom

fenarinarsa:
Hi Tom,

I understand perfectly all the concern ^^

To be honest I think most of the heatmap display issues can be addressed later - internally it works, there is three 64k tabs that record r/w/x on main, aux and ROM+IO. I saw Nick's video and find it amazing, I agree that's how it should work eventually. But it's mostly a matter of adding a more complex display function.
I'm not even sure the way I implemented the heatmap is the best! There may be another better way that I don't know.

Now about the other topics:

* Heatmap enabled/disabled: the display part is enabled, well, only when it's enabled. Only the record part is always enabled in 6502.h/65c02.h. The heatmap refresh where all values are decremented is done during display.

* Performance issue in general use mode: yes, there is an impact. I did a benchmark test between the current release of AW and this version. I get 539Mhz>356Mhz for full speed. The CPU performance with video refresh stays the same: 16.9Mhz. It could be improved by having a list of pointers to current active memory pages in R/W instead of testing the page for each R/W/X. That's what I'm actually about to do (adding two lists of pointers for the heatmap that are updated when page switching is done by the CPU).
That said, is emulating a 356Mhz Apple II instead of 539Mhz is really an issue? I do have a modern desktop CPU (i7@3.5Ghz) but I think that even on low-end setups it won't really show. On low-end setups, there is more issues on the video part. IMO it's better to use all the extra power we have to do something else, even if it's not used by general users, as long as they're not impacted.

* Moving to 65d02.h: I actually developped the heatmap in 65d02 first, that's where I found out that @Michaelangel007 planned to make one! I even used his macros. But then, when friends tested it, they wanted the heatmap with 6502 emulation. So I chose to copy back the macros in the general CPUs, because I think that it's better to avoid duplicating code for the CPU emulation part since a "debugger" version of 6502 would be also needed. But I completely agree that's open to discussion. Currently you can remove the macros from 6502/65c02 and switch to 65d02, the heatmap will work.

* Realtime tracing: I didn't plan to do it, because I didn't know it was possible. But it's actually a side effect of how the extender debugger work. I had a hard time working out how and when the debugger was called, and I wanted to make the less changes possible since the debugger and emulator are really huge beasts. So the emulator still refreshs the video output, and calls the debugger to refresh the console. I found out that by default it called the emulator with PC follow, and that's exactly what other emulators do:
https://www.youtube.com/watch?v=Lxd296tDdoo
(this was actually my model when I started to work on AW)
In the first version I disabled completely the debugger refresh when realtime tracing, because all the BitBlt() in PrintGlyph() made the performance awful (maybe 3-4 fps max). Once I changed that, I found out that I could remove all the extra code and let the debugger live its life :p
In the end, it's very useful, especially when you monitor zeropage in hex view.
=> It's possible to invert "G" and "GD". In that case, realtime tracing with "G" won't move the window around following PC.

* Window resizing
I did that for a simple reason: I didn't succeed in adding a new window! I simply don't know how to do it. I didn't do Win32 GUI code this level before, since I'm mostly a C#/.NET developper. All the Win32 code with physical keymap test, for instance, looks like it's coming from another century to me :)
Adding a new window also means handling all the resize issues, and there's already some. For instance, StretchBlt() doesn't handle bilinear resizing, so if you switch the A2 to monochrome, you miss half the dots in the debugger. In some case, graphics can completely disappear.
I agree this is not ideal, on my laptop the window needs to be in 1x scale, 2x scale in extend debug is too big. The problem here is that at least the emulator and the debugger are mandatory now, because when you request to display on graphic page (like GR1), the debugger doesn't switch to fullscreen emulation anymore.
But... well, I'm like you all I guess, I have only a few hours by week to spend on projects. Currently I'm working on AW, but I need to come back to 6502 soon ^^ I'm not sure I can spend whole weeks to find out how to add windows etc. My goal was really to improve AW as a developer tool.
So in the end, my choice was to resize the window and make that version, as a first usable version.
=> A solution would be to split the window in 4 and display the debugger in a 1/4 window (scaled 1x). I could add the three-splits window layout optional.

The whole window issue is a more wide one IMHO. When I asked other game/demos developers about the performnce of BitBlt(), the general opinion that I got was "why do yo use GDI to start with". They told me the best would be to open a modern context, like DirectX/OpenGL/SDL and make the code portable at the same time.
Then I asked them what would be the best UI framework to use for portable projects, and some directed me to Dear ImGUI, which was developed by a Sega Master System hacker to do a recent game remaster, and they told me it fits nicely in old projects:
https://github.com/ocornut/imgui
The idea then would be to open the same Win32 window and hide ImGui for the emulator, and make it pop inside the Win32 window when switch to debug mode. Then each display (debugger, emulator, etc) would an ImGui window.
Eventually, maybe ImGui offers to handle the native GUI context and then, yes, in that case the project can be made portable.

I think that's a bigger project, and ImGui needs testing. But that also made me think that keeping one single window no additional Win32 code might be the best compromise for future improvements.

tomcw:
Hi @fenarinarsa,

- Performance issue in general use mode

>I get 539Mhz>356Mhz for full speed. The CPU performance with video refresh stays the same: 16.9Mhz.

Actually the NTSC/RGB video rendering dominates over the CPU/6502 emu sub-system, see #788. 

>That said, is emulating a 356Mhz Apple II instead of 539Mhz is really an issue?

We still care about the time spent doing CPU(6502) emulation. If we are not strict about this, then gradually more and more "extras" will be added to the 6502 emu loop, and perf will degrade further. Also we want to demonstrating the right way to do this (from an AppleWin arch POV... ie. move to 65d02.h), so that anyone else contributing will (hopefully) see where to make their extensions.

- Moving to 65d02.h:

I'm sure we can figure out a way so as not to duplicate source code.

>Currently you can remove the macros from 6502/65c02 and switch to 65d02, the heatmap will work.

OK, sounds good.

- Realtime tracing:

That's interesting to see the C64 step-backwards POC. Something I'd like to experiment with in AppleWin... and getting easier to realise as the AppleWin save-state quality continues to improve.

And it looks like that might be a heatmap at 2:30 in the bottom right.

>In the first version I disabled completely the debugger refresh when realtime tracing, because all the BitBlt() in PrintGlyph() made the performance awful (maybe 3-4 fps max). Once I changed that,

Maybe you can put this up as a separate pull-request (PR)?
As I said last week (in #795): _Big pull-requests can be quite daunting, so ideally you'd break it up into smaller pieces._

- Window resizing

>All the Win32 code with physical keymap test, for instance, looks like it's coming from another century to me :)

And it is ;-)

> => A solution would be to split the window in 4 and display the debugger in a 1/4 window (scaled 1x). I could add the three-splits window layout optional.

This is something to discuss with the other AppleWin devs.

>The whole window issue is a more wide one IMHO. When I asked other game/demos developers about the performnce of BitBlt(), the general opinion that I got was "why do yo use GDI to start with". They told me the best would be to open a modern context, like DirectX/OpenGL/SDL

I guess there are historical reasons, eg. providing extended "support" for older Windows OS's.

>and make the code portable at the same time.

@audetto is already taking this approach in his own fork: separating frontend & backend. Then adding specialised platform support for Windows & Linux. See #538.

Personally I'm not so interesting in Linux or macOS support (and some people have had success with using Wine to run AppleWin) - but there are others on the AppleWin team that do have interest in this area.

re. ImGui

It would certainly be good to do some experimentation with this to understand the benefits.

>I think that's a bigger project, and ImGui needs testing.

Yes, definitely bigger! But maybe @audetto's approach is the right one, as it gives the flexibility to swap-in different frontends.

Anyway, please continue to chip-away at this.
And also consider raising a PR for 'BitBlt() in PrintGlyph()' faster replacement.

Thanks!

fenarinarsa:
> Actually the NTSC/RGB video rendering dominates over the CPU/6502 emu sub-system, see #788.

That's not really surprising since AW's video rendering aims to be the closest to the real thing.
It's not unique to AW, other emulators spend more time on cycle accurate video rendering, even for 16/32 bits computers. On the other hand, most other A2 emulators don't even try to be that accurate since 99.9% of available software don't need it.
Since I haven't made any modification to the video part, I'm focusing on the CPU perfs.

> We still care about the time spent doing CPU(6502) emulation. If we are not strict about this, then gradually more and more "extras" will be added to the 6502 emu loop, and perf will degrade further. Also we want to demonstrating the right way to do this (from an AppleWin arch POV... ie. move to 65d02.h), so that anyone else contributing will (hopefully) see where to make their extensions.
> 
> I'm sure we can figure out a way so as not to duplicate source code.

I just checked 65d02.h and it's currently not to be used at all, at first sight I can see that the 65C02 opcode cycles aren't correct (no 1 cycle opcode for instance). It must be updated.

>     * Realtime tracing:
> 
> And it looks like that might be a heatmap at 2:30 in the bottom right.

There is.
https://www.youtube.com/watch?v=RagLKuQBlsw

> > In the first version I disabled completely the debugger refresh when realtime tracing, because all the BitBlt() in PrintGlyph() made the performance awful (maybe 3-4 fps max). Once I changed that,
> 
> Maybe you can put this up as a separate pull-request (PR)?
> As I said last week (in #795): _Big pull-requests can be quite daunting, so ideally you'd break it up into smaller pieces._

In order to do that I must copy/paste code from that PR to another one, and I'm not sure I can maintain integrity... the PrintGlyph() optimization is not really needed for the old debugger.

> > => A solution would be to split the window in 4 and display the debugger in a 1/4 window (scaled 1x). I could add the three-splits window layout optional.
> 
> This is something to discuss with the other AppleWin devs.

Where? :D

> I guess there are historical reasons, eg. providing extended "support" for older Windows OS's.

Indeed. But there's no shame to move forward and drop old OS's support :)

fenarinarsa:
I removed the heatmap generation from 6502/65c02 and fixed 65d02.h (also added 6502 support on 65c02).

No more performance hit in general use :) but I couldn't do better than copying back the whole opcode macro lists from 6502/65c02 to 65d02.

tomcw:
@fenarinarsa - I've started by listing _just_ the requirements in #326.

tomcw:
A rough plan for getting debugger heatmap into AppleWin master:
- I have created a staging branch here: "debugger_heatmap" - https://github.com/AppleWin/AppleWin/tree/debugger_heatmap
  - This staging branch can be used to merge code in from your branch, which will allow me to make changes before merging to master.
  - Currently I have to ask you to make changes to code in your branch (as I did for your `bitblt` patch), which is inefficient / slow turnaround time.
- Merge small pieces into the debugger_heatmap branch
  - eg. the next piece could be the emulation extension you made to 65D02
- Agree spec for undecided things in #326
  - these are mainly UI related
- Implement to match the spec

tomcw:
@fenarinarsa - can you create a new PR for the CPU & 65D02 related changes?

I think this is another self-contains patch, that can be split off from the main body of work.

And when you create the PR, can you select `AppleWin:debugger_heatmap` to merge it into (not `AppleWin:master`).

I think we can rework the code so that it doesn't duplicate the 256 opcodes for 6502 & 65C02, but instead #include's the cpu6502.h & cpu65c02.h files. We'll need to changes things a little to achieve this though.

The advantage is that we'll get fixes like 34c8952 "for free", and there will be no chance of human error meaning one is updated, and the other isn't... as it current is in your branch :-)

Michaelangel007:
> I just checked 65d02.h and it's currently not to be used at all, 

That's correct.  In the `Configuration` dialog box:

* The drop-down box to select the model needs additional logic to support a new model, i.e. Apple //d -- which might be confusing -- , or
* We add a checkbox:  [x] `Use 65C02 Debugging CPU` if you have the //e model selected. <-- probably the better choice.

tomcw:
>I just checked 65d02.h and it's currently not to be used at all,

@Michaelangel007 - how I see it is that when doing AppleWin-debugging (eg. via F7), then emulation is instead routed through 65d02 (instead of 6502) and 65dC02 (instead of 65C02). So it's an implicit selection rather than the user explicitly picking Apple //d.

fenarinarsa:
> 
> 
> > I just checked 65d02.h and it's currently not to be used at all,
> 
> @Michaelangel007 - how I see it is that when doing AppleWin-debugging (eg. via F7), then emulation is instead routed through 65d02 (instead of 6502) and 65dC02 (instead of 65C02). So it's an implicit selection rather than the user explicitly picking Apple //d.

That's how I implemented it.

cpu65d02.h

```
static DWORD Cpu65D02(DWORD uTotalCycles, const bool bVideoUpdate, bool is65C02)
{
```

Later, it branches to the 6502 or 65c02 opcode huge `switch` (copy/pasted from 6502.c and 65c02.c)

then in CPU.cpp

```
static DWORD InternalCpuExecute(const DWORD uTotalCycles, const bool bVideoUpdate)
{
	if (GetDebugMode())
		return Cpu65D02(uTotalCycles, bVideoUpdate, GetMainCpu() == CPU_65C02);

	if (GetMainCpu() == CPU_6502)
		return Cpu6502(uTotalCycles, bVideoUpdate);		// Apple ][, ][+, //e, Clones
	else
		return Cpu65C02(uTotalCycles, bVideoUpdate);	// Enhanced Apple //e
}
```

So I could redefine READ and WRITE macros in 65d02.h to add heatmap debug code. In regular mode (when the debugger is inactive), the CPU emulation doesn't have any debug overhead.

Note: 65d02.h contains a bugged opcode definition, I replaced it with the current one found 65c02.

tomcw:
re. 65d02.h and the READ/WRITE macros

I've been looking at this.

I think we can get rid of 65d02.h completely and reconstruct this code using a combination of existing 6502.h/65c02.h and extra C Pre-Processor macros to include (or omit) the extra heatmap functionality. We will still end up with a (normal) 65c02 instance and a debug 65c02 instance, but both will derive from the same 65c02.h file.

I have something nearly done, so will share (via a new branch) later on.

The clear advantage to re-using existing code (ie. not duplicating code) is that we have just "one definition". The obvious problem at the moment is duplicated code, which diverge as bugs are fixed in one instance but not the other. The case in point being all the 6502 instructions in 65c02.h vs 65d02.h.

tomcw:
Here's an example of re-using 6502.h & 65c02.h to get rid of 65d02.h: 2eb7980f

tomcw:
FYI, PR #825 has now been merged into master, which removed cpu65d02.h.

Michaelangel007:
I haven't forgotten about this.

It would be nice to see _some_ of the cleanup here submitted as tiny PRs so that it easier to follow and eventually merge into trunk. i.e. magic numbers, variable cleanup, etc.

The one (minor?) concern I have is that the Debugger `CMD_NTSC` command was removed in commit 2d332c60.  I'm not sure why this was done? This provides a way to import/export a custom `.bmp` that users can fine-tine the RGB colors of the varies HGR pixels depending on which phase and position the bits are in.  It would be a shame to lose this functionality even if it is highly technical.

Another issue is scaling.  As I get older and my eyes aren't aren't as sharp as they once were I find myself wishing for a 3x scaling option for the main screen and debugger screen. While 2x is OK, the 1x is just too tiny for my eyes, especially on a 4K monitor. A 3x and 4x scaling would be nice to add to the road map.

fenarinarsa:
As far as I'm concerned, I gave up on the idea of merging this extended debugger (with all its features, ie realtime debugger, video pages, RAM access & CPU load monitoring) in trunk since there was a major rewrite of the video & graphic emulation two years ago I think, which collided with my own rewrites (specifically in `NTSC.h`, `Video.h` and `RGBMonitor.h`, which I already moved into C++ classes).
As soon as the modification were posted I tried to merge them back into my fork. But it proved impossible because a lot of files were split or moved/renamed, so it meant for me rewriting everything by hand within the new architecture. And also the new rewrite didn't even have the features I needed (like re-entrant code for the graphic render part).

And I came to the conclusion that there is more Apple II emulators out there than there is active Apple II developers, so in the end version 1.29.15 is enough for us to debug our own project or tweak non-emulated cards support in :)
Maybe doing another more modern debugger-centered emulator would be a better idea (like the visual C64 debugger), but IMO that wouldn't be of any use for the Apple II development community being really small.

As for the scaling, maybe the best way would be to switch the AppleWin render to cross-platform compatible SDL or OpenGL with Dear ImGUI, as a lot of emulators do now, instead of relying on the 1990s Windows' APIs.

tomcw:
re. scaling - being tracked in #572

audetto:
> But it proved impossible because a lot of files were split or moved/renamed, so it meant for me rewriting everything by hand within the new architecture. And also the new rewrite didn't even have the features I needed (like re-entrant code for the graphic render part).

I know the feeling. I've maintained the Linux port for a few years and I have always avoided like the plague a change in any AppleWin file as it is a guaranteed conflict.

This is the fate of projects like `linapple` which have decided to diverge and have become unmergeable...

Have you thought of designing it in a way that some parts are merged (re-entrant is potentially a valid feature) and organise the rest so there is some `#ifdef` or `interface` to dispatch to your code or to the original version?

If you find a file contains **good** and **bad** code, it can be split and you do not include the **bad** code, but you replace it with yours...

fenarinarsa:
> Have you thought of designing it in a way that some parts are merged (re-entrant is potentially a valid feature) and organise the rest so there is some `#ifdef` or `interface` to dispatch to your code or to the original version?
> 
> If you find a file contains **good** and **bad** code, it can be split and you do not include the **bad** code, but you replace it with yours...

As I said, I put a lot of time to understand AppleWin's code and include the new features, but I always said they were experimental. However they were useful in a way, since some bugs in CPU emulation were discovered, and I could add a better RGB support (even if it's not complete because some cards are harder to include into AppleWin).

Now what I think is that  instead of a debugger inside of en emulator, it's better to make an external debugger that could use an already available emulator engine on which it could be hooked. That's actually how modern debuggers in emulation work.

As a matter of fact I know at least two people around me that wrote a complete cycle-accurate Apple II emulation with Mockingboard and NTSC emulation in a few weeks. Code is available everywhere and the amazing work done in emulators like AppleWin can be reused in more modern environments. So there's room for a debugger-oriented solution.

However, as I also said, so far I think we're only two devs (max) to use the extended debugger features of AppleWin. So I don't really see what would be the interest of working hard again to re-implement them on the rewritten trunk :)

audetto:
> Now what I think is that instead of a debugger inside of en emulator, it's better to make an external debugger that could use an already available emulator engine on which it could be hooked. That's actually how modern debuggers in emulation work.

What a coincidence!

I was thinking how to write a debugger for https://github.com/cc65/cc65 and eventually found this: https://github.com/empathicqubit/vscode-cc65-debugger

It is a VSCode extension that let you debug 6502 code using cc65's symbol table via a remote connection.
The protocol is VICE's binary monitor https://vice-emu.sourceforge.io/vice_13.html which I am trying to implement for AppleWin.

PR #800, opened 2020-06-14
Improved debugger video performance

fenarinarsa:
Improved debugger video performance by removing GDI calls in PrintGlyph() and FillBackground().


tomcw:
Thanks @fenarinarsa - I'll take a look later this evening.

tomcw:
@fenarinarsa - quick question: why have you added a 2nd (identical) stdafx.h to the "Debugger" source sub-folder?

tomcw:
I wonder why the manual print of the character is a lot faster than BitBlt?

This StackOverflow article [here](https://stackoverflow.com/questions/9347532/are-the-gdi-functions-bitblt-and-stretchblt-hardware-accelerated-in-win32) talks about different Windows versions which have (or don't have) GDI h/w acceleration:

>This lists how to specify which operations are to be hardware accelerated. included in the list is BitBlt and StretchBlt, so yes they are hardware accelerated, but you need to have a system supporting it, and you need to enable it.

>WDDM v1.1 was added in Windows 7, this included the return of GDI hardware acceleration.

(GDI h/w acceleration was supported in XP, but not Vista! ...then supported again from Win7.)

See here for how to determine if the display miniport driver supports h/w acceleration:
https://docs.microsoft.com/en-us/windows-hardware/drivers/display/gdi-hardware-acceleration

>The display miniport driver also should set DXGK_PRESENTATIONCAPS->SupportKernelModeCommandBuffer to TRUE to indicate that it supports GDI Hardware Acceleration command buffer processing. 

Also see here for how to enable h/w acceleration for GDI:
https://docs.microsoft.com/en-us/windows-hardware/drivers/display/specifying-gdi-hardware-accelerated-rendering-operations

So if we were to enable GDI bitblt h/w acceleration would that then _outperform_ the manual print of the character?

Should be easy to verify...

fenarinarsa:
Actually, the fact that the hw accel is disabled by default and slow as hell *or* that it's enabled but slow as hell, screams that it's better *not* to use it.

What I was told is that BitBlt() perfs depends heavily on the driver and how it's handled by the graphic card (BitBlt is actually handled by the graphic driver, not by GDI). And by default the bitmap is in system memory, so it seems modern GPUs have to copy it to GPU memory before doing the BitBlt() and sending it back to system memory. That and the fact the support vary heavily from on OS version to the other, makes a logical conclusion that it's best to avoid it when not necessary.

In addition, let's say hw support is not enabled on my nVidia RTX 2070 (insert laugh here). Fine, so the OS fallbacks to software BitBlt. But why is it so slow? The answer is that BitBlt is a very generic function that's not really useful here, when copying a single character with a background - something you can do in one pass in a few loops on a modern CPU, instead of doing three or more system calls.

And of course, we can try to force hw acceleration, hoping it was disabled to begin with, but then on another GPU/driver/setup, hw accel may not be available, and Windows will fallback to the ludicrous slow BitBlt(). When I saw the profiling figures, PrintGlyph() took 56% of the rendering time, which made me say that the Amiga blitter was faster... even a software render on Atari ST is faster!

That's why I simply dumped those calls. I forced the bitmap format to BGRA so we can work directly in it like the emulator does, no need to make thousands of heavy system calls to print small characters or drawing small rectangles :)

In addition, and I agree that's a far goal, it's better to ditch GDI calls if any better realtime video render like OpenGL/SDL etc is considered one day :) that day there will be some low-level texture/bitmap handling involved anyway.

fenarinarsa:
I added a second stdafx.h because the project wouldn’t compile without it :/ the compiler complained about this missing file and changing the path of the include (../stdafx.h) wouldn’t work either.

tomcw:
I've not looked, but yes, `KernelModeCommandBuffer` very much sounds like there is a syscall per kick to the (kernel mode) cmd buffer. Doing this for each character is not going to be efficient!

I'd imagine batching them all up, and doing a single kick would be the thing to do.

I will do a bit of due diligence in perf testing this (but I'm not doubting your timings! :-)
This will in turn help me get a better look at the the code changes.

re. 2nd stdafx.h
I've been building AppleWin with VS2019, and didn't need this 2nd stdafx.h, so I'm sure we can solve this so we don't need a 2nd copy. I will take a look.

tomcw:
>re. 2nd stdafx.h
I've been building AppleWin with VS2019, and didn't need this 2nd stdafx.h, so I'm sure we can solve this so we don't need a 2nd copy. I will take a look.

I deleted this 2nd stdafx.h, and also removed the reference to it from the vcxproj & vcxproj.filters... and your branch still builds for me (both Debug & Release configurations).

In the project's properties (Additional Include Directories), there isn't an explicit reference to "source" (ie. the location of stdafx.h), but VS2019 must be just adding it to the include paths.

tomcw:
btw, if I add "source" to project's properties (Additional Include Directories), then it still builds for me (so no difference)... but now in files like Debug.cpp, it doesn't have a red squiggly line under `#include "StdAfx.h"`, and I can now open it too.

So probably worth changing `Additional Include Directories`:
- adding `source`
- removing `source\emulator`

EG:

Additional Include Directories = source;source\cpu;source\debugger;zlib;zip_lib;libyaml\include;%(AdditionalIncludeDirectories)

tomcw:
>I will do a bit of due diligence in perf testing this

For 1.5M (=400 x 48 x 80) `PrintGlyph()` calls:

. | Debug | Release
--|--|--
Old `BitBlt()` | 8268 ms | 8362 ms
New method | 1014 ms | 468 ms

So new routine is ~18x faster in Release config.

Test code is here (just replaces existing `DebuggerUpdate()` function): [test.txt](https://github.com/AppleWin/AppleWin/files/4789022/test.txt)

source/Debugger/Debugger_Display.cpp
```diff
@@ -654,74 +756,21 @@ void PrintGlyph( const int x, const int y, const int glyph )
...
-		BitBlt(
-			hDstDC,   // hdcDest
-			xDst, yDst, // nXDest, nYDest
-			CONSOLE_FONT_WIDTH, CONSOLE_FONT_HEIGHT, // nWidth, nHeight
-			g_hConsoleFontDC, // hdcSrc
-			0, CONSOLE_FONT_GRID_Y * 2,  // nXSrc, nYSrc // FontTexture[2][0] = Solid (Filled) Space
-			PATCOPY     // dwRop
-		);
+	// Manual print of character. A lot faster than BitBlt, which must be avoided.
+	int xx, yy;
```
tomcw, line 760: Can you move these int xx, yy into the scope of the for-statement

source/Debugger/Debugger_Display.cpp
```diff
@@ -654,74 +756,21 @@ void PrintGlyph( const int x, const int y, const int glyph )
...
+	char fontpx;
+	int index_src = (127-ySrc) * 16 * CONSOLE_FONT_GRID_X + xSrc;   // font bitmap
+	int index_dst = (383-yDst) * 80 * CONSOLE_FONT_GRID_X + xDst;   // debugger bitmap
+	for (yy = 0; yy < CONSOLE_FONT_GRID_Y; yy++) {
+		for (xx = 0; xx < CONSOLE_FONT_GRID_X; xx++) {
```
tomcw, line 765: Can you move the start brace ie "{" on a new-line so that it vertically lines up with the preceding "for".

source/Debugger/Debugger_Display.cpp
```diff
@@ -654,74 +756,21 @@ void PrintGlyph( const int x, const int y, const int glyph )
...
-			hDstDC,   // hdcDest
-			xDst, yDst, // nXDest, nYDest
-			CONSOLE_FONT_WIDTH, CONSOLE_FONT_HEIGHT, // nWidth, nHeight
-			g_hConsoleFontDC, // hdcSrc
-			0, CONSOLE_FONT_GRID_Y * 2,  // nXSrc, nYSrc // FontTexture[2][0] = Solid (Filled) Space
-			PATCOPY     // dwRop
-		);
+	// Manual print of character. A lot faster than BitBlt, which must be avoided.
+	int xx, yy;
+	char fontpx;
```
tomcw, line 761: Can you move this into the inner for-scope.

source/Debugger/Debugger_Display.cpp
```diff
@@ -825,12 +874,35 @@ int PrintText ( const char * pText, RECT & rRect )
...
+	long index_dst = (384-bottom) * 80 * CONSOLE_FONT_GRID_X;
+
+	for (long x = left; x < right; x++)
+	{
+		g_pDebuggerMemFramebits[index_dst + x].r = g_cConsoleBrushBG_r;
+		g_pDebuggerMemFramebits[index_dst + x].g = g_cConsoleBrushBG_g;
+		g_pDebuggerMemFramebits[index_dst + x].b = g_cConsoleBrushBG_b;
+	}
+
+	if (top != bottom) {
```
tomcw, line 895: Can you move this starting brace onto a new line, vertically under the if.

source/Debugger/Debugger_Display.cpp
```diff
@@ -825,12 +874,35 @@ int PrintText ( const char * pText, RECT & rRect )
...
+		bgra_t* src = g_pDebuggerMemFramebits + (index_dst + left);
+		bgra_t* dst = src + (80 * CONSOLE_FONT_GRID_X);
+		size_t size = (right - left) * sizeof(bgra_t);
+		for (int i = 0; i < bottom - top - 1; i++) {
```
tomcw, line 899: And move this starting brace too

fenarinarsa:
Thank you for testing the performance difference in PrintGlyph().
I didn't even bother to do it, with the realtime tracing it was obvious (3~4 fps versus 60fps and a lot of free CPU time).

tomcw:
@fenarinarsa - Can you also change 'Additional Include Directories' for the other 3 configurations (Debug, Debug NoDX and Release NoDX). Thanks.

@Michaelangel007 - if you want to take a look then please provide feedback today.

Otherwise I plan to merge to master this evening (after the include dir's change has been made).

tomcw:
Thanks. LGTM, so I will merge now...

source/Debugger/Debugger_Display.cpp
```diff
@@ -654,73 +738,22 @@ void PrintGlyph( const int x, const int y, const int glyph )
...
-			g_hConsoleFontDC, // hdcSrc
-			0, CONSOLE_FONT_GRID_Y * 2,  // nXSrc, nYSrc // FontTexture[2][0] = Solid (Filled) Space
-			PATCOPY     // dwRop
-		);
+	// Manual print of character. A lot faster than BitBlt, which must be avoided.
+	// (with Bitblt, realtime debug ("gd") is impossible)
+	int xx, yy;
+	char fontpx;
+	int index_src = (127-ySrc) * 16 * CONSOLE_FONT_GRID_X + xSrc;   // font bitmap
+	int index_dst = (383-yDst) * 80 * CONSOLE_FONT_GRID_X + xDst;   // debugger bitmap
```
Michaelangel007, line 761: These hard-coded x and y numbers should be cleaned up:
i.e.
* 384 should be replaced with DISPLAY_HEIGHT.

Michaelangel007:
@tomcw and @fenarinarsa Thanks for benchmarking and replacing that hideously slow BitBlt().  I'm not surprised by the OS over-head.  A good reminder to always profile, and not assume that just because the OS provides functionality does NOT imply that it will be fast.

Commits look good aside from a few hard-coded magic numbers.

tomcw:
@Michaelangel007 - yep: already got an issue covering these kind of things: #802 
These fix-ups can be deferred, as I assume if I were to make changes to Applewin:master branch, then my changes could collide with any that @fenarinarsa is making in this part of the code.

fenarinarsa:
I guess you can remove the magic numbers in PrintGlyph as soon as you can, I don't need to change it anymore. And I think it's better to close issue #802 without waiting for the heatmap to be finished as requested... the amount of free time I have is currently reducing and the summer holidays are going to start very soon.

tomcw:
NB. Magic numbers removed: e48ea095a3eea6d6c58f28bad6ed8502111dd6a1

PR #803, opened 2020-06-22
Some minor c++ correctness changes.

audetto:
Add const, static forward declaration.
Remove windows.h (already in StdAfx.h) and crtdbg.h (not needed).


tomcw:
Thanks!

Issue #806, opened 2020-06-30
Support 'Base 64A' - a Taiwanese Apple II clone


tomcw:
Spun off from #771.

Requested by @Drakepirate.

"Base 64A" is a Taiwanese Apple II clone: https://en.wikipedia.org/wiki/List_of_Apple_II_clones#Taiwan

---

My Base 64A Roms are 28K total ( D0:4K D8:4K E0:4K E8:4K F0:4K & F8:8K )
A reduced version of 12K cutting last 4k of F8 and 2k of rest of bin files partialy works but have no Applesoft prompt.

[Ci.Base-64A.zip](https://github.com/AppleWin/AppleWin/files/4845108/Ci.Base-64A.zip)

![Base64A](https://user-images.githubusercontent.com/47072448/86001360-5f232200-ba41-11ea-887c-749467add69c.png)

tomcw:
>Hi Tom
Sorry to post it here, may be could be in request area...May be Can be the next clone to be implemented in Applewin? 🙄
I have the manual and extracted the ROMS from my real machine, also a friend made a discover of ROM mapping (Base has a text processor embedded) .

![27D170C2-000C-4A50-BE04-D014F7C51DD2](https://user-images.githubusercontent.com/47072448/86075673-1dcf5880-babb-11ea-899d-cc12fa3b6fd0.jpeg)
![86075695-2a53b100-babb-11ea-8506-f7c03ded394b](https://user-images.githubusercontent.com/47072448/86091509-9eec1700-bade-11ea-9cd3-e21d1d42cbd3.jpg)

tomcw:
The extended (bank-switched) ROM needs to be properly supported, eg. during start-up with a disk in the drive, there are these AN0 accesses which (from the above screenshot) should bank switch the ROM:
- D530: STA $C059
- F7FA: STA $C059

Drakepirate:
My friend made a Base-64A emulator in GO and  it's free to use, may be it can help...

https://github.com/ivanizag/apple2

Drakepirate:
`package apple2

import (
	"fmt"

	"github.com/ivanizag/apple2/core6502"
)

/*
	Copam BASE64A adaptation.
*/

// newBase64a instantiates an apple2
func newBase64a() *Apple2 {
	var a Apple2

	a.Name = "Base 64A"
	a.mmu = newMemoryManager(&a)
	a.cpu = core6502.NewNMOS6502(a.mmu)
[...]

PR #807, opened 2020-07-01
Support 'Base 64A' - a Taiwanese Apple II clone

ivanizag:
PR for #806 

There are two commits. The first is small and adds support for up to four pages of bank-switched ROM on the area $D000 to $FFFF. It expects rom binaries that are multiple of 12Kb.

The second commit has the changes to support the Base 64A clone:

- It's an Apple II plus clone
- The bank-switched rom is controlled with AN0 and AN1
- It has four banks of ROM. Actually, even if the schematics show support for four pages on all the ROM chips, the actual device has 4 banks on F8 and 2 banks on D0, D8, E0, E8 and F0. On those, bank 2 is the same as bank 0 and bank 3 is the same as bank 1. The file BASE64A has some duplications to account for that.
- The character generation video rom has two full character sets. F10 is used to emulate the physical switch. The embeeded file has the english and german character sets.
- The way bits are encoded on the video rom is different from the Apple II plus.
- I don't have a bitmap for the run button, it uses the Apple logo.

I don't know if there is anything else that has to be done to add a new clone.

Thanks!


source/NTSC_CharSet.cpp
```diff
@@ -202,16 +204,31 @@ static void userVideoRom2K(csbits_t csbits, const BYTE* pVideoRom, const bool is
 			// UTAII:8-30 "Bit 7 of your EPROM fonts will control flashing in the lower 1024 bytes of the EPROM"
 			// UTAII:8-31 "If you leave O7 (EPROM Output7) reset in these patterns, the resulting characters will be inversions..."
 			// Apple II J-Plus: simplest logic is just invert if reading low 1K of video ROM
+			// Base64A: Bit 0 instead of bit 7
 			if (RA < 1024)
 			{
-				if (!(n & 0x80) || isApple2JPlus)
-					n = n ^ 0x7f;
+				if (type == A2TYPE_BASE64A) {
```
tomcw, line 210: Move open-brace "{" to next line.

tomcw:
Hi @ivanizag,

Thanks for the PR.
Out of interest, why does Base 64A have a German video character set?

In terms of the save-state, is there any extra state that needs saving?
I can't see anything (as `memmode` already encapsulates the AN0/AN1 state)... so just asking to check I've not missed anything.

Minor point:
re. the RUN icon - do you have any photos of the Base 64A, so that a representative icon could be created?

source/Memory.cpp
```diff
@@ -77,6 +77,10 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #define  SW_INTCXROM   (memmode & MF_INTCXROM)
 #define  SW_WRITERAM   (memmode & MF_WRITERAM)
 #define  SW_IOUDIS     (memmode & MF_IOUDIS)
+#define  SW_ALTROM0    (memmode & MF_ALTROM0)
+#define  SW_ALTROM1    (memmode & MF_ALTROM1)
+
+#define  SELECTED_ROM_PAGE       (SW_ALTROM0?1:0 | SW_ALTROM1?2:0)
```
ivanizag, line 83: This should be `((SW_ALTROM0?1:0) | (SW_ALTROM1?2:0))` . I'll fix that.

ivanizag:
Yes, `memmode` encapsulates the state. I had a doubt with `memrompages`, but it should not be needed as memrom is not in the save state either. Reviewing this, I see that `g_videoRomRockerSwitch` is not persisted. I can add that.

The Base 64A was sold in Europe with local character sets. I have the roms for Spanish and German. I chose German as that was what @DrakePirate shared on the feature request.

I'll do more testing and submit the changes this week-end.

tomcw:
>Reviewing this, I see that `g_videoRomRockerSwitch` is not persisted. I can add that.

Don't worry about this, as extending the save-state (and retaining backwards compatibility) can sometimes be tricky. I should have added this before.

source/Memory.cpp
sicklittlemonkey: Just a personal style thing, but if this macro is only used in one place, I'd inline it where it's used to avoid having to look for its definition. Maybe just int selectedRomPage = ...

sicklittlemonkey:
I haven't had time to run it, but it looks OK. Small changes.

tomcw:
I agree - I'll await those minor changes, test it with intent to merge to master.

ivanizag:
I think it is now complete and ready. I have done the minor changes and added a run icon by @Drakepirate.

To test, booting the Base 64A clone with a DOS 3.3 disk, I verified:
- On the Applesoft prompt, pressing F10 changes from "]" to "ü" back and forth.
- Also on the prompt, "TEST" runs the ROM self-test. 
- When loading a previously saved status, the clone model is set properly, the rom pages are loaded and the correct selected page is active.

I hope it's good to merge.

tomcw:
I built and booted the "Base 64A". From an Applesoft prompt, I tried the `TEST` command, but it just locks up. From the built-in debugger, the CPU has tried to execute a `hlt` instruction.. ie. halt. Any idea why this isn't working?

source/Memory.cpp
tomcw: I agree with Nick that this (single-use code) would be better inlined at the "call-site".

ivanizag:
My bad, I did the "git add ." inside the resource folder. The rest of the changes are missing. I'll push them once I am back to my Win box. Sorry.

ivanizag:
Missing changes added. I tested again with a clean clone of the fork.
The "hlt" you saw was due the mssing braces in `(SW_ALTROM0?1:0 | SW_ALTROM1?2:0)`. It is fixed now and moved inline.

tomcw:
OK, this looks fine to me.
And I've retested too, and TEST works now. So I will merge this shortly.

btw. @ivanizag - what else is in this 48K ROM?! @Drakepirate also mentioned a text processor?

ivanizag:
Thanks.

The text processor is launched with WRITER. It doesn't have much more that I know off.

It's not really 48KB ROM. Even if the motherboard could support 6 ROM chips with up to 8Kb, there is 5 of them with just 4KB that ignore the MSB of the address.  Only F8 has 8 kb. That makes a total of 28KB. I used a ROM file duplicating pages 0 and 1 into 2 and 3 when needed. That reflects how the memory is seen from the device, and more importantly, makes the UpdatePaging() code much simpler.

tomcw:
Simple is good!

I noticed that if I press CTRL+RESET during the little 4-note tune, then pressing RETURN will crash to the monitor at PC=$00B9. Maybe RESET should reset the annunciators back to 0?

btw. Are you OK with me adding you to the acknowledgements section of the AppleWin.chm file? (I assume yes, but it doesn't hurt to ask)

ivanizag:
Sure, thank you for the acknowledgement.

I have tested on the real hardware and yes, RESET is resetting to rom page 0.

tomcw:
Hmm... I reset the annunciators on RESET (b35ba0b), but still a CTRL+RESET during the 4-note tune will crash to the monitor at PC=$00B9. I guess on real hardware this doesn't happen?

ivanizag:
I can't reproduce on the real machine nor the emulator with `b35ba0b` . I have tried on both many times at different moments of the tune (it goes fast).

I was reproducing the crash to monitor with CTRL-RESET just after entering launching "WRITER" on the emulator before `b35ba0b` and now that is fixed.

tomcw:
I can reproduce this every time. Repro:
- Boot without a disk in drive-1
- Hit CTRL-RESET twice (eg. CTRL-F2)
- Return, and it'll break to the monitor at $00B9

tomcw:
Also from the WRITER menu, CTRL-RESET will break to the monitor at $03EC.

ivanizag:
Good news.
I was testing with a disk. With no disk I get the same error on the emulator and on the real hardware. It's a bug in the ROM that we emulate faithfully.  

IMG-3735

tomcw:
> It's a bug in the ROM that we emulate faithfully.

Great! Thanks for confirming this.

Drakepirate:
> Yes, `memmode` encapsulates the state. I had a doubt with `memrompages`, but it should not be needed as memrom is not in the save state either. Reviewing this, I see that `g_videoRomRockerSwitch` is not persisted. I can add that.
> 
> The Base 64A was sold in Europe with local character sets. I have the roms for Spanish and German. I chose German as that was what @Drakepirate shared on the feature request.
> 
> I'll do more testing and submit the changes this week-end.

Drakepirate:
Try to execute **CALL-11500** at Base-64A's Aplpesoft  Prompt.

yonglak:
AppleWin 1.29.14.0 clone support for Base64A:
Bug in the original rom
1) The SGN Basic function returns -5 when it should return -1, eg.  PRINT SGN(-5)

If I recall correctly, besides that annoying startup tune..
The original computer supported lower case and Basic command key entry using FUNC Key I think.
Besides Writer built into the rom, there's also a Apple shape table drawing program.

Thanks for adding this clone.

devorn:
There are other features hidden in the ROM that you may not have noticed as well.

A variation of the old monitor ROM exists in the Base 64A firmware along with built in Mini Assembler.

From the CALL-151 Monitor Prompt *

X enters the Mini Assembler It is a '>' prompt rather than the ! used by Apple.
*X
>1000: BIT C050 (enter)
1000-   2C 50 C0 BIT C050
>
>$FF69G
*

$ prefix will enter a monitor command
$FF69G will exit to monitor
*

Also the monitor contains commands only in the original (pre II Plus) Apple II (though later Apple //c firmware has this too).
addrT  will invoke Trace routine at (addr) until BRK is executed
addrS will execute Single Step one instruction from (addr). Subsequent instructions executed by S


I have not had a Base 64A since the mid 1980's, but I also recall that there is a basic comms routine program hidden somewhere in the ROM. Also the WRITER and TEST routines already mentioned by other posters.

Thanks for including the Base 64A into AppleWin! 
If somebody could scan the original manual (blue) book, that would be awesome!

devorn:
> AppleWin 1.29.14.0 clone support for Base64A:
> Bug in the original rom
> 
> 1. The SGN Basic function returns -5 when it should return -1, eg.  PRINT SGN(-5)
> 
> If I recall correctly, besides that annoying startup tune..
> The original computer supported lower case and Basic command key entry using FUNC Key I think.
> Besides Writer built into the rom, there's also a Apple shape table drawing program.
> 
> Thanks for adding this clone.

The keyboard of the Base 64A had a number of BASIC keywords silkscreened on the front of the keys (C64 style). Pressing the F2 followed by the key would enter the keyword (as opposed to pressing F2 and the key together). Do not know if this F2 key and keyword generation was implemented in the 'Base'-soft ROM or was in fact generated by the electronics on the keyboard PCB ? If the former, it may be possible to assign a function key in Applewin to emulate the Base F2 key?

tomcw:
>Do not know if this F2 key and keyword generation was implemented in the 'Base'-soft ROM or was in fact generated by the electronics on the keyboard PCB ?

@ivanizag - any idea?

ivanizag:
F2 expanding keywords is implemented on the Base ROM. Pressing F2 sends the char 127, (255 in the softswitch 0xc000.

And, as devorn said, there is lowercase support from the keyboard pcb with no mods and F1is a toggle not visible from the main board.

The numeric keypad with F1 pressed returns:
- key 9(run): 19
- key 8(up): 31
- key 7(home): 0
- key 6(right): 28
- key 5(text): 29
- key 4(left): 30
- key 3(ins): 20
- key 2(down): 10
- key 1(del): 18
- key 0(list): 19
- key .(CTRL-C): 3

There is also a DEL key, returning 24.

Here is a photo of the Base64A keybord with an europlus for reference:
base64keyboard

I don't feel confident enough to add the changes on AppleWin on a clean enough way.

tomcw:
Thanks @ivanizag - I've raised #912, as it should be pretty simple to add F2-key support.

Drakepirate:
Apart of international character ROM (mine was German) I also extracted that F2 ROM content...

Issue #808, opened 2020-07-04
Debugger: Graphic pages monitoring


fenarinarsa:
Here's a feature I had in mind when I actually started to work on the heatmap implementation: being able to monitor one or more graphic pages in realtime.

It fits with the goal of having a realtime render into the debugger, since when you have multiple windows, some commands like 'HGR', 'TEXT'... that replace the emulator output by still graphic pages don't feel right anymore.

However, unlike the heatmap implementation, this is a very big modification to existing code since the whole video renderer (Video/NTSC/RGBMonitor) must be moved into C++ classes in order to be instantiated at will.

It took me a few days to do this. Video.h and RGBMonitor.h were really simple to convert but NTSC.h wasn't. The C++ syntax is easier to manage but sometimes different from C declarations, and I had to move around many variable members declarations, which may have broken some code presentation (functional-wise, it's the same - well, it _should_).

I also had to be careful of memory management since video buffers and some arrays are now created & destroyed with the objects. It's not completely done.

There's a global Video object `g_pVideo`, which is created when AppleWin starts, then the Debugger creates one more (which is destroyed with the debugger).

Most of the static variable and functions were converted to regular ones since the instantiated renderers share only a handful of global properties, like physical switches for alt charsets. It was more secure and easy to avoid static members and since the C code is well written, everything is correctly initialized anyway.

Other than that I kept most of the code as is, even when it does callbacks to "parent objects" in the original C code (NTSC.h calls a lot of Video.h functions). I kept nearlt all the functions and variables' names. In the end, the code path is 99.9% the same since the functions didn't change.

Performance-wise, there's no difference in the general emulation mode since there's only one object with member functions that are called like in the original C code. There's a performance drop only when VS' Debug mode is activated.

I dropped one debugger command : 'NTSC' (could be rewritten later if needed).

A preview:
https://youtu.be/N4blh6NgN40

The top-left is the emulator output, rendered by the global Video object. You can see the cycle-accurate render in sync with the video beam.
The bottom-right shows the HGR page that is not selected by the PAGE2 switch - the page we are working on. It's currently a fixed output, eventually I'm going to change the "page" commands in the debugger (HGR, GR, HGR1, TEXT2, etc) to select what we want to watch. It simply calls `Video::VideoRefreshScreen()` on the second renderer with `bRedrawWholeScreen=true`.

[...]

Issue #802, opened 2020-06-19
Extra things to do after debugger heatmap work is done


tomcw:
- [ ] Remove `../` from Debugger source files' includes
All the Debugger/* source files include headers with `../` prefix. But since `AdditionalIncludeDirectories` has now been updated to include `source` then these prefixes aren't needed.

PR #812, opened 2020-07-13
Moving Video.h functions into a Video class

fenarinarsa:
Here's a proposition to move all Video.h functions & variables into a `Video` class. (for #808)

This is actually simpler than it looks. A lot of functions are unfortunately marked as entirely rewritten because VS fixed the code indentation during the process.
Also, variables definitions were moved around and gathered together.
Everything was originally set as public and *not* static, and I made some tuning afterwards.

Basically, there's a global Video object `g_pVideo`, which is created into `AppleWin.cpp` and destroyed in `Frame.cpp`.

All calls to members have been converted to `g_pVideo->member` or `Video::member` when static.

Static members include emulation global configs only, like video style (Color TV, Monochrome...) and the charset ROM rocker switch setting.
Some variables that were originally static are not anymore because they're linked to a specific `Video` instance.
Also, some members look like they could be static, but something always get in the way, especially afterwards when converting NTSC and RGBMonitor to classes. What you see here is a part of what I got after the complete conversion.

`VideoRefreshScreen()` now takes a scale argument and uses the global scale only when not specified. In my final version it also takes screen coordinates.

I created `VideoTypes.h`, which seems useless now, because `Video.h`, `NTSC.h` and `RGBMonitor.h` will share some types later and without this common header, cyclic class definitions arise.

In `NTSC.h`, you can see that the functions refer to a local `Video* pVideo` instead of `g_pVideo`. This is because `NTSC_VideoInit()`, which is called by `Video()`, makes some callbacks to its parent. It's bad but I tried to be as close as possible to the original codepath to avoid issues. During this init phase, `NTSC_VideoInit()` can't use `g_pVideo` which is not yet set... and the future `NTSC` class will need to hold a local `pVideo` reference to its parent anyway.
(NTSC.h was a challenge to convert: it's called from everywhere, even from the CPU emulation, and manipulates many pointers to the calling `Video`'s buffers).

Note that in 60Hz there's an assert warning in `NTSC_VideoRedrawWholeScreen()` when starting the emulator, but it doesn't look like it has any consequence whatsoever.


source/Video.cpp
```diff
@@ -44,317 +44,325 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
...
 static COLORREF      customcolors[256];	// MONOCHROME is last custom color
 
-static HBITMAP       g_hDeviceBitmap;
-static HDC           g_hDeviceDC;
-static LPBITMAPINFO  g_pFramebufferinfo = NULL;
+//-------------------------------------
 
-       HBITMAP       g_hLogoBitmap;
+	// NOTE: KEEP IN SYNC: VideoType_e g_aVideoChoices g_apVideoModeDesc
+TCHAR g_aVideoChoices[] =
```
sicklittlemonkey, line 85: This comment is not related to your changes (which are huge and sound good!) but there must be a better way to keep g_aVideoChoices and g_apVideoModeDesc in sync, e.g. merge both into an 2D array. m_PropertySheetHelper.FillComboBox iterates through the multi-string sending a message for each anyway.

fenarinarsa: Yes, there's a lot of enhancement that could be made to the existing code.
However I'm focusing on keeping as much as original code as possible in this PR to make the code base similar, even if it looks twisted. Optimizations and refactoring can be added later on ;)

tomcw: Agreed - this should be fixed, but let's defer these refactors to later, eg. we can keep a list in #802.

source/Video.cpp
```diff
@@ -1325,114 +1330,127 @@ void Config_Load_Video()
...
+
+void Video::RGB_SaveSnapshot(class YamlSaveHelper& yamlSaveHelper)
+{
+	RGB_SaveSnapshot(yamlSaveHelper);
+}
+
+void Video::RGB_LoadSnapshot(class YamlLoadHelper& yamlLoadHelper, UINT cardVersion)
+{
+	RGB_LoadSnapshot(yamlLoadHelper, cardVersion);
+}
```
tomcw, line 1456: Minor: add a newline at end of the file.

source/Frame.h
```diff
@@ -3,6 +3,9 @@
 // 1.19.0.0 Hard Disk Status/Indicator Light
 #define HD_LED 1
 
+#define FRAMEBUFFER_W 560
```
tomcw, line 6: Why have these constants been added to Frame.h? (where were they defined before?)
Can't they be moved to Video.h?

source/Video.h
```diff
@@ -125,118 +118,178 @@ struct WinCIEXYZ
...
+#endif
+
+const UINT kVideoRomSize2K = 1024 * 2;
+const UINT kVideoRomSize4K = kVideoRomSize2K * 2;
+static const UINT kVideoRomSize8K = kVideoRomSize4K * 2;
+static const UINT kVideoRomSize16K = kVideoRomSize8K * 2;
+static const UINT kVideoRomSizeMax = kVideoRomSize16K;
+
+
+class Video {
```
tomcw, line 177: Just a note on coding convention that member variables should be `m_...` prefix (not `g_...`), so ideally all of these should be rename to `m_...`

source/Video.h
```diff
@@ -125,118 +118,178 @@ struct WinCIEXYZ
...
+	void Util_MakeScreenShotFileName(TCHAR* pFinalFileName_, DWORD chars);
+	bool Util_TestScreenShotFileName(const TCHAR* pFileName);
+	void Video_SaveScreenShot(const VideoScreenShot_e ScreenShotType, const TCHAR* pScreenShotFileName);
+	void Video_MakeScreenShot(FILE* pFile, const VideoScreenShot_e ScreenShotType);
+
+};
+
+
+// Main Video renderer
+extern Video* g_pVideo;
```
tomcw, line 295: Added a newline at the end of the file.

tomcw:
>Note that in 60Hz there's an assert warning in `NTSC_VideoRedrawWholeScreen()` when starting the emulator, but it doesn't look like it has any consequence whatsoever.

Hmm... interesting. Definitely worth investigating, otherwise it'll bite us later!

source/Frame.h
fenarinarsa: If I'm correct they were defined in RGBMonitor. They're used in Video, NTSC and RGBMonitor and I think I moved them to Frame.h because in my alternate debugger I need them here. But other than that the can be moved where you'd like.

source/Video.h
fenarinarsa: Yes of course, and there's a lot of global variables that are going to be member variables from now. I preferred to keep the current names for now in order to be able to check differences with the original code if issues arise.
Same for functions starting with Video_..., they don't need that anymore if they're called with video->func() or Video::func().
But eventually yes, everything could be renamed.

tomcw:
@fenarinarsa - FYI, work is being done by another contributor to split out the Windows-specific code, and move it into a new "Windows" subfolder (see 2a7191f5ce658f99b0db972b81676cc2a2a50388). This means that your work here has now diverged from the master branch.

fenarinarsa:
Damn. That's a shame since I still maintain that for the French Touch version that now relies heavily on Video.h, NTSC.h and RGBMonitor.h being classes (since they're called for 2 different rendering).

I haven't maintained this branch because it was too much work to work on Video.h only as a class. The three files need to be moved into classes at the same time :/

tomcw:
IMO, it shouldn't be too hard to take the work you've done here and migrate it into the current master AppleWin codebase.

If you don't plan to work on this, then I can take a look at doing this in the coming weeks.
(I just want to avoid us both working on independently.)

Are you OK with me taking this?

fenarinarsa:
Yes :) I don't have much time those days anyway.

tomcw:
@fenarinarsa - OK, Video now exists as a class, with WinVideo deriving from it (see 3aa5750dcf2bae810c6e915c9ad73fd80858a9bc).

AppleWin.cpp contains a GetVideo() function to get the WinVideo object instance.

So I will close this PR.

PR #825, opened 2020-08-24
Remove 65d02

tomcw:
Get rid of 65d02.h completely and reconstruct this code using a combination of existing 6502.h/65c02.h and extra C Pre-Processor macros to include (or omit) the extra heatmap functionality. We will still end up with a (normal) 65c02 instance and a debug 65c02 instance, but both will derive from the same 65c02.h file.

The clear advantage to re-using existing code (ie. not duplicating code) is that we have just "one definition". The obvious problem at the moment is duplicated code, which diverge as bugs are fixed in one instance but not the other. The case in point being all the 6502 instructions in 65c02.h vs 65d02.h.


tomcw:
@Michaelangel007 - we can even remove the "$" C-preprocessor macro in the 6502 & 65C02 opcode switch statements.

Because in `DebugContinueStepping()`, during single-stepping, it calls `CheckBreakOpcode()` which optionally sets `g_bDebugBreakpointHit |= BP_HIT_INVALID`, and will then stop on an invalid opcode... ie. this is already working.

Michaelangel007:
_Almost_ good to go! One minor quibble;

* [x] commit 2eb7980
* [x] commit da2d6c3
* [x] commit 95b8ae21
* [ ] commit d5f770b -- In `source/CPU.cpp` would like to see `a` argument better named since `a` doesn't the user _what_ it does or for.
```
Heatmap_WriteByte(addr, uExecutedCycles, a);
```
* [x] commit b2d539c
* [x] commit bc7b1c40
* [x] commit 444155

Michaelangel007:
@tomcw Good to go! Thanks for the cleanup!

Issue #828, opened 2020-09-01
RGB Monitor mode: HIRES+AN3 bug


tomcw:
Spun off from PR #826, @fenarinarsa said:

On the HIRES+AN3 bug in RGB, here's what should show Digidream in RGB mode on a late RGB card (Apple/Feline/IIc adapter):

https://www.youtube.com/watch?v=YdZJa8QPuUw

Currently in AW when it goes to B&W, it doesn't come back to color when AN3 is enabled again.
Also, as you can see, there is no pixel shift to the right. In RGB there's no half-pixel shift at all, in any mode.

tomcw:
Triage & simple repro:
- set video mode = "Color (RGB Monitor)"
  - In `NTSC_SetVideoStyle()`, this sets `g_pFuncUpdateHuePixel = updatePixelBnWMonitorSingleScanline`
- boot to Applesoft, then switch to CALL -151
- `C050 N C057`  ; TEXT off, HIRES on
- `C05E N C05F N C05E N C05F` ; set FL1:FL0 = b#11 = 560 B&W mode
- `C05E` ; DHIRES on
- video goes B&W
  - In `NTSC_SetVideoMode()`, this sets `g_pFuncUpdateGraphicsScreen = updateScreenDoubleHires40` (*)
- `C05F` ; DHIRES off
- video remains B&W
  - In `NTSC_SetVideoMode()`, this sets `g_pFuncUpdateGraphicsScreen = updateScreenSingleHires40Simplified`
  - `RGB_Is560Mode() == true`, so updateScreenSingleHires40Simplified() renders in B&W

(*) NB. Unlike other modes, there is no implementation for a "updateScreenDoubleHires40Simplied()", so the update function uses this Color/B&W (NTSC) TV function instead.

The complete list of missing "RGB" (or simplified) functions are:
- updateScreenDoubleHires40Simplied()
- updateScreenDoubleLores40Simplied()

fenarinarsa:
This issue is not about DHIRES but about HIRES in RGB on the Apple card and others like Le Chat Mauve, IIc RGB adapters

To reproduce :
* set video mode to "Color (RGB Monitor)"
* boot to Applesoft, swith to `CALL -151`
* `C050 N C057 N C053` - TEXT off, HIRES on, MIXED on
* HGR should appear in color
* `C05E` - AN3 off ("DHIRES on")
* HGR should appear in B&W
* `C05F` - AN3 on ("DHIRES off")
[...]

PR #830, opened 2020-09-02
Fix for #828 (HGR/BW mode in RGB)

fenarinarsa:
Fix for #828

I split `updateScreenSingleHires40Simplified` into two functions (Color and BW) to get rid of the call to `RGB_Is560Mode()`.

There's no more link between HGR modes and F0/F1 DHIRES registers.

### DIFF
--- a/source/NTSC.cpp
+++ b/source/NTSC.cpp
@@ -1442,7 +1442,7 @@ void updateScreenDoubleLores80 (long cycles6502) // wsUpdateVideoDblLores
 }
 
 //===========================================================================
-static void updateScreenSingleHires40Simplified (long cycles6502)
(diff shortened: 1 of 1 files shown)


tomcw:
@fenarinarsa - sorry, I missed this PR, and today checked in a3f598917659816cd61bc6570bf2191ca21a0bb9 to deal with a similar issue. So I don't think this is necessary any more?

fenarinarsa:
I wondered about this exact issue two hours ago :) Let me check!

fenarinarsa:
Indeed it looks like this PR is not relevant anymore.

tomcw:
OK, in that case I will close this PR.

PR #826, opened 2020-08-25
Video7 SL7 RGB extra modes

fenarinarsa:
Added support for extra RGB modes of some video cards. (#819)

Currently only Apple RGB and Video7 SL7 are supported.

The RGB videocard can be defined through command-line:
`-rgb-card-type: apple (default), sl7, eve, feline`

The default text color can be modified when SL7 is selected. On the actual card this was defined by dip switches.
`-rgb-card-foreground <color>`
<color> can be 6 (blue), 9 (amber), 12 (green) and white (15)

Video7 SL7 extra modes (from the manual), modes 7 to 11. The others are the usual Apple RGB modes:
image

Note that mode 11 is actually incompatible with Apple RGB and other cards, which renders it as a B&W Hires.

In `NTSC` and `RGBMonitor`, some new functions were defined to handle RGB text and duochrome (F/B) 7-pixels cells. When the RGB monitor is selected, the text render does not use the regular NTSC functions anymore to be able to render color text anytime if needed.

Screenshots from the Video7 demo disk:

image
Amber text (hardware switches)

image
HGR + amber text

image
DGR + amber text

image
F/B Text - Colors are defined in the matching AUX pages

image
GR + F/B Text

image
F/B Hires - Again, colors are defined in AUX for each 7-pixels cell.

Issue #616, opened 2019-01-12
Can we have the option Color (TV Emulation) back?


tomcw:
Spun off from #357.

Mark / @6502-workshop is requesting this:

---

Hi Tom,

>My current preference is not to just reinstate the old 1.25 "Color (TV Emulation)"... although admittedly just adding it is the easiest option. Let me think about it.

Thanks for considering it! Currently there is no Apple II emulator I'm aware of which does vertical blending in a way where colors like yellow and dark blue are visible (other than AppleWIN 1.25.04 of course), it would be really great to have Color (TV Emulation) as a workaround until a better solution is implemented, such as #344.

But, I totally understand if you decide not to do it. I'm sure there are factors like technical debt, prioritization, and other things to consider.

Again, thanks for even considering it!

Mark

---

I also commented:
>I see "text optimized" similar to "vertical blending", ie. as a style or post-effect.

And Michael asked:
>I guess that raises a question -- are we going to bring back Text Optimized mode for 1.27 ?

I've no plans to at the moment. I'd rather limit the feature set to those that there's strong demand for (initially). Mark (above) nails it with "factors like technical debt, prioritization", and also increased testing matrix.

tomcw:
@6502-workshop: I'm considering adding this back in as an _experimental_ feature, as I'd like to get this old 1.25 code into the mainline before I forget too much(!).

Here's what I'm thinking:
- support "vertical color blending" (for hires only) as a video style (so not as a 1st class video type).
- when this style is enabled, then it's available only when you select the "Color (RGB Monitor)" video type.
  - so the current hires rendering (for "Color (RGB Monitor)") gets replaced by this "vertical color blending".
- initially only allow this style to be enabled via a command line switch.
  - there's no way to disable it without quitting AppleWin, then relaunching without the switch.

Does this sound useful to you at all?

tomcw:
Perhaps for the final point (about enabling via a command line switch), it'll be easy to extend the Config's UI with a checkbox...

6502-workshop:
Thanks for the update! 

Having "vertical color blending" only available via a command line switch wouldn't be that useful (I doubt many of our players would take the time to do it), but it sounds like you'd eventually connect the switch to the Config's UI and that I think would be useful. I think we can get players to tick a box in the config. 

In the long run, such as via #344, do you envision "vertical color blending" being be part of the default video mode at some point? 

Thanks again for considering this topic.

Mark

tomcw:
[...]

Issue #357, opened 2016-10-09
Can we have the option Color (Standard) back?


mediaguy:
NOOB post here ... Not sure where to post a wish to this forum.   I love this emulator.  It's the BEST!  However, I have one (personal) issue which is really a wish/complaint...

I know the latest releases attempt to better emulate the visual experience on the color TVs or monitors some of us used with our beloved Apple II.  However, I'm one of the minority, I'm sure, that doesn't love the blurred or hashed/alt scanlined emulated effects in the 1.26.xx releases.   

How can I request, humbly, that the video option Color (TV Emulation) last used in v1.25.04 be added back or perhaps an option to turn off the added visual "effects" which blurs or hashes the video graphics.  MAME provides for this to accommodate the visual preferences of its users.   I'd simply like the option to turn this off/on ... until that is added then I may have to stick with 1.25.04, but don't want to ...

So, help a NOOB out and tell me how to post a "feature wish" for future releases if this is the wrong area and if my request is crazy. :)

Thanks all.

tomcw:
Thanks for your positive feedback!
... and this _is_ the right place to post this kind of thing.

Just to check: Does the new 1.26 "Color Monitor" video mode not meet your needs?

mediaguy:
![applewin_screenshots compare v1 25 04 v 1 26 xx](https://cloud.githubusercontent.com/assets/22423271/19249449/a2122362-8f03-11e6-9dfa-e761cb8a3fad.png)

No, it's not really.  I know I will be in the minority, I suspect, but the graphics rendering has shifted a lot in v1.26.xx and I get why but I'd like the prior setting still available for folks like me.  I've attached a screenshot of comparisons to show what I'm seeing in case it's unique or not as you expected.  

The top portion (across) shows 3 snaps out of v1.25.04 from Lode Runner (text, title, and game snap)  and the bottom (across) shows 3 snaps out of v1.26.05 from same dsk file.

Hope it helps.

tomcw:
Thanks for the extra detail & screenshots. I'll discuss this further with the other AppleWin devs.

tomcw:
In the thread "AppleWin 1.26.0 (pre-release)", Steve Nickolas on cea2 posted:

19 Sep 2016:
> If I miss something, it's a "normal-ish" mode that blends color in
> graphics modes, but still shows clear text in text modes.
> 
> With the current version, I either get blended color and blurry,
> headache-inducing text (even in text mode), or sharp text and unblended
[...]

PR #832, opened 2020-09-07
RGB extramodes: SL7 fixes, DHIRES fixes, Feline support

fenarinarsa:
I used the same branch to continue my work on RGB extramodes, I hope this doesn't make the PR invalid?

Anyway, here's the new features/fixes:

- Fixed the SL7 color text features

In all regular text modes, when a default color is selected, all characters that are <0x80 shows in black and white. This was specified in the manual as "Inverse text always show in black and white"

The matching `UpdateText` functions now take the character value to handle this.

image

- Support for Le Chat Mauve Féline/IIc adapter (under Video7 patent)

Enabled with `-rgb-card-type feline`

It has the same features of the Apple RGB card except that the 160 DHIRES mode is not implemented.

The palette is different from the regular AW palette, so I added it directly into the code and it replaces the code-generated palette in that exact case. One of the main differences is that DARK_GREY and LIGHT_GREY are different colors. The colors were sampled from a white-balanced video capture done with my OSSC.

However, I will try to make more captures of other Féline and IIc adapters to further validate it.

Regular:
image

Féline:
image

- DHIRES color & mixed mode rewrite

The current AW implementation of DHIRES in RGB seemed dubious to me, so I analyzed the Video7 patent and the output of my IIc in RGB.
I rewrote it, but it was not as straightforward that I first thought.

1. Color mode is a real 140x192 mode with no color fringe (Video7 patent, shows that way on real hardware)

image

2. Mixed mode works with 4-bits cells that are organized in packs of 7 cells over 4 bytes:
Bit 7 defines if the 7 bits of that byte are color of B&W;
BW pixels are 1 bit wide, color pixels are usually 4 bits wide;
If a 4-bits color cell falls into a BW byte, then it immediatly reverts to the BW mode and the color pixel is less than 4 bits wide;
If a 4-bits BW cell falls into a Color byte, then the last BW pixel is repeated until the next color cell starts.

This was once again validated on my IIc adapter. The last one was tricky to understand. Some mixed DHGR pictures bundled with the drawing software "Extasie" didn't initially show correctly because of this. For instance on the "Elephant" picture, the left side of the Elephant shows 1-bit wide color pixels while the right side shows a column of repeated 1-bit BW pixels.

IIc:
image
AppleWin (feline card):
image

- DHIRES modes setup cleanup

The precondition tests that were present prevented Extasie to enter the mixed DHGR mode.

The Video7 patent doesn't state any precondition to enable the various DHGR modes through the hidden F1/F2 registers. However 80COL needs to be written to be detected, and its detection is reset when the card detects a C05E->C05F toggle (as previously implemented in AW).

This behavior triggers a bug in the game introduction of Prince of Persia: once it reverts to DHGR after the HGR animation with Jaffar, every following DHGR screen will show in BW (560) mode only. It's the same on real hardware.

IIc:
image

### DIFF
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -1333,6 +1333,7 @@ struct CmdLine
 		newVideoRefreshRate = VR_NONE;
 		clockMultiplier = 0.0;	// 0 => not set from cmd-line
 		model = A2TYPE_MAX;
+		rgb_card = RGB_Videocard_e::Apple;
 
 		for (UINT i = 0; i < NUM_SLOTS; i++)
 		{
@@ -1367,6 +1368,9 @@ struct CmdLine
 	VideoRefreshRate_e newVideoRefreshRate;
 	double clockMultiplier;
 	eApple2Type model;
+	RGB_Videocard_e rgb_card;
+	int rgb_card_foreground_color;
+	int rgb_card_background_color;
 	std::string strCurrentDir;
 };
 
@@ -1798,6 +1802,37 @@ static bool ProcessCmdLine(LPSTR lpCmdLine)
 		{
 			g_cmdLine.newVideoRefreshRate = VR_60HZ;
 		}
+		else if (strcmp(lpCmdLine, "-rgb-card-type") == 0)
+		{
+			// RGB video card valide types are: "apple", "sl7", "eve", "feline"
+			lpCmdLine = GetCurrArg(lpNextArg);
+			lpNextArg = GetNextArg(lpNextArg);
+
+			if (strcmp(lpCmdLine, "apple") == 0)
+				g_cmdLine.rgb_card = RGB_Videocard_e::Apple;
+			else if (strcmp(lpCmdLine, "sl7") == 0)
+				g_cmdLine.rgb_card = RGB_Videocard_e::Video7_SL7;
+			else if (strcmp(lpCmdLine, "eve") == 0)
+				g_cmdLine.rgb_card = RGB_Videocard_e::LeChatMauve_EVE;
+			else if (strcmp(lpCmdLine, "feline") == 0)
+				g_cmdLine.rgb_card = RGB_Videocard_e::LeChatMauve_Feline;
+			else
+				LogFileOutput("-rgb-card-type: unsupported type: %s\n", lpCmdLine);
+		}
+		else if (strcmp(lpCmdLine, "-rgb-card-foreground") == 0)
+		{
+			// Default hardware-defined Text foreground color, for some RGB cards only
+			lpCmdLine = GetCurrArg(lpNextArg);
+			lpNextArg = GetNextArg(lpNextArg);
+			g_cmdLine.rgb_card_foreground_color = atoi(lpCmdLine);
+		}
+		else if (strcmp(lpCmdLine, "-rgb-card-background") == 0)
+		{
+			// Default hardware-defined Text background color, for some RGB cards only
+			lpCmdLine = GetCurrArg(lpNextArg);
+			lpNextArg = GetNextArg(lpNextArg);
+			g_cmdLine.rgb_card_background_color = atoi(lpCmdLine);
+		}
 		else if (strcmp(lpCmdLine, "-power-on") == 0)
 		{
 			g_cmdLine.bBoot = true;
@@ -1943,6 +1978,8 @@ static void RepeatInitialization(void)
 
 		if (g_cmdLine.model != A2TYPE_MAX)
 			SetApple2Type(g_cmdLine.model);
+
+		RGB_SetVideocard(g_cmdLine.rgb_card, g_cmdLine.rgb_card_foreground_color, g_cmdLine.rgb_card_background_color);
 
 		if (g_cmdLine.newVideoType >= 0)
 		{
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -1333,7 +1333,9 @@ struct CmdLine
 		newVideoRefreshRate = VR_NONE;
 		clockMultiplier = 0.0;	// 0 => not set from cmd-line
 		model = A2TYPE_MAX;
-		rgb_card = RGB_Videocard_e::Apple;
+		rgbCard = RGB_Videocard_e::Apple;
+		rgbCardForegroundColor = 15;
+		rgbCardBackgroundColor = 0;
 
 		for (UINT i = 0; i < NUM_SLOTS; i++)
 		{
@@ -1368,9 +1370,9 @@ struct CmdLine
 	VideoRefreshRate_e newVideoRefreshRate;
 	double clockMultiplier;
 	eApple2Type model;
-	RGB_Videocard_e rgb_card;
-	int rgb_card_foreground_color;
-	int rgb_card_background_color;
+	RGB_Videocard_e rgbCard;
+	int rgbCardForegroundColor;
+	int rgbCardBackgroundColor;
 	std::string strCurrentDir;
 };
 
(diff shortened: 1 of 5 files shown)


tomcw:
>I used the same branch to continue my work on RGB extramodes, I hope this doesn't make the PR invalid?

Hmm... I would have assumed it'd be OK, but:
a) I can see similar commits above (e483d2e ... a684b7e) and also over in PR #826.
b) Just above it says "This branch has conflicts that must be resolved".
c) If I click 'Files Changed' (at the top of this PR), then I can see code that's already in AppleWin:master.

So you may need to fork AppleWin:master again, then cherry-pick the last 7 commits (from above) into this new branch. What a PITA! :-(

tomcw:
Hi @fenarinarsa,
This is some very good details about Féline and the RGB DHIRES MIX rendering.

>The current AW implementation of DHIRES in RGB seemed dubious to me, so I analyzed the Video7 patent and the output of my IIc in RGB.
>I rewrote it, but it was not as straightforward that I first thought.

Correct - the DHIRES in "Color (RGB Monitor)" is not for an RGB card at all! Instead it's just been carried over from older AppleWin (the last one being 1.25.0.4) before we switched to the 4x NTSC Monitor/TV/Color/B&W video modes at AppleWin 1.26.

And at 1.26 we got a lot of negative comments from users that didn't want these correct NTSC modes, but preferred the old "idealised" Apple II video modes! (See #357 for the whole 2+ year saga!) NB. The complaints were mainly focused on HIRES, not DHIRES.

The thing about this "idealised" DHIRES video mode is that aesthetically some users may prefer it. Put another way, if they're using DHIRES (eg. for a game), then they might not like this new chunky RGB DHIRES rendering (even though it's correct for RGB cards).

Anyway, it looks like you've done some good work in the new rendering for RGB DHIRES... and correctly supporting the MIXED (B&W 560) too, but we may have to consider having a switch to allow selecting your RGB DHIRES or "idealised" DHIRES.

I've not built/run it yet, but I will take a look and consider things from a user's point of view...
(Oh, and thanks for your effort!)

fenarinarsa:
> Correct - the DHIRES in "Color (RGB Monitor)" is not for an RGB card at all! Instead it's just been carried over from older AppleWin (the last one being 1.25.0.4) before we switched to the 4x NTSC Monitor/TV/Color/B&W video modes at AppleWin 1.26.
> 
> And at 1.26 we got a lot of negative comments from users that didn't want these correct NTSC modes, but preferred the old "idealised" Apple II video modes! (See #357 for the whole 2+ year saga!) NB. The complaints were mainly focused on HIRES, not DHIRES.

Oh, okay I understand ^^

Of course I can keep the existing render code, but we need to choose how it's going to be handled. I may for instance add an "applewin" RGB card that would be the default card, with the same behavior than an Apple RGB card but with the "false" RGB DHIRES. Then all the other cards will enable the real 140x192 render.

And I'm going to make another branch, cherry-pick the commits and make a new PR.

I actually wonder how the HIRES render works in RGB, it looks like there's a giant look-up table for some reason, while the RGB rendering in HIRES is actually quite simple to implement. I might check if it's currently ends with the same result than an RGB card, but so far I didn't detect anything strange.

tomcw:
>I actually wonder how the HIRES render works in RGB, it looks like there's a giant look-up table for some reason

Yep, a **very big** look up table for LORES, HIRES & DHIRES ...nearly 1MB in size!
Back in the Win95 / Intel Pentium-90 days, these look-up tables were needed to make the emulator performant.
We still carry this legacy (although it's evolved a little since 1995).

>[...]while the RGB rendering in HIRES is actually quite simple to implement.

Yeah, but this "idealised" HIRES is quite _tricky_, plus it optionally supports vertical blending (HIRES only).
We need to keep this "idealised" HIRES (as already explained re. negative comments from users at AppleWin 1.26).

>I may for instance add an "applewin" RGB card that would be the default card, with the same behavior than an Apple RGB card but with the "false" RGB DHIRES. Then all the other cards will enable the real 140x192 render.

Yes, this sounds like a good way to go.

Aside: perhaps we should also split our this "idealised" (or "false") HIRES & DHIRES functionality and move them into a different .cpp file. Not so important, but just an idea if the current RGBMonitor.cpp becomes too complicated by containing both "idealised" and true RGB rendering methods.

fenarinarsa:
> Yeah, but this "idealised" HIRES is quite _tricky_, plus it optionally supports vertical blending (HIRES only).
> We need to keep this "idealised" HIRES (as already explained re. negative comments from users at AppleWin 1.26).

What I mean is that in HIRES (I'm talking about regular HIRES, HGR), the final result of the "idealised" HIRES is actually... 99% the same than on a real RGB card.

And the algorithm to render it is actually very simple! We did implement it in GrafX2. And I checked it against the real hardware back then. It works on a 2-bit cells basis to define the "color value" of that cell and on a 3-bits basis to define the actual status of each pixel (black, white or "color value"). 

But I guess this will never be as fast as a LUT. The question is, is it more important to keep the code easy to understand, or faster?

Also, it needs some more checks to see if there isn't some case where the result looks different.

AppleWin (current release)
image

GrafX2
image

tomcw:
>What I mean is that in HIRES (I'm talking about regular HIRES, HGR), the final result of the "idealised" HIRES is actually... 99% the same than on a real RGB card.

In which case, we should just leave it as-is.
The (GrafX2) algorithm maybe simple, but how we can be sure that it's 100% the same as what's in AppleWin (which I think is unlikely).

If you look at #616 (which is not just for vertical blending, but also for some other subtle artifacting that needed resolving) - the final result from this took a fair bit of work to get right. Trying to mimic this output and prove it's 100% the same (IMO) isn't worth the effort... and it's guaranteed there will be new subtle bugs that we'd introduce (and only discover later on).

>But I guess this will never be as fast as a LUT. The question is, is it more important to keep the code easy to understand, or faster?

Both are good things to aim for :-)
My response is speed over maintainability, but only up to a point!

fenarinarsa:
> If you look at #616 (which is not just for vertical blending, but also for some other subtle artifacting that needed resolving) - the final result from this took a fair bit of work to get right. Trying to mimic this output and prove it's 100% the same (IMO) isn't worth the effort... and it's guaranteed there will be new subtle bugs that we'd introduce (and only discover later on).

Thank you for the link, it's VERY interesting.

For instance, see the title screen of Aztec.

AppleWin 1.25.0.4:

image

AppleWin 1.29.14:

image


RGB IIc adapter:

image


When looking at the ©1982 on the top left, we can see that the old rendering seems to work like the RGB cards rendering (without vertical blending, which is a feat I don't understand since it wrecks colors, but okay).

As I said before, there is not half pixel shift in RGB, so that may be the huge difference between the current RGB HIRES implementation and what appears on real RGB cards.

What I can do is going the same way than for DHIRES: making a new implementation that will be enabled only for specific RGB cards and keep the current render by default, with a "fake" "applewin" RGB card.

tomcw:
btw, the AppleWin 1.25.0.4 screenshot has vertical blending enabled, whereas the other two (AppleWin 1.29.14 and RGB IIc adapter) don't. Maybe it'd be a fairer comparison if the 1.25.04. screenshot also didn't have it enabled.

---

OK, so how about we add this new one:
-video-mode=rgb-original

which complements the existing switches:
-video-mode=rgb-monitor
-video-mode=composite-monitor  ; Nb. this is just VT_COLOR_MONITOR_NTSC

We can also add a new video mode to the Configuration drop-down: for "Color (RGB original)".

Where:
- "Color (RGB original)" (or -video-mode=rgb-original) is used for selecting the _current_ "idealised" / "fake" AppleWin LORES, HIRES, DHIRES rendering modes.
- "Color (RGB monitor)" (or -video-mode=rgb-monitor) is changed to be correct rendering for the real RGB cards/adapter.

fenarinarsa:
Looks okay to me. I'm just trying to figure out what would be a better name for the end user than RGB original/monitor which I finds confusing - I'm not even sure I would not mix them in the code.

Some things I thought of:

Color(....)

(original applewin)
RGB crisp
RGB original
RGB legacy
RGB AppleWin
AppleWin RGB crisp
AppleWin Special

(videocards)
RGB videocard
RGB exact
RGB faithful

sicklittlemonkey:
Legacy is the word for it!

inexorabletash:
Drive-by: "original" or "legacy" may not mean anything to new users. I like the "RGB videocard" suggestion for the other alternative, so giving a name that contrasts with that SGTM.

Other suggestions: simple, synthetic, idealized

For those of us who are driving by, maybe drop the whole listbox contents into a comment so we can review it as a whole? When trying to come up with UI strings I find a good exercise is to write out long thorough descriptions, then repeatedly edit down to the bare essentials.

tomcw:
>For those of us who are driving by, maybe drop the whole listbox contents into a comment so we can review it as a whole? 

Currently for AppleWin 1.29.14.0 the Config->Video modes are:
```
	TCHAR g_aVideoChoices[] =
		TEXT("Monochrome (Custom)\0")
		TEXT("Color (RGB Monitor)\0")
		TEXT("Color (NTSC Monitor)\0")
		TEXT("Color TV\0")
		TEXT("B&W TV\0")
		TEXT("Monochrome (Amber)\0")
		TEXT("Monochrome (Green)\0")
		TEXT("Monochrome (White)\0")
		;
```

Proposed:
```
	TCHAR g_aVideoChoices[] =
		TEXT("Monochrome (Custom)\0")
		TEXT("Color (RGB Idealized)\0")      ; newly added
		TEXT("Color (RGB Card/Monitor)\0")   ; was "Color (RGB Monitor)"
		TEXT("Color (Composite Monitor)\0")  ; was "Color (NTSC Monitor)"
		TEXT("Color TV\0")
		TEXT("B&W TV\0")
		TEXT("Monochrome (Amber)\0")
		TEXT("Monochrome (Green)\0")
		TEXT("Monochrome (White)\0")
		;
```

definition of 'idealized':
- (adj) regarded or represented as perfect or better than in reality.

fenarinarsa:
> Proposed:

Looks very good! I vote yes o/

tomcw:
@fenarinarsa ... just looking at this PR, and I think this one can be closed?

tomcw:
Closing this PR, as it was superseded by PR #837.

PR #730, opened 2019-11-27
RGB Monitor: speed up video update.

audetto:
Without vertical blend: 18.8 MHz -> 29.7 MHz
With vertical blend: 14.2 MHz -> 15.0 MHz

Signed-off-by: Andrea Odetti <mariofutire@gmail.com>

Issue #633, opened 2019-03-27
DHGR in 1.28.x.0


landloafer:
Graphic issues with RGB Monitor mode since 1.28.2.0. Particularly related to DHGR.

![reg](https://user-images.githubusercontent.com/39210397/55057086-1d1f6980-50a3-11e9-9f1d-14d171986c8d.PNG)
Renegate - Screen defects

![drag2](https://user-images.githubusercontent.com/39210397/55057182-65d72280-50a3-11e9-9bc8-e62cf2b9adc1.PNG)
Dragon Wars - Mixed B&W Graphic and Color Text

tomcw:
Renegade is "accidentally" enabling RGB 160x192 color mode (see #621):

All video soft-switches (excluding $C054/5) from boot:

```
Renegade:

$C056: PC=$FB36, VideoMode=00000040
$C051: PC=$FB3C, VideoMode=00000040
$C00E: PC=$CD8B, VideoMode=00000040
$C058: PC=$FA72, VideoMode=00000040 (not video)
$C05A: PC=$FA75, VideoMode=00000040 (not video)
$C05D: PC=$C2B8, VideoMode=00000040 (not video)
$C05F: PC=$C2BB, VideoMode=00000040 ; DHGR off
$C001: PC=$CA9C, VideoMode=00000048
$C000: PC=$CAB7, VideoMode=00000040
$C00C: PC=$25B7, VideoMode=00000040
$C000: PC=$25BA, VideoMode=00000040
$C056: PC=$FB36, VideoMode=00000040
$C051: PC=$FB3C, VideoMode=00000040
$C001: PC=$217E, VideoMode=00000048
$C000: PC=$219C, VideoMode=00000040
$C000: PC=$1E0D, VideoMode=00000040
$C050: PC=$1D16, VideoMode=00000000
$C057: PC=$1D19, VideoMode=00000004
$C052: PC=$1D1C, VideoMode=00000004
$C05E: PC=$1D1F, VideoMode=00000006 ; DHGR on
$C00D: PC=$1D22, VideoMode=00000007
$C000: PC=$1D25, VideoMode=00000007
$C000: PC=$1C20, VideoMode=00000027
$C000: PC=$1C20, VideoMode=00000027
$C000: PC=$1C20, VideoMode=00000027
$C000: PC=$1C20, VideoMode=00000027
$C00C: PC=$F23C, VideoMode=00000026
[...]

PR #837, opened 2020-09-24
RGB Videocards

fenarinarsa:
This is a revised version of PR #832.

New features since #832:

- New video mode, as proposed in #832
- Legacy RGB rendering is back
- On-the-fly RGB palette switch for some cards
- HIRES RGB rendering
- This is a new branch with only the most recent commits

Previous PR features:
- Fixed the SL7 color text features
- Support for Le Chat Mauve Féline/IIc adapter (under Video7 patent)
- DHIRES color & mixed mode rewrite
- DHIRES modes setup cleanup _(needs review)_

Details:

- New video mode, as proposed in #832

The available video modes are now:
```
	const char *g_apVideoModeDesc[ NUM_VIDEO_MODES ] =
	{
		  "Monochrome (Custom)"
		, "Color (RGB Idealized)"
		, "Color (RGB Card/Monitor)"
		, "Color (Composite Monitor)"
		, "Color TV"
		, "B&W TV"
		, "Monochrome (Amber)"
		, "Monochrome (Green)"
		, "Monochrome (White)"
	};
```

The new RGB supported features (extra modes, different palette etc) show only when `Color (RGB Card/Monitor)` is selected.

The new mode is referenced as `VT_COLOR_VIDEOCARD_RGB` and can be selected with the switch `-video-mode=rgb-videocard`.

- Legacy RGB rendering is back

`Color (RGB Idealized)`
This is the idealized legacy RGB mode that the users of AppleWin are used to see. This does not emulate an actual RGB card but represent an idealized clean version of the composite output.

- On-the-fly RGB palette switch for some cards

The RGB palette is now a pointer that can be used instead of the static `Pal2IndexRGB` array that contains the actual palette and then, the palette can be changed on the fly.
`VideoSwitchVideocardPalette()` does that.
Some RGB cards may use a different color palette than the regular NTSC colors, so in that case the palette is switched when the user goes in or out of the `Color (RGB Card/Monitor)` mode.

- HIRES RGB rendering

Enabled only in `Color (RGB Card/Monitor)` mode.
The main difference with other composite/ideal modes is that there is no half-pixel shift when bit 7 changes. That's why "Press any key" in this capture looks distorted. On the opposite, ©1982 looks okay in RGB but is distorted in composite.

As a matter of fact, I think that the RGB render looks a lot like what a generic Apple II emulator without precise NTSC timing handling would display.
It was checked against real hardware (Le Chat Mauve Feline card).

image

- Fixed the SL7 color text features

In all regular text modes, when a default color is selected, all characters that are <0x80 shows in black and white. This was specified in the manual as "Inverse text always show in black and white"

The matching `UpdateText` functions now take the character value to handle this.

image

- Support for Le Chat Mauve Féline/IIc adapter (under Video7 patent)

Enabled with `-rgb-card-type feline`

It has the same features of the Apple RGB card except that the 160 DHIRES mode is not implemented.

The palette is different from the regular AW palette, so I added it and it's selected when the users activates the `Color (RGB Card/Monitor)` mode. One of the main differences is that DARK_GREY and LIGHT_GREY are different colors. The colors were sampled from a white-balanced video capture done with my OSSC.

However, I will try to make more captures of other Féline and IIc adapters to further validate it.

Regular:
image

Féline:
image

- DHIRES color & mixed mode rewrite

The current AW implementation of DHIRES in RGB was not an emulation of the real RGB cards, so I analyzed the Video7 patent and the output of my IIc in RGB.
I rewrote it, but it was not as straightforward that I first thought.

1. Color mode is a real 140x192 mode with no color fringe (Video7 patent, shows that way on real hardware)

image

2. Mixed mode works with 4-bits cells that are organized in packs of 7 cells over 4 bytes:
Bit 7 defines if the 7 bits of that byte are color of B&W;
BW pixels are 1 bit wide, color pixels are usually 4 bits wide;
If a 4-bits color cell falls into a BW byte, then it immediatly reverts to the BW mode and the color pixel is less than 4 bits wide;
If a 4-bits BW cell falls into a Color byte, then the last BW pixel is repeated until the next color cell starts.

This was once again validated on my IIc adapter. The last one was tricky to understand. Some mixed DHGR pictures bundled with the drawing software "Extasie" didn't initially show correctly because of this. For instance on the "Elephant" picture, the left side of the Elephant shows 1-bit wide color pixels while the right side shows a column of repeated 1-bit BW pixels.

IIc:
image
AppleWin (feline card):
image

- DHIRES modes setup cleanup

**Since #832 I found out that it doesn't work well with Matterhorn Screamer.** I'm still scratching my head over this one.

The precondition tests that were present prevented Extasie to enter the mixed DHGR mode.

The Video7 patent doesn't state any precondition to enable the various DHGR modes through the hidden F1/F2 registers. However 80COL needs to be written to be detected, and its detection is reset when the card detects a C05E->C05F toggle (as previously implemented in AW).

This behavior triggers a bug in the game introduction of Prince of Persia: once it reverts to DHGR after the HGR animation with Jaffar, every following DHGR screen will show in BW (560) mode only. It's the same on real hardware.

IIc:
image

### DIFF
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -1751,6 +1751,10 @@ static bool ProcessCmdLine(LPSTR lpCmdLine)
 		{
 			g_cmdLine.newVideoType = VT_COLOR_MONITOR_RGB;
 		}
+		else if (strcmp(lpCmdLine, "-video-mode=rgb-videocard") == 0)
+		{
+			g_cmdLine.newVideoType = VT_COLOR_VIDEOCARD_RGB;
+		}
 		else if (strcmp(lpCmdLine, "-video-mode=composite-monitor") == 0)	// GH#763
 		{
 			g_cmdLine.newVideoType = VT_COLOR_MONITOR_NTSC;
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -2045,6 +2045,9 @@ static void RepeatInitialization(void)
 		FrameCreateWindow();	// g_hFrameWindow is now valid
 		LogFileOutput("Main: FrameCreateWindow() - post\n");
 
+		// Init palette color
+		VideoSwitchVideocardPalette(RGB_GetVideocard(), GetVideoType());
+
 		// Allow the 4 hardcoded slots to be configurated as empty
 		// NB. this state is not persisted to the Registry/conf.ini (just as '-s7 empty' isn't)
 		// TODO: support bSlotEmpty[] for slots: 0,4,5
--- a/source/Configuration/PageConfig.cpp
+++ b/source/Configuration/PageConfig.cpp
@@ -149,7 +149,7 @@ BOOL CPageConfig::DlgProcInternal(HWND hWnd, UINT message, WPARAM wparam, LPARAM
 			if(HIWORD(wparam) == CBN_SELCHANGE)
 			{
 				const VideoType_e newVideoType = (VideoType_e) SendDlgItemMessage(hWnd, IDC_VIDEOTYPE, CB_GETCURSEL, 0, 0);
-				EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (newVideoType == VT_COLOR_MONITOR_RGB || newVideoType == VT_COLOR_VIDEOCARD_RGB) ? TRUE : FALSE);
+				EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (newVideoType == VT_COLOR_MONITOR_RGB) ? TRUE : FALSE);
 			}
 			break;
 
@@ -204,7 +204,7 @@ BOOL CPageConfig::DlgProcInternal(HWND hWnd, UINT message, WPARAM wparam, LPARAM
 			CheckDlgButton(hWnd, IDC_CHECK_FS_SHOW_SUBUNIT_STATUS, GetFullScreenShowSubunitStatus() ? BST_CHECKED : BST_UNCHECKED);
 
 			CheckDlgButton(hWnd, IDC_CHECK_VERTICAL_BLEND, IsVideoStyle(VS_COLOR_VERTICAL_BLEND) ? BST_CHECKED : BST_UNCHECKED);
-			EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (GetVideoType() == VT_COLOR_MONITOR_RGB || GetVideoType() == VT_COLOR_VIDEOCARD_RGB) ? TRUE : FALSE);
+			EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (GetVideoType() == VT_COLOR_MONITOR_RGB) ? TRUE : FALSE);
 
 			if (g_CardMgr.IsSSCInstalled())
 			{
--- a/source/Configuration/PageConfig.cpp
+++ b/source/Configuration/PageConfig.cpp
@@ -149,7 +149,7 @@ BOOL CPageConfig::DlgProcInternal(HWND hWnd, UINT message, WPARAM wparam, LPARAM
 			if(HIWORD(wparam) == CBN_SELCHANGE)
 			{
 				const VideoType_e newVideoType = (VideoType_e) SendDlgItemMessage(hWnd, IDC_VIDEOTYPE, CB_GETCURSEL, 0, 0);
-				EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (newVideoType == VT_COLOR_MONITOR_RGB) ? TRUE : FALSE);
+				EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (newVideoType == VT_COLOR_MONITOR_RGB || newVideoType == VT_COLOR_VIDEOCARD_RGB) ? TRUE : FALSE);
 			}
 			break;
 
@@ -204,7 +204,7 @@ BOOL CPageConfig::DlgProcInternal(HWND hWnd, UINT message, WPARAM wparam, LPARAM
 			CheckDlgButton(hWnd, IDC_CHECK_FS_SHOW_SUBUNIT_STATUS, GetFullScreenShowSubunitStatus() ? BST_CHECKED : BST_UNCHECKED);
 
 			CheckDlgButton(hWnd, IDC_CHECK_VERTICAL_BLEND, IsVideoStyle(VS_COLOR_VERTICAL_BLEND) ? BST_CHECKED : BST_UNCHECKED);
-			EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (GetVideoType() == VT_COLOR_MONITOR_RGB) ? TRUE : FALSE);
+			EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (GetVideoType() == VT_COLOR_MONITOR_RGB || GetVideoType() == VT_COLOR_VIDEOCARD_RGB) ? TRUE : FALSE);
 
 			if (g_CardMgr.IsSSCInstalled())
 			{
--- a/source/NTSC.cpp
+++ b/source/NTSC.cpp
@@ -1569,6 +1569,36 @@ static void updateScreenSingleHires40Duochrome(long cycles6502)
 	}
 }
 
+//===========================================================================
+static void updateScreenSingleHires40RGB(long cycles6502)
+{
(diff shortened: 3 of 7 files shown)


tomcw:
Thanks Cyril, this looks like an impressive amount of investigation & work. I will take a proper look later on.

tomcw:
>This behavior triggers a bug in the game introduction of Prince of Persia: once it reverts to DHGR after the HGR animation with Jaffar, every following DHGR screen will show in BW (560) mode only. It's the same on real hardware.

For "Color (RGB Card/Monitor)", yes, we should match the real hardware behaviour.

But for "Color (RGB Idealized)", then perhaps we should not?
... which may mean that this video mode doesn't behave like an RGB Card.

The reason is that I can see AppleWin users picking this video mode, and then raising a bug because Prince of Persia turns B&W for DHGR screens.

---

Also try Renegade ([here](https://mirrors.apple2.org.za/ftp.apple.asimov.net/images/games/action/Renegade%20Disk%201%20of%202%20%281989%29%28Taito%29%28Clean%20crack%29.dsk)): for both "Color (RGB Card/Monitor)" & "Color (RGB Idealized)" this erroneously selects 160-mode (at the intro screen), then switches to HGR B&W for game-play! Clearly this game was never tested with RGB cards hence the video mode problems, but again, perhaps "Color (RGB Idealized)" should not react to these $C05E/F RGB mode changes.

NB. For Renegade, see #633. I actually fixed AppleWin for Renegade with commit 2463aae545 - see the specific change in RGB_SetVideoMode().

And in this [comment](https://github.com/AppleWin/AppleWin/issues/633#issuecomment-487394072), there's a guy using Le Chat Mauve where the Renegade intro screen is in 140, not 160 mode. Can you check Renegade on your real hardware?

fenarinarsa:
> For "Color (RGB Card/Monitor)", yes, we should match the real hardware behaviour.
> 
> But for "Color (RGB Idealized)", then perhaps we should not?
> ... which may mean that this video mode doesn't behave like an RGB Card.
>
> The reason is that I can see AppleWin users picking this video mode, and then raising a bug because Prince of Persia turns B&W for DHGR screens.

It looks like it's a more complex issue than I though. Maybe I can revert to the original behavior and work on that later.

Is there a way to log the softswitches access when running AppleWin at full speed?

> 
> And in this [comment](https://github.com/AppleWin/AppleWin/issues/633#issuecomment-487394072), there's a guy using Le Chat Mauve where the Renegade intro screen is in 140, not 160 mode. Can you check Renegade on your real hardware?

I can answer without checking on the actual hardware: there is no support for 160 mode on any Le Chat Mauve RGB card, for what I know.

tomcw:
>It looks like it's a more complex issue than I though. Maybe I can revert to the original behavior and work on that later.

Let me think about this some more... but maybe "Color (RGB Idealized)" becomes "Color (Idealized)", and so has no support for these extra RGB card video modes?

Put another way: what reasons do we have for "Color (RGB Idealized)" to support the _extra_ RGB card video modes, ie. 160 mode, B&W mode, MIX mode + Video7 modes?

fenarinarsa:
> Put another way: what reasons do we have for "Color (RGB Idealized)" to support the _extra_ RGB card video modes, ie. 160 mode, B&W mode, MIX mode + Video7 modes?

If we consider that's it's an idealized composite mode: none :)

fenarinarsa:
I had some time tonight to add support for the RGB HGR mode. Enabled only in `Color (RGB Card/Monitor)` mode.
The main difference with other composite/ideal modes is that there is no half-pixel shift when bit 7 changes. That's why "Press any key" in this capture looks distorted. On the opposite, ©1982 looks okay in RGB but is distorted in composite.

As a matter of fact, I think that the RGB render looks a lot like what a generic Apple II emulator without precise NTSC timing handling may display.

RGB:
image

Idealized:
image

Composite:
image

tomcw:
>Is there a way to log the softswitches access when running AppleWin at full speed?

No, you'll have to add your own, but it's pretty easy...

Just use the function LogOutput(...) from Log.h. This will output to Visual Studio's Output pane. Then you can just copy this text when you have completed your logging.

Alternatively you can use LogFileOutput(...) which will output to the log file (but you need to specify -l from the command line).

I tend to just add these as a temporary measure to the code I'm logging, then remove them before I commit.

tomcw:
Minor build warning:
RGBMonitor.cpp(670,6): warning C4101: 'value': unreferenced local variable

tomcw:
A minor Configuration UI thing:
Disable the "Vertical blend" check-box when "Color (RGB Card/Monitor)" is selected in the drop-down menu.
...so basically "Vertical blend" check-box is only enabled for "Color (Idealized)".

tomcw:
Cyril - this is all looking very good.

I just want to disable the extra RGB card video modes when in "Color (RGB Idealized)" mode.
... so perhaps (as you suggested) we think about renaming this to "Color (Composite Idealized)"?

fenarinarsa:
Here you go :) Let me now what you think of the current config menu.
Note that now the new video modes names are cropped in the final dialog menu, but I don't feel comfortable resizing that window, so maybe I can let you do it?

tomcw:
Thanks Cyril. All merged into master now.

btw, in RGBMonitor.cpp, there are 2 globals, only used by the function `UpdateDHiResCellRGB()`:
```
bool dhgr_lastcell_iscolor = true;
int dhgr_lastbit = 0;
```

Can you explain why they need to be global, and can't be inside the function scope?
EG. they could be static within the function scope.

fenarinarsa:
> Can you explain why they need to be global, and can't be inside the function scope?
> EG. they could be static within the function scope.

That's me not used to C static variables.
I put them there as I would do for private variables in a C++ class - which how they will end in the classes version.

So of course since they're global in this version, they can perfectly be moved as static variables inside the function scope.

tomcw:
>Note that now the new video modes names are cropped in the final dialog menu, but I don't feel comfortable resizing that window, so maybe I can let you do it?

I adjusted this in the commit b5739b862a.

re. the global vars - I just renamed them (for code consistency) and made them module-static... probably better than function-static if (for some reason) their state needed resetting eg. on a video mode change.

tomcw:
btw, I will update the AppleWin.chm [Acknowledgements](https://github.com/AppleWin/AppleWin/blob/master/help/acknowledgements.html) to credit all your RGB card/monitor work.

Let me know if you prefer your name (Cyril, as I currently have it) or your handle (Fenarinarsa) or both?

tomcw:
btw, a quick Config->Benchmark, 3 runs for each
Setup: Release build, desktop PC (Win7-64, AMD Phenom II dual-core @3ghz), 1x windowed, 50%, no mouse, no sound cards, 60Hz.

- Color (Composite Idealized), No Vertical Blend
  - Pure CPU MHz = (22.9,23.0,23.0) = 23.0 MHz (mean)  -  same as #730 (which is expected)
- Color (RGB Card/Monitor)
  - Pure CPU MHz = (15.4,16.7,15.4) = 15.8 MHz (mean)
- Color (TV)
  - Pure CPU MHz = (10.5,10.5,10.4) = 10.5 MHz (mean)

It's just useful to have a rough idea of performance of this new "Color (RGB Card/Monitor)" video mode.

fenarinarsa:
> 
> 
> btw, I will update the AppleWin.chm [Acknowledgements](https://github.com/AppleWin/AppleWin/blob/master/help/acknowledgements.html) to credit all your RGB card/monitor work.
> 
> Let me know if you prefer your name (Cyril, as I currently have it) or your handle (Fenarinarsa) or both?

You can use my real name - Cyril Lambin.

>     * Color (Composite Idealized), No Vertical Blend
>       * Pure CPU MHz = (22.9,23.0,23.0) = 23.0 MHz (mean)  -  same as #730 (which is expected)
>     * Color (RGB Card/Monitor)
>       * Pure CPU MHz = (15.4,16.7,15.4) = 15.8 MHz (mean)

As expected, since Idealized uses a LUT.

tomcw:
Hi @fenarinarsa 
When I run the FT DIGIDREAM demo, (and using command line: `-rgb-card-type sl7`), then for "Color (RGB Card/Monitor)", the screen is very corrupted when the digidrum plays! Can you reproduce this?

fenarinarsa:
> Hi @fenarinarsa
> When I run the FT DIGIDREAM demo, (and using command line: `-rgb-card-type sl7`), then for "Color (RGB Card/Monitor)", the screen is very corrupted when the digidrum plays! Can you reproduce this?

Yes, that's normal behavior since the switches that enable HIRES monochrome mode on Apple and Feline card are the same switches that enable F/B (duochrome) HIRES mode on the SL7 card ;)

That's one of the few incompatibilities between the SL7 and the more recent cards.

image

tomcw:
Thanks (and I confirmed that AppleWin only shows this corruption for the sl7 RGB card).

Issue #612, opened 2019-01-07
What is the performance cost of checking 6522 underflow each opcode?


tomcw:
In #608, Broadside's Mockingboard detection routine needed finer grain 6522 counter underflow detection than 128 cycles.

A more robust fix would be to check after each opcode, but what impact does this have on Full-speed emulation?

A more intelligent way to check underflow would be to determine when the next 6522 counter will underflow, and use this as the cycle check interval. NB. Each 6522 has 2 timers, and each Mockingboard has 2 6522's, and AppleWin currently supports 2 Mockingboards; so that's 8x 6522 timers.

sicklittlemonkey:
I've meant to bring the following topic up before ... if I'm reading the
problem correctly.

This issue is a reminder of the overly simple architecture of AppleWin with
respect to timed events.
Multiple device update functions checking the current clock against
separate counters isn't flexible.

OpenEmulator and the recent CLK (and JACE I think) have a low-level
clock-based architecture which helps handle this.
Virtu which a friend and I worked on over 20 years (in assembly, then C++,
then C#) used an efficient event queue.
Presumably the above emulators (and perhaps JACE) might also have this. The
core loop counts down to the next event.
Events are scheduled into the queue by devices with a callback - e.g.
Joysticks can schedule strobe events etc.

Such an event queue might be the most efficient and easiest thing to
"retro-fit" (bonus pun!) into the CPU.
The deprecated device update functions might not improve performance much,
but you'd get the accuracy you need.

Cheers,
Nick.

tomcw:
Yep - sounds like we're both describing the same thing, eg. an OS's kernel will have an ordered queue of hi-resolution timer events and only a single h/w timer for all of them. The h/w timer is just set to the value of the next one (ie. the next in the queue is the next count to expire).

We'd just need to check for underflow for the counter with the lowest value (after that number of cycles have executed). And the Mousecard's VBL can also be added to this ordered queue (but other "async" IRQ sources, like mouse button/movement & serial Tx/Rx can't be added).

We can handle "just in time" 6522 counter reads accurately too... actually this is already done by AppleWin (eg. the 6522 counter is only updated when a 6522 r/w access is done).

tomcw:
#608 has regressed, so the priority of this one should be raised.

But to quickly answer this:
>What is the performance cost of checking 6522 underflow each opcode (in Full-speed) ?

[...]

Issue #608, opened 2019-01-05
Broadsides Does not Detect Mockingboard without Workaround


GreatHierophant:
The game Broadsides supports the Mockingboard, but it does not detect it as it should through AppleWin v1.27.13.0.  You can get the game to detect the Mockingboard by entering the debug console and setting a breakpoint with BP, it does not matter the value, then reboot the system.

tomcw:
Thanks for reporting this.

I've reproduced with 1.27.13, 1.27.0 and 1.25.0.4 - so not a regression.
(Just set a bp at 0)

tomcw:
Here's the Broadside ISR and Mockingboard detection code:

```
; IRQ loads PC with ($FFFE) = 0xC3FA on Enhanced //e.
; Which eventually vectors to the user ISR...
ISR:
5250:8A                   TXA        
5251:48                   PHA        
5252:98                   TYA        
5253:48                   PHA        
5254:A9 FF                LDA #$FF   
5256:85 06                STA $06    ; set interrupt flag
5258:A9 40                LDA #$40   
525A:8D 0E C4             STA $C40E  ; IER = disable TIMER1 interrupt
525D:68                   PLA        
525E:A8                   TAY        
525F:68                   PLA        
5260:AA                   TAX        
5261:A5 45                LDA A5H/ACC
5263:40                   RTI        ; TIMER1_COUNTER.w is 0xFF47, and time in interrupt is 0xBA cycles
; ...return to INTCX firmware, and then finally RTI to userland after a TOTAL time in interrupt is 0x143 cycles

MockingboardCheck:
5200:A9 00                LDA #$00    
5202:85 06                STA $06     ; clear interrupt flag
5204:A9 00                LDA #$00    
5206:8D 0B C4             STA $C40B   ; ACR = one-shot mode
5209:A9 50                LDA #$50    
520B:8D FE 03             STA IRQL    
520E:A9 52                LDA #$52    
5210:8D FF 03             STA IRQH    
5213:A9 C0                LDA #$C0    
5215:8D 0E C4             STA $C40E   ; IER = enable TIMER1 interrupt
5218:A9 01                LDA #$01    
521A:8D 04 C4             STA $C404   ; TIMER1_LATCH.l = 0x01
521D:A9 00                LDA #$00    
521F:8D 05 C4             STA $C405   ; TIMER1_LATCH.h = 0x00 && TIMER_COUNTER.w = 0x0001
5222:58                   CLI         ; (2)
[...]

PR #841, opened 2020-10-04
Support for synchronous events

tomcw:
PR for #612.

Switched Mockingboard/6522 Timer interrupts & Mousecard's VBlank interrupt to use synchronous events.
This is a linked-list of ordered timer-based event, where _only the head_ of the list needs updating after every opcode.

Improvements:
- 6502 interrupts are now cycle accurate in full-speed (unthrottled) mode (eg. fixes Broadsides, #608)
  - Also FT demos (eg. MAD2) can be toggled full-speed on/of without losing the video/6522 sync.
- The emulation performance is now unaffected by having Mockingboard 6522 timers, as it's not having to update the 4x 6522 timers after every opcode!

Tested with _all_ FT demos (eg. "CHIP" uses 6522-A & 6522-B timer1 interrupts during the video frame, "TRIBU" re-uses the 6522-A timer1 interrupt as it scans down the video frame). Also tested with Dazzle Draw (for Mousecard VBlank).

### DIFF
--- a/AppleWinExpress2008.vcproj
+++ b/AppleWinExpress2008.vcproj
@@ -758,6 +758,14 @@
 					>
 				</File>
 				<File
+					RelativePath=".\source\SynchronousEventManager.cpp"
+					>
+				</File>
+				<File
+					RelativePath=".\source\SynchronousEventManager.h"
+					>
+				</File>
+				<File
 					RelativePath=".\source\Tape.cpp"
 					>
 				</File>
--- a/AppleWinExpress2019.vcxproj
+++ b/AppleWinExpress2019.vcxproj
@@ -96,6 +96,7 @@
     <ClInclude Include="source\Speech.h" />
     <ClInclude Include="source\SSI263Phonemes.h" />
     <ClInclude Include="source\StdAfx.h" />
+    <ClInclude Include="source\SynchronousEventManager.h" />
     <ClInclude Include="source\Tape.h" />
     <ClInclude Include="source\Tfe\Bittypes.h" />
     <ClInclude Include="source\Tfe\Bpf.h" />
@@ -184,6 +185,7 @@
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">Create</PrecompiledHeader>
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Release NoDX|Win32'">Create</PrecompiledHeader>
     </ClCompile>
+    <ClCompile Include="source\SynchronousEventManager.cpp" />
     <ClCompile Include="source\Tape.cpp" />
     <ClCompile Include="source\Tfe\Tfe.cpp">
       <PrecompiledHeader Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">NotUsing</PrecompiledHeader>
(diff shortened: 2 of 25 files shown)


tomcw:
Hi Nick / @sicklittlemonkey - can you take a look at this PR if you have time?

The key change is this new linked-list for synchronous events:
https://github.com/AppleWin/AppleWin/blob/GH612-synchronous-events/source/SynchronousEventManager.cpp

This is something we discussed here: https://github.com/AppleWin/AppleWin/issues/612#issuecomment-452173787

The head of the list gets updated after every opcode in the 6502/65C02 switch statement - see the `Update()` function. This replaces the hardcoded calls to Mockingboard & Mousecard to update timers and generate interrupts.

My definition of a synchronous event is for a deterministic event that will occur in N cycles' time, eg. 6522 Timer & Mousecard VBlank. (As opposed to async events, like SSC Rx/Tx interrupts.)

Live events can also be removed - necessary, since 6522 timer events can be started, and then cancelled later, eg. the periodic timer, which generates an interrupt every N cycles. If you change the period, then they need to be removed & re-added.

tomcw:
btw, I realised in the 6502/65C02 opcode switch-statement, that the call to `NTSC_VideoUpdateCycles()` after every opcode, could probably be replaced by a sync event for every video line (65 cycles). This would simplify things in this switch-statement loop. And may improve performance too (`NTSC_VideoUpdateCycles()` called ~21 times per video line to just once).

I'd need to cope with demos that change the video mode across the horizontal line (eg. FT demo ANSI STORY & Sather's Little Text Window), but think this is totally doable too by just splitting up the 65-cycle sync event on a video soft-switch access.

Something to investigate after this PR has been merged in.

sicklittlemonkey (approved):
Hey, looks good.

Quite an involved change with the mouse SyncEvent handler calling NTSC_GetCyclesUntilVBlank etc. I just skimmed over the MB stuff - trust you there! ; - )

Thinking back to how my mate Sean did this in the past, we managed to optimize away the "if (!pCurrEvent) return;" in SynchronousEventManager::Update by always having at least one sentinel event in the queue - which was the next VBL when we handled it this way. I suppose you could also have a dummy event as a sentinel, but then you decrement the cycles so maybe the null check is cheaper.

The NTSC_VideoUpdateCycles change sounds interesting. It might be faster.

Hmm! Fun stuff!

Cheers,
Nick.

tomcw:
>Thinking back to how my mate Sean did this in the past, we managed to optimize away the "if (!pCurrEvent) return;" in SynchronousEventManager::Update by always having at least one sentinel event in the queue

I suppose if I did get an advantage from doing NTSC video lines updates every 65 cycles, then there would always be an event in the queue, so I could also optimise away the "if (!pCurrEvent) return;".

sicklittlemonkey:
Yeah, I was thinking the same thing. Every little bit helps.

fenarinarsa:
I just saw this PR because I'm merging all recent commits to master into another branch :D

Does it mean that it could be used to easily implement the VSync interrupt of the IIc?

From what I could gather on my actual IIc, the IRQ occurs exactly when VBlank starts, at the end of the visible area of the last displayed line.

Maybe it's the same than for the mousecard Vblank interrupt?

tomcw:
>From what I could gather on my actual IIc, the IRQ occurs exactly when VBlank starts, at the end of the visible area of the last displayed line.

Yes, that's the same as the mousecard's VBlank interrupt.

PR #842, opened 2020-10-07
Source clean-up, portability and debugger fixes.

ThorstenBr:
Another set of commits fixing compiler warnings and portability issues.
While the changes may not be required for MSVC (some will depend on the compiler warning lelvel) , they should not hurt - and even provide some clean-up.
It's mainly issues reported by gcc, such as some naming conflicts of local variables with gcc standard library symbols. It would really help/ease the synchronization of sources with the Linux port, if you could take these minor changes upstream. 

The final two commits fix two general issues:
* confusing behavior when entering the builtin (mini) assembler mode,
* and also fixes the "cl" (clear flag) command - which was documented, but did not work at all (due to an "interesting" command/syntax conflict...).

### DIFF
--- a/source/CPU.cpp
+++ b/source/CPU.cpp
@@ -428,7 +428,7 @@ static __forceinline void IRQ(ULONG& uExecutedCycles, BOOL& flagc, BOOL& flagn, 
 		PUSH(regs.pc & 0xFF)
 		EF_TO_AF
 		PUSH(regs.ps & ~AF_BREAK)
-		regs.ps = regs.ps | AF_INTERRUPT & ~AF_DECIMAL;
+		regs.ps = (regs.ps | AF_INTERRUPT) & (~AF_DECIMAL);
 		regs.pc = * (WORD*) (mem+0xFFFE);
 		UINT uExtraCycles = 0;	// Needed for CYC(a) macro
 		CYC(7)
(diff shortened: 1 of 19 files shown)


tomcw:
@ThorstenBr thanks again for these clean-ups.

I wonder if we don't just get rid of the `cl ...` command altogether, as there are many other ways to (set or ) clear processor flags, eg:

1. `r p <value>`
   - actually this is documented in AppleWin.chm, but doesn't work!
   - and (as an aside) the debugger's built-in help says `r sp <value>` but actually it's `r s <value>`!
2.  `clc`, `clz`, `cli`, ...
3. `rc`, `rz`, `ri`, ...
4.  SHIFT+Mouse Left Button click on the GUI's processor flags

@Michaelangel007 - any opinion on removing the `cl` command? (which hasn't worked for years) and also remove `se` too.
I'd also suggest removing the legacy `rc`, `rz`, `ri`, ... & `sc`, `sz`, `si`, ... too!

tomcw (approved):

tomcw:
@Michaelangel007 - I'm not saying that this PR is dependent on removing these commands, but just something to consider doing post-commit.

tomcw:
Thanks @ThorstenBr - I've now merged this into master.

tomcw:
>Please keep the CLx clear register and SEx commands. ... Being able to use the 6502 mnemonics makes this a QoL.

I agree - see #844, where it's a bit clearer what the intent is.

PR #847, opened 2020-10-18
More clean-up, portability and another debugger fix

ThorstenBr:
Again, another set of minor fixes for compiler warnings etc, making AppleWin more portable.
Most changes are no-brainers, separated into individual commits for clarity.

The only functional change is another debugger fix for the data window. When entering the debugger and switching to the data window ("data" command in the console) then the cursor keys did not work. They only worked, once the mini dump #1 in the right panel was also activated ("md1" command) - which made no sense. When the data is visible in the main window, then the cursor keys should also work.

Commit comment concerning g_uMaxExPages/g_uActiveBank below is not quite accurate: they are still used, of course, but declared "static" and now local within Memory.cpp, which contradicts their "extern" declaration in the header (and gcc does not like that).

### DIFF
--- a/source/Debugger/Debug.cpp
+++ b/source/Debugger/Debug.cpp
@@ -3521,13 +3521,10 @@ void _CursorMoveDownAligned( int nDelta )
 {
 	if (g_iWindowThis == WINDOW_DATA)
 	{
-		if (g_aMemDump[0].bActive)
+		if (g_aMemDump[0].eDevice == DEV_MEMORY)
 		{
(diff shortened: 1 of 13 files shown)


Michaelangel007:
@tomcw Please review.

These commits look fine to me to merge.

tomcw:
@Michaelangel007 - Yes, this all looks fine to me too.
I can merge (and test) later this week... but feel free to go ahead.

tomcw:
@ThorstenBr thank for this - I've now merged this into master.

Issue #691, opened 2019-09-16
Save State Default Folder


exoscoriae:
When using command line to launch applewin with a specific dsk\woz\etc file, pressing f11 to save state will default to the last folder you loaded an image from rather than the current one.

eg: if I run a command line to load choplifter, then exit, then run a command line to load championship lode runner, and then press f11, it will open the choplifter folder.

I am currently working around this by launching the current game twice. Once with "-screenshot-and-exit" and then again. By launching it twice, it then defaults to the last folder, which happens to be the same as the current folder.

tomcw:
Thanks for creating a new issue for this.
(Perhaps similar to #663.)

exoscoriae:
I may be misunderstanding #663 but my relative paths are working with the -d1 flag. But yes, further down there is talk of setting the current directory, so I imagine it would directly relate to that.

tomcw:
From #790, @Harbinger-365 said:

>While looking into this I think I found whats causing #691 as well. I think its something to do with when the "starting directory" line is being updated when using the -d1\-d2 command line switch. When you load applewin and press f11 or f12 it will open whatever folder was in the "starting directory" line before applewin loaded, not the folder of the game currently loaded. This makes me think that whatever is behind f11\f12 is looking at the "starting directory" line before it has been updated to the current path. I currently get around this by loading a game, quiting, and loading it again in my bat file. Its even worse when using the -h1\-h2 command line switch as it never updates the "starting directory" line at all due to using "HDV Starting Directory" so even on a subsequent load f11\f12 will not load the folder of the current game. For some reason F3\F4 is not affected by this issue and will always open the current folder.

exoscoriae:
Seems as though we came up with similar work arounds.

tomcw:
@exoscoriae - there's a new AppleWin 1.29.14.0 [here](https://github.com/AppleWin/AppleWin/releases/download/v1.29.14.0/AppleWin1.29.14.0.zip), which contains an update for this issue. Can you give it a try?

Python-Exoproject:
Hi @tomcw ,

I have tested the latest release for the save\load state issue I had. Results as follows:

-------------------------------------------------------------------------
Using Load State (F12): Seems to be no change on this one for my issue, when loading a disk image with -d1 and go straight to using F12 to load save state the shown folder is still that of the last -d1 image loaded before the current one. As before if current image is loaded with -d1 then exiting and loading again updates the behind the scenes value and it will be correct. For example:

[...]

PR #849, opened 2020-10-23
Improve determining path & filename when saving/loading a save-state

tomcw:
PR for #691.

Be smarter about both the path & filename when saving/loading a state-state file:

On save:
- Filename = (in priority order) try: S7D1-HDD or S7D2-HDD or S6D1-floppy or S6D2-floppy or S5D1-floppy...
  - if still empty, then use last save-state name
  - if still empty, then leave it empty
- Path = (in priority order) try: S7D1-HDD or S7D2-HDD or S6D1-floppy or S6D2-floppy or S5D1-floppy...
  - if still empty, then use last save-state path
  - if still empty, then use `g_sCurrentDir`

On load:
- Filename = last save-state name
  - if empty then (in priority order) try: S7D1-HDD or S7D2-HDD or S6D1-floppy or S6D2-floppy or S5D1-floppy...
  - if still empty, then leave it empty
- Path = last save-state path
  - if still empty, then use `g_sCurrentDir`

### DIFF
--- a/help/CommandLine.html
+++ b/help/CommandLine.html
@@ -61,7 +61,7 @@
 		Emulate a RamWorks III card with 1 to 127 pages (each page is 64K, giving a max of 8MB) in the auxiliary slot in an Apple //e machine.<br><br>
 		-load-state &lt;savestate&gt;<br>
 		Load a save-state file (and auto power-on the Apple II).<br>
-		NB. This takes precedent over the -d1, -d2, -d#s#, -h1, -h2, s0-7, -model and -r switches.<br><br>
+		NB. This takes precedent over the -d1, -d2, -s#d#, -h1, -h2, s0-7, -model and -r switches.<br><br>
 		-f<br>
 		Start in full-screen mode<br><br>
 		-fs-height=&lt;best|nnnn&gt;<br>
--- a/source/Applewin.cpp
+++ b/source/Applewin.cpp
@@ -1179,6 +1179,12 @@ static bool DoDiskInsert(const UINT slot, const int nDrive, LPCSTR szFileName)
 {
 	Disk2InterfaceCard& disk2Card = dynamic_cast<Disk2InterfaceCard&>(GetCardMgr().GetRef(slot));
 
+	if (szFileName[0] == '\0')
+	{
+		disk2Card.EjectDisk(nDrive);
+		return true;
+	}
+
 	std::string strPathName = GetFullPath(szFileName);
 	if (strPathName.empty()) return false;
 
@@ -1191,6 +1197,12 @@ static bool DoDiskInsert(const UINT slot, const int nDrive, LPCSTR szFileName)
 
 static bool DoHardDiskInsert(const int nDrive, LPCSTR szFileName)
 {
+	if (szFileName[0] == '\0')
+	{
+		HD_Unplug(nDrive);
+		return true;
+	}
+
 	std::string strPathName = GetFullPath(szFileName);
 	if (strPathName.empty()) return false;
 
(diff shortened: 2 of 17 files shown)


Python-Exoproject:
For load you say:

Filename = last save-state name
Path = last save-state path

Would this be an issue if I run (via command line using an ini for settings) game A, save it, quit, run game B, save it, quit, go back to game A, try to load? The logic says to me in that case it would give me the path and name of game B when what I want is game A?

tomcw:
You're right! ...and you did state this workflow the other month in #691.
NB. the above logic does resolve the original comment when #691 was raised (relating to path shown when pressing F11).

Anyway it seems like a tweak to the above logic to include updating the "last save-state path" on a disk (or harddisk) insertion (both GUI & CUI) should be enough.

tomcw:
Test cases:

- From command line:
  - insert -d1 pathA\A.dsk, save the state (F11) as default (A.aws.yaml) and quit AppleWin
  - insert -d1 pathB\B.dsk
  - save/F11 should offer pathB\B.aws.yaml, then cancel save
  - load/F12  should offer pathB\B.aws.yaml, then cancel load and quit AppleWin

- From GUI:
  - insert (drive-1) pathA\A.dsk, save the state (F11) as A2.aws.yaml and quit AppleWin
  - restart AppleWin
  - save/F11 should offer pathA\A2.aws.yaml, then cancel save
  - load/F12 should offer pathA\A2.aws.yaml, then cancel load and quit AppleWin

- From GUI:
  - insert pathA\A.dsk, save the state (F11) as A2.aws.yaml
  - enable HDD, and plug-in (harddrive-1) pathC\C.2mg, then OK... confirm the emulator restart
  - save/F11 should offer pathC\C.aws.yaml, then cancel save
  - load/F12 should offer pathC\C.aws.yaml, then cancel load
  - disable HDD, then OK... confirm the emulator restart
  - save/F11 should offer pathA\A2.aws.yaml, then cancel save
  - load/F12 should offer pathA\A2.aws.yaml, then cancel save and quit AppleWin

tomcw:
New logic...

On any harddisk or disk insertion (eg. from command line, GUI or save-state):
- Path & Filename = (in priority order) try: S7D1-HDD or S7D2-HDD or S6D1-floppy or S6D2-floppy or S5D1-floppy...
- if path is same as current save-state path then do nothing
  - eg. this is for when the user isn't using the default save-state name
- else set new save-state's path & filename

When saving/loading a state-state file (NB. treat saving & loading the same)
- path = save-state's path
  - if still empty, then use `g_sCurrentDir`
- file = save-state's filename
- on a successful save or load of a save-state then:
  - set new save-state's path & filename
  - save this pathname to Registry (or conf.ini)

Let me attach a pre-release build of AppleWin so you can test it (coming in the next comment)...

tomcw:
@Python-Exoproject , @exoscoriae: here's a pre-release build: [Applewin1.29.14.1.zip](https://github.com/AppleWin/AppleWin/files/5433467/Applewin1.29.14.1.zip)
Please test this build (just unzip over the top of your 1.29.14.0 AppleWin folder), and let me know if it works for your cases.

Python-Exoproject:
I get this small issue when using "load game". The default load name is just the last 6 characters of my image name (Game disk image is called "2400 A.D. (1988) side A.dsk"). If I then click out of the AppleWin window and back into the window the name updates with the full image name.

image

Other than that, as far as I can tell from my testing its all good!

Python-Exoproject:
Just as a side question is there a way to deal with the last floppy that was booted still being listed when the next game is a hard drive game? I.E. In the below image a hard disk image with "Eamon" was booted, but the last game loaded was the floppy of "Choplifter!". You can still see the "Choplifter!" image is loaded into the disk 1 slot. Its not a big deal, and doesnt cause any problems playing, just wondering if there is a command line argument I havent noticed that might remove it when loading the hard disk game.

image

tomcw:
>I get this small issue when using "load game".

I noticed this too. The full filename is there, but only the end part is displayed.
Searching, it appears to be a Win32 API bug:
https://community.intel.com/t5/Intel-Fortran-Compiler/GetOpenFileName-default-file-display-problem/td-p/1107308

EG. Press the "Home" key or Control+Left Arrow to reveal the name from the beginning.

NB. This is the same behaviour as in previous versions of AppleWin... so not a regression.

Python-Exoproject:
Ah, your right. All good then, its working as expected. Thank you very much for your assistance on all of this Tom.

tomcw:
>As an aside to the issue above, is there a way to deal with the last floppy that was booted still being listed when the next game is a hard drive game?

Two ways that I can think of:

1. applewin.exe -d1 ""
   - but this shows an warning message box ("Failed to insert floppy disk(s) - see log file") that the user has to click OK.
2. applewin.exe -s6 empty
   - this removes the floppy controller from slot-6. It is a temporary thing, and doesn't persist the state to the Registry. But it means that since there's no floppy controller, then no floppies can be inserted during this AppleWin session.

I will improve (1) such that -d1 "" (or -d2,-h1,-h2 "", etc) eject/unplug the disk/harddisk without any msg box.

Python-Exoproject:
Much appreciated

PR #861, opened 2020-11-07
Cumulative changes required to enable / ease compilation in Linux with gcc.

audetto:


tomcw:
In general, this looks fine from my POV.

At some later date (not this pull-request) we could change the `namespace DIMouse` to a class, and then instantiate it alongside the other global singleton classes in AppleWin.cpp. I don't see any strong need to do this - perhaps driven more from an code asthetics / consistency view point.

NB, you only updated the Visual Studio 2017 proj files. I guess you are only using VS2017?
I will fix-up the VS2008 & VS2019 proj files post-merge.

I only use VS2008 (still for AppleWin releases!) and VS2019. I'll raise a new issue proposing we drop VS2013, VS0215 & VS2017 supported sln/vcproj files... this'll mean that you (and others) will need to switch to VS2019.

audetto:
An NP-Hard problem, Microsoft has not yet found a solution for, and when they do, someone will patent it! :-)

Sure, happy to move.

tomcw:
@audetto - can you explain this commit:
>* Allow to include stock unzip.h from minizip.

I see from the [zlib1211.zip](https://zlib.net/zlib1211.zip) that unzip.h is located here: `contrib\minizip\unzip.h`
But I don't understand why you can't just use the unzip.h that's included in AppleWin?

audetto:
In a world where libraries are available at zero cost it is easy to reuse.
Rather than adding 3 more projects yaml, zip, zip_lib I have just used the ones already shipped with the os.

I see one file (iowin32.c) uses <windows.h> which means more work on the compatibility layer...: all in all reusing is a win-win situation in linux (add on top of that you get bug fixes...)

But I am sure one could ship these libs with AppleWin and use them in linux too.

audetto:
But you are right

file is supposed to be included from minizip

andrea@bomba:~$ pkg-config minizip --cflags
**-I/usr/include/minizip**

I will update the PR.

audetto:
Updated 2019.vcproj and removed non necessary minizip change.

PR #866, opened 2020-11-11
Ensure all header files can be included after stdafx.h.

audetto:
This avoid the needs to include them in a particular order.
Make sure they have a "#praga once" at the top too (make this uniform, even if not always strictly needed).

If one sticks to the convention that a cpp file called xyz.cpp should inlcude xyz.h immediately after stdafx.h, then all header files will be always self sufficient.


tomcw:
>If one sticks to the convention that a cpp file called xyz.cpp should inlcude xyz.h immediately after stdafx.h, then all header files will be always self sufficient.

Yes, I understand.

We should add this rule (and the #pragma rule) to the CodingConventions.txt file... which itself could do with some updating! I'll take this action.

This PR all looks fine to me.

Issue #864, opened 2020-11-11
Anti-M 1.8 doesn't work


Keatah:
https://github.com/a2-4am/anti-m

Just switches to drive 2 and sits there. Don't know if it's a problem with the program itself or AppleWin. But it does work on older versions of AppleWin.

tomcw:
From the instructions on that anti-m page:

>To disable auto-boot and show the prompt, even if Anti-M would normally skip it, you can press either joystick button (open-apple or closed-apple key) while launching Anti-M.

This at least is a workaround.

tomcw:
Equally you can put the original disk you want to boot into drive-2 and just let Anti-M automatically boot it.

tomcw:
OK, the change in behaviour of AppleWin happens at 1.29.8.0, probably due to this change:
- [Bug #748] DiskII: data latch returns a rand() value when no disk is in drive.

At 1.29.7.0 it drops to the Anti-M menu, but at 1.29.8.0 it switches to drive-2 and spins forever.
NB. It's the same with Anti-M 1.4, 1.5, 1.6, 1.7 and 1.8.

tomcw:
#### Anti-M

@peterferrie said:
>It's intended to detect a disk in drive 2. If there's either no drive, or no disk in the drive, then the returned value should be a constant (might not even be a full nibble)

```
20E3: ; CheckDrive2
  LDA $C0E9 ; enable motor on
  LDA $C0EB ; enable drive-2
  ; delay $2683D cycles = 0.157s
  ; Loop 256 times:
  ;   A = A EOR $C0EC
  LDA $C0E8 ; enable motor off
  CMP #0
  RTS
```

I guess the delay stabilises the value in the latch.

NB. AppleWin doesn't support a DiskII card with drive-1, but no drive-2.

So possible emulation logic is:
- After motor on & new drive enable, then:
  - if less than 0.150s then latch returns a random value ; Wizardry IV / V case (meaning drive connected)
  - afterwards latch returns a fixed value ; Anti-M case (meaning drive connected, but no disk)

sicklittlemonkey:
It's your call, but this is getting into some pretty case-specific
behaviour, further and further away from the real hardware.

The Wizardry disk check looks annoying - it seems you must have a disk in
drive 2 for it to be activated. The reason a DSK image doesn't work is
because that code doesn't return partial nibbles, right? (If that's not the
case any more then you can stop reading! : - ) One solution is for DSKs to
[...]

PR #870, opened 2020-11-14
Improved support for latch when drive is empty

tomcw:
PR for #864.


tomcw:
Hi Nick / @sicklittlemonkey - if you can, then take a quick look at this PR for the Anti-M / Wizardry V support for empty drive detection.

As proposed in #864, the logic change is (for an empty drive) what value gets loaded into the latch before/after 150ms after motor on.

PR #872, opened 2020-11-18
Split Video.cpp to DirectVideo.cpp to remove most of the Win32 specif…

audetto:
…ic functions.

Video.cpp is now "pure" emulator (with the exception of a couple of MessageBox calls).


audetto:
Moved the Windows only files (Direct...) to Windows folder.
For a clear split they should not be included in the core emulator code, but they still are in a couple of cases.

Probably too soon to fix this, at least until the other 2 big files (AppleWin and Frame) are split.

tomcw:
I'm wondering about the name "DirectVideo.cpp" (and .h), because in DirectX there isn't a DirectVideo (but there is a DirectInput, so DirectInput.cpp/h makes sense for now). 

This "DirectVideo.cpp" file doesn't even use DirectX... instead it uses Windows APIs for the frame buffer (aka DC or Device Context), and GDI just for `StretchBlt()`... and the built-in debugger still uses `BitBlt()`.

NB. There is an init call to DirectDraw, but the function that uses this is commented out: `WaitForVerticalBlank()` - it's commented out as it was experimental, and unfortunately is a blocking call!

I think rename these to "WinVideo.cpp" and "WinVideo.h".

tomcw:
Thanks @audetto - all merged to master now.

Issue #827, opened 2020-08-31
NSC is recognised on Apple II/II+ when slot-3 is empty


tomcw:
The NSC is a 28-pin chip, but on the Apple II/II+ the 6 ROM chips are all 24-pin (eg. 6x 2K = 12K ROM). See Sather's UTAII, Ch.6 (page 6-1)... so this means that NSC can only be supported via a peripheral card with a 28-pin/4K ROM (but does such a card exist? since $C800 space is only 2K).

Repro:
- Select Model: Apple II or II+
- Config->Ethernet settings... and set Ethernet=Disabled
- in VS, set a breakpoint after a successful call to `IsPotentialNoSlotClockAccess()`
- drop to the Monitor, then read C300
- the BP will trigger

So the fix is to disallow NSC for II/II+ models.

NB. the BP triggers because in `IsPotentialNoSlotClockAccess()` this is true:
```
	((SW_INTCXROM || !SW_SLOTC3ROM) && (AddrHi == 0xC3))
```
But these soft-switches don't exist on the II/II+.

tomcw:
John Brooks pointed out that:
>NSC is available for 24-pin or 28-pin. See: https://manilagear.com/product/no-slot-clock/
>The Dallas 1216 clock only uses 2 pins, A0 and A2, so doesn't really care about the size of the ROM (or dram for DClock) that it is under.

From the Manila Gear manual, the NSC (24-pin config) does work on the II/II+, eg. when piggybacking the SSC's 24-pin ROM.

Anyway, the bug is still valid. To support NSC for the II/II+, we'd need:
- to support the NSC via the external SSC ROM (this would also work with the //e)
- or a ProDOS driver that eg. supported internal F8 ROM (for II/II+ only)

tomcw:
Info:
- No-Slot Clock Utilities v1.2 (C)1990 initially writes to $C300 just before reading the NSC's time (or setting the time). See $2A23 and $2AA2 respectively, eg:
```
2A23: STA $C300
      LDA $C304
```

- ProDOS8 2.5.0a8 doesn't write to $C300 (for NoSlotClk.IIe). Just before reading the NSC's time it only reads $C304 (I've no idea about setting the time), eg:
```
D74B: BIT $C304
```

AIUI, these initial accesses to $C300/C304 are to reset the NSC's comparitor. Presumably the P8 2.5.0a8 team think there's no need for the initial write to $C300.

Same for NoSlotClk.II (there's no write to $F8xx):
[...]

PR #873, opened 2020-11-22
Support NSC for Apple II/II+ via F8-ROM

tomcw:
PR for #827.

### DIFF
--- a/source/CPU.cpp
+++ b/source/CPU.cpp
@@ -442,8 +442,8 @@ static __forceinline void IRQ(ULONG& uExecutedCycles, BOOL& flagc, BOOL& flagn, 
 
 //===========================================================================
 
-#define READ _READ_NMOS
-#define WRITE(value) _WRITE_NMOS(value)
+#define READ _READ_WITH_IO_F8xx
+#define WRITE(value) _WRITE_WITH_IO_F8xx(value)
 #define HEATMAP_X(address)
 
 #include "CPU/cpu6502.h"  // MOS 6502
@@ -453,8 +453,8 @@ static __forceinline void IRQ(ULONG& uExecutedCycles, BOOL& flagc, BOOL& flagn, 
 
 //-------
 
-#define READ _READ_CMOS
-#define WRITE(value) _WRITE_CMOS(value)
+#define READ _READ
+#define WRITE(value) _WRITE(value)
 
 #include "CPU/cpu65C02.h" // WDC 65C02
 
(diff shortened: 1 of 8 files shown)


tomcw:
I have changed the 6502 (but not the 65C02) emulation so that r/w accesses to $F8xx now supports I/O too, specifically for the NSC.

So 6502 based Apple II's (ie. II, II+, II J-Plus) which all have an F8 Rom (but obviously not a Rom at $C300 / "CD" Rom) can support the NSC.

The 6502 based _unenhanced_ //e will use this 6502 emulation code path too, but since it's a //e it won't support the NSC at F8... and for a //e you'd expect the NSC to be at C3 (ie. the "CD" Rom).

NB. If you were to put a 65C02 into an Apple II/II+/II J-Plus then the NSC at F8 would no longer work, since the I/O support at $F8xx is only added for the 6502 emulation. (Currently the only way to do this is to manually edit the save-state file.)

Why not just support I/O at $F8xx for the 65C02 too? Well it's slightly more code, which could affect performance; also there's no use-case for this (unless you manually edit the save-state file for Apple II's!).

Other performance options are to only support I/O for reads (and not writes) to $F8xx, since currently this is all ProDOS 2.5.0a8 does. But since ProDOS 2.5.0a8 doesn't fully work with the NSC driver, then I consider this a "moving target" so it's safer to start with full support, then consider optimisations once ProDOS 2.5.0's NSC is stable.

tomcw:
Hi Nick / @sicklittlemonkey - FYI, I've created this PR to support NSC for the older (6502-based) Apple II models.

In #827, I've fixed AppleWin to not support the NSC at $C3xx!
So this PR complements that fix, so allow AppleWin to support the NSC at $F8xx. (As supported by ProDOS 2.5.0)

No need to review this PR, but I just wanted to give you a heads-up, since the NSC code is yours.

sicklittlemonkey:
Yeah, when I wrote the code I just chose somewhere that the drivers picked it up IIRC. I see you've logged issues against the ProDOS 2.5 testing repo. Nice one - thanks for all of this.

tomcw:
**UPDATE:** I fixed a bug (26414b7) where LC mode (read ROM / write RAM) wasn't working for II/II+ models with NSC at $F8-ROM.

NB. this bug doesn't exist in any release of AppleWin, as "NSC F8-ROM" support will be added at 1.30.0.0, and this bug was identified and fixed prior to release.

The details are it was flagged by an a2audit test:
- lda c080 ; reset things
- lda c081
- lda c089 ; ROM read, bank 1 write
- inc fe1f ; read $60 from ROM, write $61 to RAM

So in addition to not doing NSC for `HIGHRAM` (LC: r/w RAM), I also exclude `WRITERAM` (LC: r:RAM/w:ROM).

PR #875, opened 2020-11-24
Applewin split

audetto:
Split AppleWin.cpp in 3

1. Core.cpp: code needed by the emulator itself (plus strictly related functions / variables)
2. Utilities.cpp: more generic code that can be used by different archs / ports
3) AppleWin.cpp: WinMain() of the current AW (with execution loop)


Michaelangel007:
@tomcw Reviewing commits ... will update as I go ...

* [x] 00e58004 Looks good
* [ ] b0ead60
* [ ] 7dcd230
* [ ] 03135bf

audetto:
Re UseClockMultiplier: I have not seen any in the current repo. I can make up something if you want.

audetto:
Sorry, made a mess of a simple comment.
Now it reads right.

tomcw:
Hi @audetto 

1. In source/Core.h, there's this line:
```
extern __interface IPropertySheet& sg_PropertySheet;
```
Does GCC (or Clang) properly support this Microsoft-specific language extension?

2. In source/Windows/AppleWin.cpp, the entire `ProcessCmdLine()` has remained in AppleWin.cpp (which is now Windows-specific). But I think that most of this function is platform-agnostic code.
Maybe put it in Source/Core.cpp? (with the Windows-specific bits wrapped in #ifdef _MSC_VER)



NB. In source/Utilities.cpp, there are functions dealing with Disk and HardDisk, eg. `DoDiskInsert()` and `DoHardDiskInsert()`, I should probably move these to Disk.cpp and Harddisk.cpp. No need to take any action for now.

audetto:
1) No, but it is easy enough to add
``#define __interface struct``
and get rid of it. The whole property sheet is something I have ignored so far as it is too much Windows code.
This is all is **really** needed to run the CPU:
https://github.com/audetto/AppleWin/blob/master/source/linux/duplicates/IPropertySheet.cpp
I am not familiar with ``__interface``, could it be replaced by a C++ class with virtual functions?

2) I did not put it in Core.cpp because I left Core.cpp for the code needed by the emulator when it runs (basically ``CpuExecute()``)
Since it is a big chunk of uniform code, I put it in a standalone file CmdLine.cpp.

audetto:
There is more common code that should go to ``source``: the initialisation and shutdown.

The problem is that it is currently split some in ``WinFrame`` some in ``AppleWin`` and it mixes initialisation of the emulator (which is really common) and initialisation of the host system (which is not).
So I have not done anything about it. But could be next.

tomcw:
re. __interface - see https://docs.microsoft.com/en-us/cpp/cpp/interface?view=msvc-160
Basically a class with only pure virtual functions, but you don't need to prefix the functions with `virtual` or suffix with `= 0`.
So it'll be easy to convert these back to standard C++.

>Since it is a big chunk of uniform code, I put it in a standalone file CmdLine.cpp.

Yes, moving `ParseCmdLine()` to a dedicated .cpp looks fine.

>There is more common code that should go to `source`: initialisation, different restarts and shutdown.

Yeah, eg. in shutdown (via the WM_CLOSE & WM_DESTROY messages in WinFrame.cpp)... I guess just replace/move all this to a core shutdown function.

PR #878, opened 2020-11-28
Select a Alpha value that works in Windows and Linux (qt / sdl).

audetto:
In Windows it does not seem to make any difference, so 0xFF is selected for maximum compatibility.

Alternatively, I could add an #ifdef _MSC_VER, but then I am not clear why NTSC.cpp and RGBMonitor.cpp set the alpha to different values (0xFF and 0x00 respectively).

See: https://github.com/AppleWin/AppleWin/blob/master/source/NTSC.cpp#L176 and related usage.

Ideally all the 3 #define should go to a common .h (Video.h?) to avoid they diverge.

Will improve the patch if I get feedback on the above.


tomcw:
Yes, we should be consistent for the alpha value.

As you suggest, can you move the alpha-related #defines / const's into Video.h.

PR #881, opened 2020-11-29
Cleanup types and #include to ease gcc compilation.

audetto:
Some small commits to ease compilation with gcc.

Most of it is around duplicate definitions or #include of types from ``windows.h``.
In Linux a windows.h compatibility layer is required and it becomes harder if different files redefine some of it independently.


tomcw:
Thanks (merged now).
btw. I also removed alarm.h from the project, since it's not used at all now (e2d7eb7)

PR #884, opened 2020-12-05
Remove VirtualAlloc & Co.

audetto:
Replaced by malloc, free & memset.

The only tricky bit has been the fact that MEM_COMMIT implies zero-initialisation (which I had completely forgotten).

This was apparently only ever required in TestCPU6502.cpp (replaced by calloc), while other parts of the code were ignoring this feature and calling ZeroMemory just after a MEM_COMMIT.

### DIFF
--- a/source/Configuration/PropertySheetHelper.cpp
+++ b/source/Configuration/PropertySheetHelper.cpp
@@ -158,7 +158,7 @@ std::string CPropertySheetHelper::BrowseToFile(HWND hWindow, TCHAR* pszTitle, TC
 	std::string pathname = szFilename;
 
 	OPENFILENAME ofn;
-	ZeroMemory(&ofn,sizeof(OPENFILENAME));
+	memset(&ofn, 0, sizeof(OPENFILENAME));
 
 	ofn.lStructSize     = sizeof(OPENFILENAME);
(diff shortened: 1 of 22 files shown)


tomcw:
btw. the VirtualAlloc()'s for the 64K regions (aligned to 0x10000) may have had some benefit in the (very) old x86 assembly days (for the 6502 emulation). The only benefit these days is when debugging: as once you have the base address in the Memory window, then you can easily go to any 64K address by just changing the low 0x????nnnn address.

>Replaced by malloc, free

Actually can you use new and delete [] instead.
The rest looks fine.

audetto:
Good point.
I kept VirtualAlloc (under #ifdef just for Memory.cpp).

tomcw:
I would say you only need to keep VirtualAlloc() for the allocations of size = _6502_MEM_LEN.
For all the other sizes, just use new/delete[].

For the GCC versions of ALIGNED_ALLOC(size) and ALIGNED_FREE(ptr) then can you also define these as new & delete too? EG
```
#define ALIGNED_ALLOC(size) new BYTE [size]
#define ALIGNED_FREE(ptr) delete [] ptr
```

tomcw:
Thanks - merged this into master.

PR #885, opened 2020-12-06
Fix memory leak in CardManager

audetto:
And remove one-off NULL check (slots are never NULL).


tomcw:
Thanks - merged this in.

Issue #380, opened 2017-01-22
"Unable to capture PrintScreen key" warning on startup


kdechant:
A warning box that reads "Unable to capture PrintScreen key" appears on AppleWin startup. See attached screenshot. This appears every time I start AppleWin but doesn't seem to have any other adverse effects.

My PC is running Windows 10 on the Insider Slow ring. I noticed this error right after I upgraded to build 14986. There is possibly some incompatibility with a new feature of Windows.

This occurred in both AppleWin 1.24.x and 1.26.0.6.

![unable to capture printscreen](https://cloud.githubusercontent.com/assets/727401/22180176/b1bc2d4c-e01d-11e6-805e-6dc9a407d6ec.PNG)

mediaguy:
Only time I had this issue was when another instance of AppleWin was running already ... but that's just one possibility.

sicklittlemonkey:
This will happen when any other software has already registered to capture
the Print Screen key.

Check if you have Snagit installed etc.

Applewin should probably have a checkbox option to not show this warning
again, since it is a pretty trivial issue for most.

Cheers,
Nick.

sicklittlemonkey:
From Marco Verpelli in comp.emulators.apple2:
"OneDrive for Win10 "steal" by default the PrintScreen key, then after a (re)installation of Windows, hit that key and simply say no to the application."

Seems like this is a common annoyance from OneDrive:
https://answers.microsoft.com/en-us/onedrive/forum/odoptions-oddesktop/onedrive-has-hijacked-my-print-screen-key/43c40cdb-38ee-4719-afb6-39ae2da71cc9?auth=1

Michaelangel007:
Feature Requests:

* [ ] Dialog option to disable warning -- we already have a command line: `-no-printscreen-dlg`
* [ ] Add normal function key as a fallback
* [ ] Add custom hotkeys

MICROMATICA1:
Hi
Close or stop any program that uses or accesses the 'PrintScreen' key.
Example: Lightshot

Close one by one, the programs that load at startup and are concurrent with AppleWin.
Will solve
I am available to be a beta tester.
Milton

sicklittlemonkey:
OneDrive is probably the most common culprit these days.

Cheers,
Nick.

kdechant:
I don't use the PrintScreen key in AppleWin and I don't care if it works. I just want to get rid of the dialog.

[...]

PR #891, opened 2020-12-18
Fix for #380 ("Unable to capture PrintScreen key" warning on startup)

kevinmarty:
This is a fix for issue #380 that happens when some other program has already registered the PrintScreen key as a hotkey (OneDrive or the Windows 10 Snip & Sketch tool are common culprits).

I changed the error message in the RegisterHotKeys() function from using the MessageBox() API to using the Windows shlwapi function SHMessageBoxCheck() API, which adds a "Don't show this message again" checkbox to the message.

The GUID in the last parameter is randomly-generated, it is used by the Windows API to store whether the box has been checked in the registry.


tomcw:
This looks fine to me.

NB. This SHMessageBoxCheck() API is supported on Windows XP and above, so by merging this in, then we need to implicitly state that Win XP is the oldest theoretically "supported" version for AppleWin. Bear in mind that I'm not testing against Win XP, hence why I'm saying "theoretically supported".

kevinmarty:
I changed the code to attempt to dynamically load the SHMessageBoxCheck API and to fall back on using a standard MessageBox if that fails, so this will take care of running on pre-Win XP versions as well.

tomcw:
I've made the change to use v141_xp toolset for AppleWin releases (8d7d2b9647f0fa139444440c924d578f782693a5), which means that Windows XP is the oldest "supported" version. So I will take your initial commit that statically links against shlwapi.lib rather than the newer DLL version.

Can you revert? (perhaps just resubmit your initial change again?) I'll do a 'squash & merge' so it won't matter about these unused changes.

kevinmarty:
Sure, I just removed the commit that added the dynamic linking.

tomcw:
Thanks @kevinmarty - I've merged this to master now.

tomcw:
NB. **SHMessageBoxCheck** records the message boxes that the user has chosen to suppress under the following registry key:
HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Explorer\DontShowMeThisDialogAgain

sicklittlemonkey:
Interesting, I'd never seen this API.

PR #892, opened 2020-12-19
Property sheet

audetto:
Make property sheet into a standard c++ class 
and 
move it to Interface.h as it is mostly arch-specific code.


source/Windows/WinFrame.h
```diff
@@ -20,14 +24,14 @@
 	void    FrameCreateWindow(void);
 	HDC     FrameGetDC ();
 	void    FrameReleaseDC ();
-	void    FrameRefreshStatus (int, bool bUpdateDiskStatus = true );
```
tomcw, line 23: Did you mean to remove this default parameter value?

And to the 2 functions below too:
- SetViewportScale(int nNewScale, bool bForce = false);
- GetBestDisplayResolutionForFullScreen(UINT& bestWidth, UINT& bestHeight, UINT userSpecifiedHeight=0);

I think these default values are still needed.

audetto: Yes.
The real default value is in Interface.h (and this is why I have included).

The code in WinFrame.h / .cpp is just the implementation.
Probably a better solution is to remove these functions from WinFrame.h and only leave them in Interface.h.

audetto:
See if this is more clear.

Now there is only one version of the declaration for each function and the default arguments go there.

tomcw:
Yes, in general I think this is OK for now.

Perhaps a bit odd with WinVideo.cpp prototypes (declarations) now in Interface.h, but not in WinVideo.h.

If (or when) we come to create a base Video.cpp/h and a sub-class LinuxVideo.cpp/h, then then related prototypes will move to Video.h. So similar to what I said in my email on 23/11/2020:
> I was thinking of having a platform-agnostic base class (Frame), then subclasses for Windows (WinFrame) and Linux (LinuxFrame).

Anyway, I will merge this in...

PR #893, opened 2020-12-20
Move interface into a class FrameBase.

audetto:
Move the Win32 implementation into ``Win32Frame``.

The logic behind all of this is
- if it is called by core emulator code (i.e. portable .cpp in ``source),`` then it **must** go to ``FrameBase`` (do not ``#include "Windows/..."`` there)
- otherwise it can go either to a global variable or as a member variable of ``Win32Frame`` (**not** in ``FrameBase``, as it would force it to all other archs)
- should the main event loop be moved into ``Win32Frame``? Yes and No. I would vote for No at the moment. Put in ``Win32Frame`` some more closely related functions, but draw the line somewhere.
- why are ``g_hInstance`` and ``g_hFrameWindow`` in? They do not belong there, but if I remove them, then I need to add many more functions to ``FrameBase`` (``MessageBox``, ``LoadBitmap``... too much work for today)
- why files called ``WinFrame.cpp`` and ``Win32Frame.cpp``? I run out of imagination
- why ``Win32Frame.cpp`` is empty? Because it would take a lot of work at this stage to move all necessary related functions from ``AppleWin.cpp``, ``WinFrame.cpp`` and ``WinVideo.cpp`` there.
- this PR makes the code in ``source`` cleaner, but makes a mess of ``Windows``? I disagree, but I see your point. This just another step, more are needed.

This PR does not attempt to move relevant global variables and functions from ``AppleWin.cpp``, ``WinFrame.cpp`` and ``WinVideo.cpp`` to ``Win32Frame``.

### DIFF
--- a/AppleWinExpress2019.vcxproj
+++ b/AppleWinExpress2019.vcxproj
@@ -78,6 +78,7 @@
     <ClInclude Include="source\DiskImageHelper.h" />
     <ClInclude Include="source\DiskLog.h" />
     <ClInclude Include="source\Frame.h" />
+    <ClInclude Include="source\FrameBase.h" />
     <ClInclude Include="source\Harddisk.h" />
     <ClInclude Include="source\Interface.h" />
     <ClInclude Include="source\Joystick.h" />
@@ -120,6 +121,7 @@
     <ClInclude Include="source\Video.h" />
     <ClInclude Include="source\Windows\AppleWin.h" />
     <ClInclude Include="source\Windows\DirectInput.h" />
+    <ClInclude Include="source\Windows\Win32Frame.h" />
     <ClInclude Include="source\Windows\WinFrame.h" />
     <ClInclude Include="source\Windows\WinVideo.h" />
     <ClInclude Include="source\YamlHelper.h" />
@@ -155,6 +157,7 @@
     <ClCompile Include="source\Core.cpp" />
     <ClCompile Include="source\CPU.cpp" />
     <ClCompile Include="source\Disk2CardManager.cpp" />
+    <ClCompile Include="source\FrameBase.cpp" />
     <ClCompile Include="source\RGBMonitor.cpp" />
     <ClCompile Include="source\SAM.cpp" />
     <ClCompile Include="source\Debugger\Debug.cpp" />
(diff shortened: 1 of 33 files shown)


tomcw:
There is a class relationship between FrameBase (FrameBase.cpp/h) and Win32Frame (Win32Frame.cpp/h), ie. Win32Frame IS_A FrameBase (ie. Win32Frame is a subclass of FrameBase).

We also have Frame.cpp/h and WinFrame.cpp/h where Frame.cpp was the original file, and the Windows-specific functions have been moved to WinFrame.cpp. These are just cpp files without classes: so just free-standing functions/variables at a global or static scope.

In time, can we aim to reduce all this into just 2 classes...?
- Frame.cpp which defines `class Frame`
- WinFrame.cpp which defines `class WinFrame`  (where WinFrame IS_A Frame).

Ultimately I don't want Frame/FrameBase/WinFrame/Win32Frame, but just want this reduced to two.
So I'm OK with this PR, if it's just a stepping-stone on the way to getting to just two.

audetto:
Indeed,

I think we can easily free https://github.com/AppleWin/AppleWin/blob/master/source/Frame.h because it currently contains a few functions strictly linked to ``g_pFramebufferbits`` at https://github.com/AppleWin/AppleWin/blob/master/source/Video.h#L184

The sizes are really tightly linked to the array as they allow to interpret the video memory buffer, so in my opinion they really belong to the emulator core (Video.h, maybe the Video class at https://github.com/AppleWin/AppleWin/pull/812?) rather than the arch dependent frame.

And then clean up a bit Win32Frame.cpp and WinFrame.cpp to move some free standing functions inside the class (private to the WinFrame class file).

tomcw:
Yes, I agree about functions related to `g_pFramebufferbits`, and moving them (out of Frame) into Video class sounds fine to me.

For me, Frame.cpp (now WinFrame.cpp) has mainly been for `FrameWndProc()`, ie. the big switch statement to handle Windows messages, and functions to draw the frame (ie. the Window, windows title, buttons, cross-hairs, etc).

re. the Video class at #812 - maybe the sooner we do this better? I did hope to find some time to do this, but so far I've had other priorities. Maybe next week...

Anyway, I will merge this PR now.

PR #901, opened 2020-12-29
WinVideo: share some generic code

audetto:
Move generic functions to the base class as they can be reused by other archs.

Split VideoRefreshScreen() to extract the system-specific code to VideoPresentScreen().


tomcw:
That's good to see more generic code being moved back to the base class.

btw, `TCHAR` (and the `_tcsrchr()`, `_tcsncpy()` etc functions) can all be converted to `char` and their equivalent `string.h` functions.

PR #908, opened 2021-01-01
Video & FrameBase: better split

audetto:
While implementing ``Video`` and ``FrameBase`` for the linux frontends I realised the split was not very clear.

With this PR the ``Video`` class is very close to be only the Apple ][ video device, while host management is deferred to ``FrameBase``:

- no more virtual functions in ``Video``
- ``FrameBase`` uses ``Video``, but **not** the other way round
- Win32 code interacts mostly with ``FrameBase``, emulator with ``Video``


tomcw:
@audetto - there are two `g_pFramebufferbits`:
- Video::g_pFramebufferbits
- Win32Frame::g_pFramebufferbits

One of them should be removed.

tomcw:
In `void Win32Frame::Destroy(void)` when `DeleteObject(g_hDeviceBitmap)` is called then this unmaps (deallocates) the memory pointed to by `Video::g_pFramebufferbits`.

So I think immediately after this `DeleteObject(g_hDeviceBitmap)` line, you need to `SetFrameBuffer(NULL)`... but this is a private function, so perhaps just move the `GetVideo().Destroy();` line to be after the `DeleteObject(g_hDeviceBitmap)` line, eg...

```
	DeleteObject(g_hDeviceBitmap);
	GetVideo().Destroy();
```

tomcw:
OK, merged this into master.

Issue #611, opened 2019-01-06
Holding down F9 to cycle through the video modes results in occasional flicker with garbage


tomcw:
Spun off from #357, where Nick said:
>One thing I happened to notice though - holding down F9 cycles through the modes, but occasionally a line will flicker with garbage - for just one frame I guess.

tomcw:
Probably related to this issue...

From the start-up logo: set video mode = Color (NTSC Monitor), and then 'Benchmark Emulator'. On switching to HGR, the top line will appear shifted right by a 7M pixel.

The shift occurs for this initial/top line, because in `updateVideoScannerAddress()`, this condition is false:
```
g_eVideoType == VT_COLOR_MONITOR_NTSC && GetColorBurst()
```

ie. `GetColorBurst()` is false, and so the line offset isn't adjusted by -2.

NB. If you do the benchmark after starting normal emulation, then it won't be the top line that's shifted, but the current video scanner's vpos.

tomcw:
NB. CTRL+F9 (which currently does nothing) will also cause some flicker.
In `FrameWndProc()`, this seeming no-op still calls:
- `VideoReinitialize()`
  - calls: `updateVideoScannerAddress()` which resets `g_pVideoAddress` to start of current line
- `VideoRefreshScreen()`

I tried a quick hack so that `updateVideoScannerAddress()` is not called during an F9+modified combo, and this seems to fix this glitching.

sicklittlemonkey:
Looks good, thanks!

I mean, of course it wasn't when holding down F9 that it bothered me.
Just occasionally when I hit F9 to change the colour mode sometimes I would
catch a bad frame out of the corner of me eye. ; - )
Cheers,
Nick.

PR #910, opened 2021-01-05
FrameBase: some changes to ease implementations

audetto:
This PR contains a few things (happy to split if necessary and to improve on feedback)

- simplify the interface FrameBase to make it easier to implement it for different cases (remove HDC and make parameters more explicit)
- remove functions which are only called on a Win32Frame (in which case a cast is guaranteed to succeed)
- otherwise there is the risk that every FrameBase implementation wants to add its own variants.
- video.VideoReinitialize(false); when changing the MonoChrome color: should the video address be reinitialised? I looked at the calls made for other video changes (F9, Shift-F9...) and for consistency changed it. But it might a mistake of mine.

### DIFF
--- a/source/Windows/Win32Frame.cpp
+++ b/source/Windows/Win32Frame.cpp
@@ -499,3 +499,10 @@ void Win32Frame::DDUninit(void)
 }
 
 #undef SAFE_RELEASE
+
+Win32Frame& Win32Frame::GetWin32Frame()
+{
+	FrameBase& frameBase = GetFrame();
+	Win32Frame& win32Frame = static_cast<Win32Frame&>(frameBase);
+	return win32Frame;
+}
--- a/source/Windows/Win32Frame.cpp
+++ b/source/Windows/Win32Frame.cpp
@@ -517,6 +517,6 @@ void Win32Frame::DDUninit(void)
 Win32Frame& Win32Frame::GetWin32Frame()
 {
 	FrameBase& frameBase = GetFrame();
-	Win32Frame& win32Frame = static_cast<Win32Frame&>(frameBase);
+	Win32Frame& win32Frame = dynamic_cast<Win32Frame&>(frameBase);
 	return win32Frame;
 }
--- a/source/Windows/Win32Frame.cpp
+++ b/source/Windows/Win32Frame.cpp
@@ -22,6 +22,20 @@ Win32Frame::Win32Frame()
 	g_hLogoBitmap = (HBITMAP)0;
 	g_hDeviceBitmap = (HBITMAP)0;
 	g_hDeviceDC = (HDC)0;
+	g_bAltEnter_ToggleFullScreen = false;
+	g_bIsFullScreen = false;
+	g_bShowingCursor = true;
+	g_bLastCursorInAppleViewport = false;
+	g_uCount100msec = 0;
+	g_TimerIDEvent_100msec = 0;
+	g_bUsingCursor = FALSE;
+	g_bAppActive = false;
+	g_bFrameActive = false;
+	g_windowMinimized = false;
+	g_bFullScreen_ShowSubunitStatus = true;
+	g_win_fullscreen_scale = 1;
+	g_win_fullscreen_offsetx = 0;
+	g_win_fullscreen_offsety = 0;
 }
 
 void Win32Frame::videoCreateDIBSection(Video & video)
--- a/source/Configuration/PageConfig.cpp
+++ b/source/Configuration/PageConfig.cpp
@@ -28,6 +28,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 #include "../Windows/AppleWin.h"
 #include "../Windows/WinFrame.h"
+#include "../Windows/Win32Frame.h"
 #include "../Registry.h"
 #include "../SerialComms.h"
 #include "../resource/resource.h"
@@ -115,7 +116,7 @@ INT_PTR CPageConfig::DlgProcInternal(HWND hWnd, UINT message, WPARAM wparam, LPA
 			break;
 
 		case IDC_MONOCOLOR:
-			GetFrame().ChooseMonochromeColor();
+			Win32Frame::GetWin32Frame().ChooseMonochromeColor();
 			break;
 
 		case IDC_CHECK_CONFIRM_REBOOT:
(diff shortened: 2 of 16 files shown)


audetto:
I realise this is a bit confused. Give me one more day to propose something more meaningful.

audetto:
There are a few more variables to move into ``Win32Frame``. I think the idea looks good.

And there are more things to remove: https://github.com/AppleWin/AppleWin/blob/master/source/FrameBase.h#L12-L13
It would require adding functions like ``MessageBox`` to ``FrameBase``. It is a bit of a boring job, but I don't see any hurdle on the way.

This PR is ready, but if people prefer I complete the job first, I can push more changes and review at the end.

source/Windows/Win32Frame.cpp
```diff
@@ -499,3 +513,10 @@ void Win32Frame::DDUninit(void)
 }
 
 #undef SAFE_RELEASE
+
+Win32Frame& Win32Frame::GetWin32Frame()
+{
+	FrameBase& frameBase = GetFrame();
+	Win32Frame& win32Frame = static_cast<Win32Frame&>(frameBase);
```
tomcw, line 520: Since you are casting up/down the class hierarchy, then this should be a dynamic_cast<>.

tomcw:
>video.VideoReinitialize(false); when changing the MonoChrome color: should the video address be reinitialised? I looked at the calls made for other video changes (F9, Shift-F9...) and for consistency changed it. But it might a mistake of mine.

I'm looking for the code you are talking about...

The monochrome colour is changed in here: `void Win32Frame::ChooseMonochromeColor(void)`
And I can see this function calls `video.VideoReinitialize();`  (... and the 1st arg defaults to true)

And when video mode changes (F9/Shift+F9), then this called `GetVideo().VideoReinitialize(false);`

VideoReinitialize() calls `NTSC_SetVideoStyle()`
And `NTSC_SetVideoStyle()` sets up the mono colour: see the switch case `VT_MONO_CUSTOM`.

But your question is:
>should the video address be reinitialised?

By this do you mean `Video::g_pFramebufferbits` ?
No - this never needs reinitialising (unless you "restart" the Apple II virtual machine).

tomcw:
btw, there are lots of cases where you changed the code from:
```
	GetFrame().FrameRefreshStatus(DRAW_TITLE);
```
to
```
	GetFrame().FrameRefreshStatus(DRAW_TITLE | DRAW_DISK_STATUS);
```

Was this a bug before?

audetto:
No no bug, but before the interface was (IMHO) more complicated:

2 arguments: ``int drawflags, bool bUpdateDiskStatus``
and this logic
``drawflags |= bUpdateDiskStatus ? DRAW_DISK_STATUS : 0;``
in ``WinFrame.cpp``

So I just made it more explicit and passed ``DRAW_DISK_STATUS`` where needed, and another ``FrameBase`` implementation just redraws according to the flags.

audetto:
> I'm looking for the code you are talking about...
> 
> The monochrome colour is changed in here: `void Win32Frame::ChooseMonochromeColor(void)`
> And I can see this function calls `video.VideoReinitialize();` (... and the 1st arg defaults to true)

Yes.
So my question is
why argument = ``true`` here: https://github.com/AppleWin/AppleWin/blob/master/source/Windows/Win32Frame.cpp#L295
and explicitly ``false`` here: https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1248
?

None of them requires a full video reinitialisation like
https://github.com/AppleWin/AppleWin/blob/master/source/Configuration/PageConfig.cpp#L317
https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1137
where I understand the ``true``.

audetto:
Moreover, I imagine that changing the VideoMode or the Monochrome color should be handled very similarly.

Should these 2 sections be the same?
https://github.com/AppleWin/AppleWin/blob/master/source/Windows/Win32Frame.cpp#L295-L300
https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1248-L1266

But I might be a bit too naive.

tomcw:
>So my question is
>why argument = true here: https://github.com/AppleWin/AppleWin/blob/master/source/Windows/Win32Frame.cpp#L295
>and explicitly false here: https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1248
>?

The only time `VideoReinitialize(false)` is called, ie. passing `false` is when changing mode with F9/Shift+F9.

For the F9/Shift+F9 case, if you remove this `false` (so that it defaults to passing `true`), then _holding down F9_ will result in some glitching in the rendering. EG. drop to the monitor (`CALL-151`) and do `E000L` to get lots of text on screen, then hold down F9.

In fact, see this bug and comment when I made this specific fix: https://github.com/AppleWin/AppleWin/issues/611#issuecomment-473525332

So for other cases (eg. change monochrome colour) you'll get this glitch too, but it'll just be a one-off glitch, since you can't do it repeatedly quickly enough.

Basically if you are changing some video characteristic mid-scanline don't call `updateVideoScannerAddress()`... only call it at the start of a scanline (ie. when `g_nVideoClockHorz == 0`). This means pass `false` when calling `VideoReinitialize()`.

And so there's arguably a subtle bug in `NTSC_VideoReinitialize()`, where maybe `g_nVideoClockHorz != 0` and then there's a call to `updateVideoScannerAddress()`. NB. this glitch is harmless and corrects itself at the end of the scanline.

Anyway, as an extension to your question then the following calls to `VideoReinitialize()` should explicitly pass `false`:
-  https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1248... already passes `false`.
- `ChooseMonochromeColor(void)`: https://github.com/AppleWin/AppleWin/blob/master/source/Windows/Win32Frame.cpp#L295
- `CPageConfig::DlgOK(HWND hWnd)`... this is the same as F9/Shift-F9, just from the config GUI.
- WndProc() for `case WM_DISPLAYCHANGE:`

tomcw:
>Moreover, I imagine that changing the VideoMode or the Monochrome color should be handled very similarly.
>Should these 2 sections be the same?

Yes, they should be the same... as should the relevent code in `CPageConfig::DlgOK(HWND hWnd)`.

tomcw:
Merged to master.

btw, do you plan to merge WinFrame.cpp and Win32Frame.cpp into a single .cpp file?
ie. put everything into class Win32Frame.

audetto:
Yes, with the global sin WinFrame almost gone, there is no point in keeping it separate.

PR #911, opened 2021-01-10
Videoreinitialize: make code more uniform

audetto:
This was originally discussed in https://github.com/AppleWin/AppleWin/pull/910#issuecomment-757303832

When VideoMode changes, call ``VideoReinitialize(false)``.

I still have one open question.
If this code is needed after a ``F9``: https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1254-L1258

should it be needed here as well:
https://github.com/AppleWin/AppleWin/blob/master/source/Windows/Win32Frame.cpp#L296
and here
https://github.com/AppleWin/AppleWin/blob/master/source/Configuration/PageConfig.cpp#L318
?
I might be trying too hard to make uniform parts of the code which are not meant to, they just look very similar.

### DIFF
--- a/source/Configuration/PageConfig.cpp
+++ b/source/Configuration/PageConfig.cpp
@@ -312,15 +312,8 @@ void CPageConfig::DlgOK(HWND hWnd)
 
 	if (bVideoReinit)
 	{
-		GetVideo().Config_Save_Video();
-
 		win32Frame.FrameRefreshStatus(DRAW_TITLE);
-
-		GetVideo().VideoReinitialize(false);
-		if ((g_nAppMode != MODE_LOGO) && (g_nAppMode != MODE_DEBUG))
-		{
-			win32Frame.VideoRedrawScreen();
-		}
+		win32Frame.ApplyVideoModeChange();
 	}
 
 	//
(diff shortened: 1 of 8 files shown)


tomcw:
Great - getting rid of that default argument is much better.

>I still have one open question.
>If this code is needed after a `F9`:

Sorry, the code has moved after the previous merge! But I think it's this section of code:
https://github.com/AppleWin/AppleWin/blob/master/source/Windows/WinFrame.cpp#L1227-L1245

```
			GetVideo().VideoReinitialize(false);

			if (g_nAppMode != MODE_LOGO)
			{
				if (g_nAppMode == MODE_DEBUG)
				{
					UINT debugVideoMode;
					if ( DebugGetVideoMode(&debugVideoMode) )
						VideoRefreshScreen(debugVideoMode, true);
					else
						VideoPresentScreen();
				}
				else
				{
					VideoPresentScreen();
				}
			}

			GetVideo().Config_Save_Video();
```

And yes, that same code should be run for both `Win32Frame::ChooseMonochromeColor()` and `CPageConfig::DlgOK(HWND hWnd)`.  So perhaps refactor and add a new function to FrameBase that encapsulates this code.

tomcw:
Thanks - merged to master.

PR #915, opened 2021-01-17
Remove usages of g_hFrameWindow and g_hInstance from core emulator files. And Resource related Win32 functions.

audetto:
Add 3 virtual functions to ``FrameBase``:

- ``FrameMessageBox``
- ``FrameLoadBitmap``
- ``Restart``

After this, the only usages left of ``g_hFrameWindow`` and ``g_hInstance`` are in the file open dialogs for the disk selection (relative to the portable code only).
Will tackle these next and move them from ``FrameBase`` to ``Win32Frame``.


source/Windows/Win32Frame.cpp
```diff
@@ -575,3 +575,21 @@ Win32Frame& Win32Frame::GetWin32Frame()
...
+HBITMAP Win32Frame::FrameLoadBitmap(LPCSTR lpBitmapName)
+{
+	return LoadBitmap(g_hInstance, lpBitmapName);
+}
+
+void Win32Frame::Restart()
+{
+	// do we really need to PostMessage?
+	// could we just execute it here directly
+	PostMessage(g_hFrameWindow, WM_USER_RESTART, 0, 0);
```
audetto, line 594: Not an expert, but I think we can replace with
```
// Changed h/w config, eg. Apple computer type (][+ or //e), slot configuration, etc.
g_bRestart = true;
PostMessage(window,WM_CLOSE,0,0);
```
and axe ``WM_USER_RESTART`` altogether.

audetto:
I've added as well a function in FrameBase to hide the Resource framework and allow alternative implementations.

source/Windows/Win32Frame.cpp
tomcw: Yes, that's correct - you can axe `WM_USER_RESTART` and replace with:
```
g_bRestart = true;
PostMessage(window,WM_CLOSE,0,0);
```

source/Windows/Win32Frame.cpp
```diff
@@ -575,3 +575,42 @@ Win32Frame& Win32Frame::GetWin32Frame()
...
+
+	DWORD dwResSize = SizeofResource(NULL, hResInfo);
+	if (dwResSize != dwExpectedSize)
+		return NULL;
+
+	HGLOBAL hResData = LoadResource(NULL, hResInfo);
+	if (hResData == NULL)
+		return NULL;
+
+	BYTE* pVideoRom = (BYTE*)LockResource(hResData);	// NB. Don't need to unlock resource
```
tomcw, line 613: Minor thing: rename this `pVideoRom` variable to something more generic, eg. `resource`.

tomcw:
Thanks... merged in.

PR #919, opened 2021-01-20
SpkrSetEmulationType: simplify logic and remove unreachable code.

audetto:
The check (soundtype != newtype) is always false.

So we can remove dependency on a HWND argument.


tomcw:
Actually I'd like to keep that `MessageBox()` message, but ideally move it to `SpkrInitialize()`, and have it appear only if the call to `Spkr_DSInit()` fails.

tomcw:
Thanks. Merged in.

Issue #916, opened 2021-01-18
Pravets: fix or remove the printer card support for converting from Latin to Kir chars


tomcw:
Spun off from #912.

Basically the function `Pravets::ConvertToPrinterChar(BYTE value)` is broken after switching from ANSI to UTF-8 (initially broken 2 years ago from PR #604).

The problem is there's no test (unit or regression) for this, so any fix will need to be validated.

Perhaps better to just remove this feature?

audetto:
Question 1: is this a feature of the actual computer or the printer attached to it? did they send a font or just a byte to be converted by the printer firmware.
So, who decides that 65 is A? The computer sending a dot matrix or the printer's firmware?

Question 2: what does a real printer do for _non printable characters_?

audetto:
Something like this https://github.com/AppleWin/AppleWin/pull/917?

tomcw:
Thanks - that looks good.
I'll accept the PR, but leave this issue open, as it'd be good to add a regression test.

---

As for your questions:

>Question 1: is this a feature of the actual computer or the printer attached to it? did they send a font or just a byte to be converted by the printer firmware.
>So, who decides that 65 is A? The computer sending a dot matrix or the printer's firmware?

Not the Pravets computer... it'll be the printer interface card's firmware or the printer itself.

>Question 2: what does a real printer do for _non printable characters_?

Depends on the printer interface card and/or the printer.
EG. some will just accept 7-bit values, and _non printable characters_ will do special operations, eg. 0x0D/CR=carriage return, 0x0C/FF=form feed (ie. move to the next sheet of paper).
And some will also accept 8-bit values, and maybe the characters with bit7=1 will print special characters.

sicklittlemonkey:
Hi guys, great work tidying this up. It did make the core print routine a
bit messy.

Characters printed are determined by the printer. Most of the parallel or
serial interface cards only handle flow control and escaping etc. The
character sets are in the printer.

Cheers,
Nick.

audetto:
> It did make the core print routine a bit messy.

I agree. If there was a data type to represent a single code point, it would probably be better.
Could ``wchar_t`` help?

tomcw:
Perhaps you could programmatically convert the UTF-8 strings back to ANSI, eg. see the hoops I had to jump through to the other way (ANSI -> unicode -> UTF-8), here:

[...]

PR #920, opened 2021-01-21
Alternative way to handle UTF-8 in Pravets printer conversion

tomcw:
Just convert the UTF-8 strings to Ansi.

Also trimmed the Lat8A ASCII string by 1 char, so that it matches the 2 upper/lowercase Kir8A strings.

### DIFF
--- a/source/Pravets.cpp
+++ b/source/Pravets.cpp
@@ -34,6 +34,9 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #include "Keyboard.h"
 #include "Tape.h"
 
+const char Pravets::m_Lat8A[] = "abwgdevzijklmnoprstufhc~{}yx`q|";
+const char Pravets::m_Lat82[] = "abwgdevzijklmnoprstufhc^[]yx@q{}~`";
+
 Pravets::Pravets(void)
 {
 	// Pravets 8A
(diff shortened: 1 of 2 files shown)


audetto:
Using a single ``char`` makes the code simpler but will the result not depend on the codepage selected?

On the other hand a given PC the conversion and the print file are consistent as they us the same codepage.

tomcw:
>Using a single `char` makes the code simpler but will the result not depend on the codepage selected?

My understanding of codepages is not great, but yes (as I understand it) if the character is eg. á, and the AppleWin user is running on Windows with a codepage that doesn't have this character then the conversation will fail in the ctor, and then `ConvertToPrinterChar(BYTE value)` will just return `value` unmodified.

tomcw:
Looking again at what `ConvertToPrinterChar(BYTE value)` is doing: eg, based on a lookup in `Lat82[]` it returns the equivalent BYTE in `Kir82[]`.

`Kir82[]` was originally just a BYTE array of: 0xC0, 0xC1, ... 0xDA, 0xDC, 0xDE, 0xDF, 0x5B, 0x5D, 0x5E, 040... but written as a string of accented Latin characters.

These accented Latin characters are nothing to do with Cyrillic, so perhaps the Kir82 array was just a convenient way to express the above hex values?

EG. the Pravets printer understands these hex values as Cyrillic characters.

But the AppleWin.chm says the Config Advanced checkbox: "For Pravets 8 this converts from MIK to ASCII."
NB. [MIK](https://en.wikipedia.org/wiki/MIK_(character_set)) is an 8-bit Cyrillic code page from the Pravets 16.
Can I trust this AppleWin.chm info? (clearly the function converts from Latin/Ascii to accented Latin characters)

So perhaps it's simpler to replace the Kir82 (and Kir8A) arrays with an array of hex values instead?

audetto:
Definitely, if you replace with a literal "char *" using \x it will work.
The codepage thing is that the resulting file only has the desired meaning if one uses a latin1 codepage, which is probably good enough.

sicklittlemonkey:
> So perhaps it's simpler to replace the Kir82 (and Kir8A) arrays with an
array of hex values instead?

Hex values make much more sense, especially given that Wikipedia page about
MIK.

Who still has a Pravets setup? I know that this fellow enthusiast does:
https://twitter.com/cosmoza - but it seems we would need a printer to test
this.

tomcw:
It still totally clear to me... eg. converting the first char to a hex value of 0xC0 (from the `Kir82[]` array), then:
- _either_ outputting 0xC0 to a .txt file (perhaps makes sense if the code page is correctly set to MIK or Cyrillic, otherwise it's just gibberish).
- _or_ outputting 0xC0 directly to a real printer (which must be Pravets compatible to output the Cyrillic).

Much better to output to a virtual printer, eg. output to PDF format in Cyrillic.
But even that's a bit of a jump, since the description / comments for this feature are so poor that it's only a guess about what it truly intended.

As I suggested in #916- Perhaps better to just remove this feature?

audetto:
Given today's printers, it is very uncommon to send bytes directly to a printer (mine is over the network...).
The way I interpret "printer" in AppleWin is output to a text file which can then be sent to a printer (or not), using some other application.
And as soon as one needs non ASCII characters, the most universal standard at the moment seems to be UTF-8 which does not need any code page to be set anywhere.

But as your last question, I vote for removal.

sicklittlemonkey:
Sigh. If only we could print to a PDF. I need to do some more work on that
someday. ; - )

sicklittlemonkey:
It's uncommon but not impossible. Most still support ASCII, but
interestingly it seems some support UTF-8 too:
https://h30434.www3.hp.com/t5/LaserJet-Printing/Unable-to-print-some-Unicode-chars-using-PCL5/td-p/5686618

This doesn't help with the current dilemma though.

sicklittlemonkey:
I linked to the manual and character set in issue #916.

And in case anyone in the Apple II Bulgaria group is interested I pointed them to this thread.

PR #923, opened 2021-02-07
Enable ConformanceMode in VS2019.

audetto:
This reduces the chances of incompatibility between gcc and vs.

99% is "const char *" literals.
1 case of a goto jumping over a variable declaration.


tomcw:
Enable ConformanceMode means: /permissive-
Ref: https://docs.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-160

tomcw:
I've merged it in.

PR #937, opened 2021-03-18
Debugger: fix bound on maximum bytes to move backward.

audetto:
The higher g_nDisasmCurLine the more we will have to travel back to find the instruction at the top of the window.

In AppleWin it does not matter because the line is halfway and so the result is the same.

### DIFF
--- a/source/Debugger/Debugger_Disassembler.cpp
+++ b/source/Debugger/Debugger_Disassembler.cpp
@@ -627,7 +627,7 @@ void FormatDisassemblyLine(const DisasmLine_t& line, char* sDisassembly, const i
 //===========================================================================
 void DisasmCalcTopFromCurAddress(bool bUpdateTop)
 {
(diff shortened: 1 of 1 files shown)


tomcw:
@Michaelangel007 - if you get a moment can you cast your eye over this PR?

>In AppleWin it does not matter because the line is halfway and so the result is the same.

This looks correct from what I have observed, but perhaps there is an edge case I don't know about?

audetto:
If you need a bit of background.

When I was playing with the debugger in linux, I tried to move the current line up and down and realised that if I move it down after a bit it stops finding a solution.

Looking into it, it turns out that the more it is moved down, and the more it needs to look backward to find the instruction at the top of the page, the less it is allowed to do so.

On the other hand, the more I move it up, the more room it is given, but this is irrelevant as it needs less to find a solution.

tomcw (approved):

tomcw:
I've merged this to master now.

PR #940, opened 2021-03-28
Avoid leaks when the destructor of SY6522_AY8910 is not called.

audetto:
Moreover, it needs a constructor for correct initialisation.


tomcw:
Thanks - merged this in.

PR #943, opened 2021-04-17
Uthernet: enable support on linux

audetto:
A few small changes to make the TFE code compile and work on linux.

Only remaining issue is that it only works as root.


tomcw (approved):

tomcw:
Merged to master.

PR #945, opened 2021-04-23
REJECTED - Drag and drop / double-click support for .po and .hdv images

tomcw:
Discussion with Nick/@sicklittlemonkey in email:

> [Tom] In the 2x Windowed-mode we have spare space to add more icons, so we could just add 2 more for slot-7 HDD/PO devices 1 & 2. Most people must have a display that capable of 2x Windowed-mode these days!

[Nick] Hmm, a very good point, and in fact I always use AppleWin in 2x mode. Although I was just thinking about it when your email arrived - the regular drive icons could do double duty to cover non-2x mode. If you drop a large image on it, it should just mount to the appropriate HD. That's reasonable default behaviour. And hover could give info for both the respective drive & HD. Right clicking could have an eject (and eventually select ...) option for the respective HD too. 

 > [Tom] or **better** perhaps reboot with Open Apple pressed (which the HDD firmware detects and skips to the $C600 firmware)

[Nick] Ah! I didn't know that! Obvious but super handy! Hmm, it could probably be made even handier ... like a prompt. Maybe it's going too far, but an option to decide slot 7 boot behaviour might be nice. Alt to skip, Alt to boot, or prompt.

> [Tom] or **even better** maybe the "Open Apple press" is just the default in this case (so we dispense with the Message box).

[Nick] Yes, I agree. I think in the double-click and drop in main window cases, we should direct the card to boot. 

 I was thinking it looks like a simple approach might be:
- `WM_DDE_EXECUTE` and `WM_DROPFILES` could call a common function
- That would call `ImageOpen()` to analyse the image and decide which card to use
- `InsertDisk()` and `HD_Insert()` could be refactored to have variants which take an `ImageInfo*`

### DIFF
--- a/source/Disk.cpp
+++ b/source/Disk.cpp
@@ -38,6 +38,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #include "Core.h"
 #include "CPU.h"
 #include "DiskImage.h"
+#include "DiskImageHelper.h"
 #include "Log.h"
 #include "Memory.h"
 #include "Registry.h"
@@ -655,7 +656,7 @@ void Disk2InterfaceCard::GetLightStatus(Disk_Status_e *pDisk1Status, Disk_Status
 //===========================================================================
 
 // Pre: pszImageFilename is *not* qualified with path
-ImageError_e Disk2InterfaceCard::InsertDisk(const int drive, LPCTSTR pszImageFilename, const bool bForceWriteProtected, const bool bCreateIfNecessary)
+ImageError_e Disk2InterfaceCard::InsertDisk(const int drive, LPCTSTR pszImageFilename, const bool bForceWriteProtected, const bool bCreateIfNecessary, ImageInfo* pImageInfo /* = NULL */)
 {
 	FloppyDrive* pDrive = &m_floppyDrive[drive];
 	FloppyDisk* pFloppy = &pDrive->m_disk;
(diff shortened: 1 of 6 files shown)


audetto:
I think it would be really useful.

When I wrote the libretro core, I did not have any better options than to try both

https://github.com/audetto/AppleWin/blob/master/source/frontends/libretro/game.cpp#L29

Moreover in the imgui version I have added another option which I find very useful.
If the file dropped ends with `.yaml`, it is loaded as a save state. I use it to quickly try all the test yaml files.

And another goodie could be to load `text` files to the clipboard. Could be useful to build A2osX.

sicklittlemonkey:
Ugh, looks like a line-ending difference of opinion ruined the tiny contribution I made tonight. I'm outta time.

tomcw:
Hi Nick,
I'm no expert (and who is?!) but my git config (--system) has `core.autocrlf = true`, ie.

```
$ git config --system --get core.autocrlf
true
```

I read on stackoverflow [here](https://stackoverflow.com/questions/10418975/how-to-change-line-ending-settings), that
> It is a good idea to keep a .gitattributes file as we don't want to expect everyone in our team set their config. This file should keep in repo's root path and if exist one, git will respect it.

So do we need a `.gitattributes` file?

sicklittlemonkey:
Heh, as you can see, I'm no git expert either. As the latest commit says, I tried "git revert --no-commit HEAD~1" (one of the suggestions [here](https://stackoverflow.com/questions/4114095/how-do-i-revert-a-git-repository-to-a-previous-commit)) but pushing that still ended up with a commit. I think we might need to recreate this branch, sorry. I would hate to have any mess in the commit or file blame history.

I have now set core.autocrlf = true on this (newish) machine!

A .gitattributes file might be handy just to document which files are text and binary. This [SO question](https://stackoverflow.com/questions/170961/whats-the-best-crlf-carriage-return-line-feed-handling-strategy-with-git) linked to a [repo of handy samples](https://github.com/alexkaratarakis/gitattributes).

tomcw:
OK, lets abandon this PR, and switch to this new one: #950.

tomcw:
Closing.

PR #947, opened 2021-04-25
Uthernet: fix usability and settings

audetto:
The Uthernet card had 2 non standard behaviours:

1. It would not be reapplied after a restart (missing call to `tfe_reset`)
2. Its settings would not behave like everything else: needed a full restart and could not be undone (hardware changes)

This patch fixes the above issues and makes the Uthernet card behave like any other hardware-changing setting.

The logic is that when reading the settings, the enabled and the iface name are simply stored, and when `tfe_init` is called in https://github.com/AppleWin/AppleWin/blob/27fe4e5ed0e9ab6637bd58b4f55a110786ccace5/source/Windows/AppleWin.cpp#L945 the card is first removed and then recreated taking into account all (new) settings

### DIFF
--- a/source/Configuration/Config.h
+++ b/source/Configuration/Config.h
@@ -25,14 +25,8 @@ public:
 		m_Slot[SLOT5] = GetCardMgr().QuerySlot(SLOT5);
 		m_Slot[SLOT7] = GetCardMgr().QuerySlot(SLOT7);
 
-		int tfe_enabled;
-		get_tfe_enabled(&tfe_enabled);
-		m_tfeEnabled = tfe_enabled != 0;
-		void * iface = get_tfe_interface();
-		if (iface)
-		{
-			m_tfeInterface.assign((char *)iface);
-		}
+		m_tfeEnabled = get_tfe_enabled();
+		m_tfeInterface = get_tfe_interface();
 	}
 
 	const CConfigNeedingRestart& operator= (const CConfigNeedingRestart& other)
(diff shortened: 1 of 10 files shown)


audetto:
A couple of things could be simpler around `CConfigNeedegRestart`

1. Its constructor https://github.com/AppleWin/AppleWin/blob/27fe4e5ed0e9ab6637bd58b4f55a110786ccace5/source/Configuration/Config.h#L13 is almost the same as `CPropertySheetHelper::SaveCurrentConfig` https://github.com/AppleWin/AppleWin/blob/27fe4e5ed0e9ab6637bd58b4f55a110786ccace5/source/Configuration/PropertySheetHelper.cpp#L380 Could they be merged?
2. `CPropertySheetHelper::RestoreCurrentConfig` https://github.com/AppleWin/AppleWin/blob/27fe4e5ed0e9ab6637bd58b4f55a110786ccace5/source/Configuration/PropertySheetHelper.cpp#L399 is this not redundant? These settings are never actually stored until the user accepts them. Does it need a `Restore`?

audetto:
And further simplification if the TFE code stored the settings `tfe_enabled` and `tfe_interface` as `bool` and `std::string`.
If nobody objects, I will incorporate in this PR.

tomcw:
Hi @audetto - thanks for taking this Uthernet code on to refactor! It's a bit unusual, and as you commented it is non-standard compared to other card implementations in AppleWin.

Since it was heavily derived from other open-source (eg. VICE C64 emulator) I had been reluctant to make too many changes, in case I needed to merge (or update) from the original VICE source code. But the reality is that I never needed to do this, as it's always just worked (or been "good enough").

So anyway, I think you refactoring is fine, and also making the code (more) compliant with established code in AppleWin is probably the right thing to do.

audetto:
I know the feeling.
Of the changes in the original code, only `pcap_close_t` is really needed. The rest simplifies the logic.

If I kept the old code, there would be the usual c-string manipulation code repeated a few times.

tomcw:
Merged to master.

>A couple of things could be simpler around `CConfigNeedegRestart`
 :
>1) Could they be merged?

Probably... best to just try to simplify it, and if you can then just raise a new PR.

>2) `CPropertySheetHelper::RestoreCurrentConfig` - is this not redundant?

I don't know. I will take a look and test the code (I've not looked at this for years).

PR #951, opened 2021-05-07
Fix to sound not working in modern environments.

hasseily:
Need to use FAILED() instead of boolean check.
This is an update to the original PR https://github.com/AppleWin/AppleWin/pull/931/commits/ac214879e3581f93fbfa3fe85127f415068c806b


tomcw:
Thanks @hasseily - merged into master.

PR #955, opened 2021-05-22
Simplify logic around CConfigNeedingRestart

audetto:
1. separate class initialisation from loading settings (to avoid static variable initialisation loop)
2. include all slots (so to match the usage in SaveState.cpp)
3. in SaveState simply create an object before and after loading the yaml file, and compare them
4. remove RestoreCurrentConfig as it does not actually do anything (AW state is not touched so it does not need to be reset)

### DIFF
--- a/source/Configuration/PropertySheetHelper.cpp
+++ b/source/Configuration/PropertySheetHelper.cpp
@@ -376,23 +376,11 @@ void CPropertySheetHelper::ApplyNewConfig(void)
 void CPropertySheetHelper::SaveCurrentConfig(void)
 {
 	// NB. clone-type is encoded in g_Apple2Type
-	m_ConfigOld.m_Apple2Type = GetApple2Type();
-	m_ConfigOld.m_CpuType = GetMainCpu();
-	m_ConfigOld.m_Slot[SLOT3] = GetCardMgr().QuerySlot(SLOT3);
-	m_ConfigOld.m_Slot[SLOT4] = GetCardMgr().QuerySlot(SLOT4);
-	m_ConfigOld.m_Slot[SLOT5] = GetCardMgr().QuerySlot(SLOT5);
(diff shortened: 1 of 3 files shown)


source/Configuration/PropertySheetHelper.cpp
```diff
@@ -405,7 +395,6 @@ void CPropertySheetHelper::RestoreCurrentConfig(void)
 	SetSlot(SLOT5, m_ConfigOld.m_Slot[SLOT5]);
 	HD_SetEnabled(m_ConfigOld.m_Slot[SLOT7] == CT_GenericHDD);
 	GetPropertySheet().SetTheFreezesF8Rom(m_ConfigOld.m_bEnableTheFreezesF8Rom);
-	m_ConfigNew.m_videoRefreshRate = m_ConfigOld.m_videoRefreshRate;	// Not SetVideoRefreshRate(), as this re-inits much Video/NTSC state!
```
tomcw, line 408: Why do you remove this line?
If the video frequency changes (50/60Hz) then it won't get reverted on a cancelled Configuration.

tomcw (commented):
Other than the comment about `m_ConfigNew.m_videoRefreshRate` not being restored, this all looks good.

source/Configuration/PropertySheetHelper.cpp
audetto: Because it looks to me it is dead code.
After this line, the next time a call to `SaveCurrentConfig` will overwrite `m_ConfigNew` completely.
And why restoring just `m_videoRefreshRate`?

If one wanted to keep the state super consistent, one could add a line like

`m_ConfigNew = m_ConfigOld;`

to ensure the config reflects the state?

tomcw: Yes, you are right - the video freq is only used (ie. set in the Registry) by `ApplyNewConfig()`... but for the _cancel_ path, there's nothing to revert.

tomcw:
Thanks @audetto - merged it in.

PR #956, opened 2021-05-22
Missing cards and yaml files

audetto:
This patch is orthogonal to https://github.com/AppleWin/AppleWin/pull/955.

A yaml file only contains information about present cards.
So if a slot is empty (e.g. 4), the yaml file contains no info about it.

Should the system after a state reload keep the Slot 4 **unchanged** or make it **empty**?

Either way, the current code behaves differently according to which cards are in slot 4 and 5.
If a Mouse card is in slot 4, then it will be reset by a yaml file, anything else will stay.

I personally think all cards missing from a yaml should go to their default state.

### DIFF
--- a/source/SaveState.cpp
+++ b/source/SaveState.cpp
@@ -480,11 +480,10 @@ static void Snapshot_LoadState_v2(void)
 			ConfigOld.m_Slot[2] = CT_Empty;
 		}
 
-		if (GetCardMgr().QuerySlot(SLOT4) == CT_MouseInterface)
-			GetCardMgr().Remove(SLOT4);		// Remove Mouse card from slot-4
-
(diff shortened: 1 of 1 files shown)


tomcw:
I agree with what you are saying here, but we should be able to remove all the cards (not just for slot 4 & 5).

tomcw:
I will give it a go, and so will possibly end up just closing this PR.

tomcw:
This PR was covered by commit ed298b4fd90da874abb54e5ac21ff80e9f4fa8a0 (so closing this PR).

>Should the system after a state reload keep the Slot 4 unchanged or make it empty?
All slots are first emptied (ie. all cards removed) before cards get added from the save-state file.

PR #959, opened 2021-05-25
Disks: save full pathname to registry.

audetto:
This was already happening for Hard Disks (although in a convoluted way).
Extend to Floppy Disks.


tomcw:
Thanks Andrea - merged it into master.

PR #973, opened 2021-07-18
Move some "char *" -> "const char *", and avoid non necessary c_str().

audetto:
A few small changes to improve string handling.


tomcw:
Thanks @audetto - merged this into master.

Issue #977, opened 2021-08-14
Improve support for slot configuration in the Registry


tomcw:
Currently in the Registry we have:
- "Preferences" with "Last Disk Image 1" and "Last Disk Image 2" which are just for slot-6

This doesn't scale for adding more Disk II controllers in other slots.
And in general it doesn't scale for configuration of cards in other slots.

The proposal here is to have the following hierarchy:
```
+ Configuration
  + Slot Auxiliary
    + Card type = ...
  + Slot 0
    + Card type = ...
  + Slot 1
    + Card type = ...
    :
  + Slot 6
    + Card type = 1
    + Last Disk Image 1
    + Last Disk Image 2
  + Slot 7
    + Card type = ...
```

(where Card type=1 is Disk II Controller)

tomcw:
In the Registry, AppleWin has:
```
+ CurrentVersion
  + Configuration
    + ...
  + Preferences
    + ...
```

Just focusing on `Configuration` and `Preferences`... what is the intention of each?

I found this question asked on StackOverflow [here](https://stackoverflow.com/questions/2074384/options-settings-properties-configuration-preferences-when-and-why).

>Following an approximate lead from Visual Studio and other Microsoft products:
>- _Properties_...
>- _Options_...
>- _Settings and Preferences_ change qualities of how the application works. The implication here is to change, not disable: for example, "Metric measurements" or "British Imperial measurements".
>- _Configuration_ is often where an application is customised for each user or group. 

So applying to AppleWin:
- Configuration:
  - Apple II machine's hardware configuration, eg. the cards in each slot
- Preferences:
  - AppleWin's window's x/y start-up position
  - Starting Directory, eg. for screenshots
  - Window Scale (eg. x1, x2) -- currently under Configuration

But not everything is obviously one or the other.
And some things (Configuration->Version) shouldn't be stored under either!

Anyway, the proposed hierarchy for slots should be under `Configuration`, even though it will include pathnames like `Last Disk Image 1` (for Disk II card).

Issue #975, opened 2021-07-28
feature request: slot 5 UI config


peterferrie:
I've been using the "-s5 diskii" command-line option for a second disk controller.  It's great.
However, somehow a "Slot 5" value got saved to the registry, causing it to be permanently in use as a drive.
That blocked slot 5 for sound when later run without the "-s5" option, until I edited the registry.
I would like to see a UI element to control the selection in slot 5 (or that slot 5 isn't ever saved to the registry).

tomcw:
Hi Peter - thanks for the report and request.
In fact it's good to hear of someone actually using this "-s5 diskii" feature!

I was able to repro this by doing:
- create save-state1: s4=phasor, s5=empty
- create save-state2: s4=empty, s5=diskii
- load state1
- load state2  (this saves "Slot 5" to Registry with a value of "1" ie. diskii)

I tried to revert using command line "-s5 empty", but this isn't currently supported for slot 5 (nor slots 4 and 7!).
Workarounds are: manually edit the Registry (or conf.ini), or load a prior save-state with no diskii in slot 5.

>I would like to see a UI element to control the selection in slot 5 (or that slot 5 isn't ever saved to the registry).

It's definitely easier (as a quick fix) not to save "Slot 5" = "diskii" to the Registry.

A minimum UI would be:
- Configuration Disk UI: check box for a Disk II controller in slot 5
- Configuration Disk UI: 2 extra combo drop-downs for the 2 extra drives

NB. There's already a tooltip over the Drive-1/2 UI buttons to show what's in the slot 5 drives.

Let me try to squeeze in this minimum UI.

tomcw:
[Note to self] Something to think about is I need to consider how to persist the slot-5 image names to the Registry.
OTTOMH:
- Extend "Preferences" with "Last Disk Image 1 (s5)" and "Last Disk Image 2 (s5)"
  - And keep "Last Disk Image 1" and "Last Disk Image 2" implicitly for slot 6.
- Or create a sub-folder "Preferences\Last Disk Image", with value names "s6,d1", "s6,d2", "s5,d1", "s5,d2"
  - And deprecate "Last Disk Image 1" and "Last Disk Image 2".

tomcw:
>I need to consider how to persist the slot-5 image names to the Registry.

Spun this off into a generalised approach in #977.

tomcw:
[...]

PR #978, opened 2021-08-27
Config GUI for DiskII card in slot-5 & improved slot config in Registry

tomcw:
PR for #975 and #977


tomcw:
@audetto - this conflicts with and relates to your PR #955.

I'm now using `RestoreCurrentConfig()`, since I may have created a Disk2InterfaceCard object (for slot-5) from the Config->Disk property sheet page. But need to restore to the previous card if the user cancels their configuration choices.

I know that a better way of supporting this GUI configuration is to create copies of the card objects for all the slots, then let the user configure them (eg. insert images into drives attached to a slot-5 card). Then if the user cancels, just delete the objects. And if the user OK's the config, then use these new objects and delete the old ones.

But the blocker at the moment is that not all cards are implemented as classes, eg. Mockingboard is hardcoded, across slots 4 and 5 (and doubles up as a Phasor card too!).

I think this current extension is "good enough" (for now), but I want to next add classes for cards that are selectable, eg. the Mockingboard, Phasor, SAM, Hard disk and CP/M cards.

PR #995, opened 2021-10-31
Make harddisk a class

tomcw:


audetto:
Hi
I saw the work to make HD a class and I totally agree with it.
I was wondering:

1. are you planning to make `InitializeIO` a virtual function in `Card` too?
2. could `Update` become a virtual function in each `cart` too? https://github.com/AppleWin/AppleWin/blob/d96ed5b8c55f36c02bbdd9df9ed0d80e1b787e11/source/Windows/AppleWin.cpp#L220-L223
so that each card has its own time to process whatever it needs to be processed (if anything at all)

and then one only has to call the equivalent on `CardManager` that would loop on all cards?

tomcw:
Yes, both of these suggestions are good - so I'm happy to take a PR, or if you prefer I can make the changes soon.

btw, my intention is to gradually make all the remaining cards into classes, eg. Printer, Uthernet, Z80, Mockingboard/Phasor + the aux 80-col cards.

PR #999, opened 2021-11-08
Card::InitializeIO

audetto:
Add `Card::InitializeIO`: so one can remove all the inlined checks for card type.

Other areas of improvement:

- `Card::Reset()` probably exactly the same could be done: see https://github.com/AppleWin/AppleWin/blob/ff7c9dc185aed1352e570c583c206af900a8e4f8/source/Utilities.cpp#L541-L559
- `Card::Update(executedCycles)`: https://github.com/AppleWin/AppleWin/blob/ff7c9dc185aed1352e570c583c206af900a8e4f8/source/Windows/AppleWin.cpp#L220-L223
- `Card::Destroy()`?

### DIFF
--- a/AppleWinExpress2008.vcproj
+++ b/AppleWinExpress2008.vcproj
@@ -658,6 +658,10 @@
 					>
 				</File>
 				<File
+					RelativePath=".\source\Card.cpp"
+					>
+				</File>
+				<File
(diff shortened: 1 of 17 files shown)


tomcw:
Great. This is much simpler. (Merged to master)

PR #1000, opened 2021-11-12
Card::Update()

audetto:
Add a method to allow cards to execute their update code.

A couple of points:

- `GetCardMgr().GetDisk2CardMgr().Update(executedcycles);` is only used in the benchmark code. Should it become `CardManager::Update()`?

https://github.com/AppleWin/AppleWin/blob/8662a9917964d0ee1005829153ff0bbb082b813f/source/Windows/Win32Frame.cpp#L299

- This line https://github.com/AppleWin/AppleWin/blob/8662a9917964d0ee1005829153ff0bbb082b813f/source/Windows/Win32Frame.cpp#L300 passes ``cycles`` rather then ``nExecutionPeriodUsec``, is it ok?

- Looking at the rest of the code e.g. https://github.com/AppleWin/AppleWin/blob/8662a9917964d0ee1005829153ff0bbb082b813f/source/Utilities.cpp#L539 I wonder if there should be a small set of well defined operations that all card should implement

1. Boot
2. Init
3. Reset(false) (CommReset?)
4. Reset(true) (CommReset?)
5. Destroy

But I am not sure I can propose something meaningful. I don't really understand the subtle differences.

### DIFF
--- a/source/Card.cpp
+++ b/source/Card.cpp
@@ -52,3 +52,21 @@ void DummyCard::InitializeIO(LPBYTE pCxRomPeripheral)
 		break;
 	}
 }
+
+void DummyCard::Update(const ULONG nExecutedCycles)
+{
+	switch (QueryType())
+	{
+	case CT_GenericPrinter:
+		PrintUpdate(nExecutedCycles);
+		break;
+	case CT_MockingboardC:
+	case CT_Phasor:
+		// only in slot 4
+		if (m_slot == SLOT4)
+		{
+			MB_PeriodicUpdate(nExecutedCycles);
+		}
+		break;
+	}
+}
--- a/source/Card.h
+++ b/source/Card.h
@@ -36,6 +36,7 @@ public:
 	virtual void InitializeIO(LPBYTE pCxRomPeripheral) = 0;
 	virtual void Init(void) = 0;
 	virtual void Reset(const bool powerCycle) = 0;
+	virtual void Update(const ULONG nExecutedCycles) = 0;
 	SS_CARDTYPE QueryType(void) { return m_type; }
 
 protected:
@@ -56,6 +57,7 @@ public:
 	virtual void InitializeIO(LPBYTE pCxRomPeripheral) {}
 	virtual void Init(void) {}
 	virtual void Reset(const bool powerCycle) {}
+	virtual void Update(const ULONG nExecutedCycles) {}
 };
 
 //
@@ -69,4 +71,5 @@ public:
 	virtual void InitializeIO(LPBYTE pCxRomPeripheral);
 	virtual void Init(void) {}
 	virtual void Reset(const bool powerCycle) {}
+	virtual void Update(const ULONG nExecutedCycles);
 };
--- a/source/CardManager.cpp
+++ b/source/CardManager.cpp
@@ -199,3 +199,14 @@ void CardManager::InitializeIO(LPBYTE pCxRomPeripheral)
 		}
 	}
 }
+
+void CardManager::Update(const ULONG nExecutedCycles)
+{
+    for (UINT i = 0; i < NUM_SLOTS; ++i)
+    {
+        if (m_slot[i])
+        {
+            m_slot[i]->Update(nExecutedCycles);
+        }
+    }
+}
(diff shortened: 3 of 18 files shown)


tomcw:
re. `JoyUpdateButtonLatch()` - it is confusing what the purpose of this actually is! ie. `int buttonlatch[3]`

The TL;DR summary is that it's harmless as-is, and can probably be completely removed. But let's leave it for now. I'll do some checks and will aim to remove in a latter commit (unrelated to this PR).

---

Over time `JoyUpdateButtonLatch()` has morphed away from the original AppleWin 1.00 code, where `JoyUpdateButtonLatch()` was called once every 2048 emulated cycles, and each `buttonlatch` was decremented by 1 each update. NB. 2048 x BUTTONTIME(5000) = 10M cycles = 10 seconds.

The logic in `CheckJoystick0()` was (and still is) like this:
- if button-0 is pressed:
  - if button-0 wasn't pressed last time, then `buttonlatch[0] = BUTTONTIME` (ie. 5000)
  - `joybutton[0] = TRUE`
- if button-0 not pressed:
  - `joybutton[0] = FALSE`

Then later when the Apple II reads $C061 (BUTTON0 I/O port):
- `pressed = buttonlatch[0] || joybutton[0]`
- `buttonlatch[0] = 0`
- `return pressed<<7`

So the original code would latch the button-0 press for up to 10 seconds!

I've never heard of this behaviour, and there's nothing in Sather (Understanding the Apple II) about this either.
**Easy to check on real h/w though.**

But today, the main `ContinueEmulation()` loop decrements each `buttonlatch` by ~1000 (or ~1ms) each loop, so the button down state is only latched for ~5ms.

NB. The comment in Joystick.cpp after BUTTONTIME about "debounce" is mine, and it is wrong.

tomcw:
>I wonder if there should be a small set of well defined operations that all card should implement

Yes, I think so. We can gradually extend this small set as and when it make sense.

btw, No cards currently use `Init()` - but they all implement it as `virtual void Init(void) {}`. As things are still evolving, then I think leave it for now, but we can review whether to remove it later on.

audetto:
> btw, No cards currently use `Init()` - but they all implement it as `virtual void Init(void) {}`. As things are still evolving, then I think leave it for now, but we can review whether to remove it later on.

I was thinking in particular about: Reset(true), Reset(False), Destroy(). (and add CommReset() & CommDestroy() which seem to be equivalent under different spelling)

Is there a rule when the first 3 are meant to be called?

tomcw:
Destroy() functions should only be called for the WM_DESTROY message.
So CommDestroy() can be renamed to Destroy().
And adding `virtual void Card::Destroy(void) = 0;` sounds like a sensible thing to do.

Reset(false) is for CTRL+RESET.
Reset(true) is for a power-cycle, eg. the h/w configuration has changed, so the emulator goes through the "restart" routine.
So CommReset() can be renamed to Reset(const bool powerCycle), and the `powerCycle` value just ignored for the SSC card.

tomcw:
Thinking some more... maybe we drop Init() and assume it's done in the constructor. And similarly drop Destroy() and do it in the destructor. So WM_DESTROY msg would remove all the cards, and so destruct them all.

sicklittlemonkey:
Another option, rather than separate functions for all (or some of) these
things would be a single Event method on Cards which could use a switch
statement if needed. I wonder if that would differ in performance.

Anyway, all this consolidation of code is great!

tomcw:
Makes sense if we plan (or anticipate) to add more functions in future, as the interface remains unchanged - so nicely extensible, without having to update the derived classes for all the specific cards each time.

NB. You'd pass a void* to the Event() method to encapsulate the param (or struct containing the params).

We should consider this next time we need to add a new function.
(This time it we remove Init(), and put Destroy() in the dtor, then no interface changes need to be made to Card.)

audetto:
While making some further changes, I have realised that the 2 yaml files for test: `4Play-slot4.aws.yaml` and `SNES_MAX-slot4.aws.yaml` result in a black screen. (Using the released 1.30.6.0).

What should I see when loading these 2 state files? I wanted to make sure I don't break anything.

tomcw:
>What should I see when loading these 2 state files?

Ah, the blank screen is correct.
But to give a bit more confidence, I've just updated these 2 state files with a simple title (for the card).

To run the test, you need to Configure to use a PC Joystick-1, and it needs to be a joystick like a PlayStation 4 or 5 or XBox controller.

PR #1006, opened 2021-12-05
Slot0: make it more uniform to other slots.

audetto:
While I was working on the other PRs re Card interface, I realised that the `Slot0` card information is stored in separate locations

1. the `Slot0` type (but the card is always a `DummyCard`)
2. the `g_pLanguageCard` global variable (which contains the instance)

This change moves all the information to the slot 0 of the `CardManager` and I think simplifies the logic.

There is an important point not to forget, is that for a //e the Slot 0 MUST ALWAYS contains a `LanguageCardUnit` (of type `CT_LanguageCardIIe`). And I have added an `ASSERT` about this, so it cannot be missed.

This is not new, it used to be done in https://github.com/AppleWin/AppleWin/blob/443545b0f6d31349eb365956ab948561e889cc15/source/Memory.cpp#L352

### DIFF
--- a/source/CardManager.cpp
+++ b/source/CardManager.cpp
@@ -104,14 +104,19 @@ void CardManager::InsertInternal(UINT slot, SS_CARDTYPE type)
 		break;
 
 	case CT_LanguageCard:
-		m_slot[slot] = m_pLanguageCard = new LanguageCardSlot0();
+		_ASSERT(m_pLanguageCard == NULL);
+		if (m_pLanguageCard) break;	// Only support one language card
+		m_slot[slot] = m_pLanguageCard = new LanguageCardSlot0(type, slot);
 		break;
 	case CT_Saturn128K:
-		m_slot[slot] = m_pLanguageCard = new Saturn128K(Saturn128K::GetSaturnMemorySize());
+		_ASSERT(m_pLanguageCard == NULL);
+		if (m_pLanguageCard) break;	// Only support one language card
+		m_slot[slot] = m_pLanguageCard = new Saturn128K(type, slot, Saturn128K::GetSaturnMemorySize());
 		break;
-
-	case CT_LanguageCardIIe:	// not a card
-		m_slot[slot] = m_pLanguageCard = new LanguageCardUnit();
+	case CT_LanguageCardIIe:
+		_ASSERT(m_pLanguageCard == NULL);
+		if (m_pLanguageCard) break;	// Only support one language card
+		m_slot[slot] = m_pLanguageCard = new LanguageCardUnit(type, slot);
 		break;
 
 	default:
@@ -132,17 +137,27 @@ void CardManager::Insert(UINT slot, SS_CARDTYPE type, bool updateRegistry/*=true
 
 void CardManager::RemoveInternal(UINT slot)
 {
-	if (m_slot[slot] && m_slot[slot]->QueryType() == CT_MouseInterface)
-		m_pMouseCard = NULL;	// NB. object deleted below: delete m_slot[slot]
+	if (m_slot[slot])
+	{
+		// NB. object deleted below: delete m_slot[slot]
+		switch (m_slot[slot]->QueryType())
+		{
+		case CT_MouseInterface:
+			m_pMouseCard = NULL;
+			break;
+		case CT_SSC:
+			m_pSSC = NULL;
+			break;
+		case CT_LanguageCard:
+		case CT_Saturn128K:
+		case CT_LanguageCardIIe:
+			m_pLanguageCard = NULL;
+			break;
+		}
 
-	if (m_slot[slot] && m_slot[slot]->QueryType() == CT_SSC)
-		m_pSSC = NULL;			// NB. object deleted below: delete m_slot[slot]
-
-	if (slot == 0)
-		m_pLanguageCard = NULL;
-
-	delete m_slot[slot];
-	m_slot[slot] = NULL;
+		delete m_slot[slot];
+		m_slot[slot] = NULL;
+	}
 }
 
 void CardManager::Remove(UINT slot, bool updateRegistry/*=true*/)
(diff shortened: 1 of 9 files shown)


source/CardManager.h
```diff
@@ -9,7 +9,8 @@ class CardManager
 public:
 	CardManager(void) :
 		m_pMouseCard(NULL),
-		m_pSSC(NULL)
+		m_pSSC(NULL),
+		m_pLanguageCard(NULL)
 	{
 		InsertInternal(SLOT0, CT_Empty);
```
tomcw, line 15: Should this slot0 insertion be for `CT_LanguageCardIIe` now? (ie. the default for a //e)

audetto: What happens in the constructor is almost irrelevant.

The actual card is inserted during `SetExpansionMemType` or during the loading of the yaml file.

The only case which requires special care is when loading a yaml file which does not contain any information about a slot 0 card (which is the case for a //e).

https://github.com/audetto/AppleWin/blob/8b8ef8624b51d20717998f617ac2fe4a79900342/source/SaveState.cpp#L466-L470

And this is why I have added an ASSERT at the point of no return to catch any issue: https://github.com/audetto/AppleWin/blob/8b8ef8624b51d20717998f617ac2fe4a79900342/source/CardManager.cpp#L202

Maybe one could make this more explicit add a `.clear()` method to the `CardManager` which puts `CT_Empty` everywhere, except on a //e when the correct slot 0 **must** be inserted. This is not exactly the same a the constructor which inserts a couple more things.

audetto: Maybe to avoid code duplication, I should replace the lines https://github.com/audetto/AppleWin/blob/8b8ef8624b51d20717998f617ac2fe4a79900342/source/SaveState.cpp#L469-L470 with `SetExpansionMemType(CT_Emmpty)`?

tomcw: Perhaps... but what if `g_Apple2Type` changes? - Which can happen later in `ParseUnit()`.

The important case is when it changes from "not a //e" to a "//e or above", as this is the case when you need to call `GetCardMgr().Insert(SLOT0, CT_LanguageCardIIe)`.

Is this not guaranteed to happen later on when loading the save-state?

audetto: Correct.
I have regenerated it. The code is more similar that the original version. It happens like before in `CreateLanguageCard` which now is just checking the inserted cards are correct, and if not, fixing them.

tomcw: Thanks - let me resolve this conversation.

source/CardManager.cpp
```diff
@@ -138,6 +138,9 @@ void CardManager::RemoveInternal(UINT slot)
 	if (m_slot[slot] && m_slot[slot]->QueryType() == CT_SSC)
 		m_pSSC = NULL;			// NB. object deleted below: delete m_slot[slot]
 
+	if (slot == 0)
```
tomcw, line 141: There's a discrepancy here:
- InsertInternal() lets you insert a Language Card or Saturn in any slot and will always update `m_pLanguageCard` (for any slot)
- RemoveInternal() only sets `m_pLanguageCard = NULL` for slot 0

Perhaps best to only allow Lang Card / Saturn to be inserted into slot 0?

audetto: They can indeed be only inserted in slot 0 (sort of) because of

https://github.com/audetto/AppleWin/blob/ffa869f610c98bf04cc05820304dac5f447fde6a/source/LanguageCard.cpp#L42

on the other hand these cards ignore the `slot` in https://github.com/audetto/AppleWin/blob/ffa869f610c98bf04cc05820304dac5f447fde6a/source/CardManager.cpp#L107-L114

which overall is not great either.

So I will pass the `slot` down and still `throw` in the constructor of their base class `LanguageCardUnit` if the slot is not 0.
So the net effect will be zero, but a bit more explicit.

Moreover, what you reported needs to be made more self explanatory. But I need to try a few options to see what looks nicer.

tomcw: Closing this & starting a new review...

source/CardManager.cpp
```diff
@@ -104,14 +104,19 @@ void CardManager::InsertInternal(UINT slot, SS_CARDTYPE type)
 		break;
 
 	case CT_LanguageCard:
-		m_slot[slot] = m_pLanguageCard = new LanguageCardSlot0();
+		_ASSERT(m_pLanguageCard == NULL);
+		if (m_pLanguageCard) break;	// Only support one language card
+		m_slot[slot] = m_pLanguageCard = new LanguageCardSlot0(type, slot);
```
tomcw, line 109: I don't like this `new LanguageCardSlot0(type, slot)` - since you shouldn't have to pass `type`, as it's implicit in the object you are creating.

Same for the other 2 lines below:
- `new Saturn128K(type, slot, Saturn128K::GetSaturnMemorySize())`
- `new LanguageCardUnit(type, slot)`

Maybe just don't pass `type` ?

audetto: I knew you were going to say it.
The reason is really boring, but here it is

1. except `Saturn128K` the other 2 are both base and concrete classes, so they need to accept a type or already have it
2. the type is the 1st argument, so I cannot give it a default value (before, the slot, 2nd arg was NOT passed)
3. I could flip the order of the 2 arguments, but this would be misleading
4. I could write 2 constructors, but I don't think VS2008 can chain them
5. I could leave the constructor `protected` for inheritance and have a new `create` for the concrete class...
6. in the end I did not do any of the above

boring, innit?

audetto: I have committed option 4, but I don't know if VS2008 will accept it. I cannot check.

audetto: Actually, storm in a teapot. Latest version is almost the same and works everywhere.

tomcw: Great... I'm going to commit this to master.

PR #1011, opened 2021-12-17
throw std::runtime_error

audetto:
Change all `throw std::string` to `std::runtime_error`.

And add some extra info to `yaml` errors.


tomcw:
Thanks - merged this to master.

PR #1012, opened 2021-12-18
Clean up some warnings.

audetto:
1) enable warning to all remaining "printf" functions (and fix a couple of mistakes)
2) clean up 2 declarations
3) remove redundant ()

### DIFF
--- a/source/Debugger/Debugger_Console.h
+++ b/source/Debugger/Debugger_Console.h
@@ -257,11 +257,17 @@
 	// Buffered
 	bool ConsolePrint( const char * pText );
 	bool ConsolePrintVa( char* buf, size_t bufsz, const char* pFormat, va_list va );
+
 	template<size_t _BufSz>
 	inline bool ConsolePrintVa( char (&buf)[_BufSz], const char* pFormat, va_list va )
 	{
 		return ConsolePrintVa(buf, _BufSz, pFormat, va);
 	}
+
+#ifndef _MSC_VER
+	inline bool ConsolePrintFormat( char* buf, size_t bufsz, const char* pFormat, ... ) __attribute__ ((format (printf, 3, 4)));
+#endif
+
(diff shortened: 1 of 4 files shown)


source/Debugger/Debugger_Console.h
```diff
@@ -257,11 +257,17 @@
...
 	bool ConsolePrintVa( char* buf, size_t bufsz, const char* pFormat, va_list va );
+
 	template<size_t _BufSz>
 	inline bool ConsolePrintVa( char (&buf)[_BufSz], const char* pFormat, va_list va )
 	{
 		return ConsolePrintVa(buf, _BufSz, pFormat, va);
 	}
+
+#ifndef _MSC_VER
+	inline bool ConsolePrintFormat( char* buf, size_t bufsz, const char* pFormat, ... ) __attribute__ ((format (printf, 3, 4)));
```
tomcw, line 268: What does this `__attribute__ ((format (printf, 3, 4)))` do?

audetto:
So, it tells the compilers (`gcc` and `clang` so far, I don't know VS) that the arguments 3 and 4 follow the exact same rules as the arguments of `printf`.

This way the compiler can do a compile-time check that the variable arguments provided match exactly the `%` formats in the literal string.

This is how I found out that `%s` was missing from `"File: "` and it warns that `%d` is not the correct format for a `sizeof`, which should be `%zu` for maximum portability and correctness.

audetto:
Wait a minute. Which compiler are we targeting?
This code works well in VS2019

Screenshot from 2021-12-19 16-00-08

But I haven't got VS2008 any longer.

and moreover, VS can do that too

https://devblogs.microsoft.com/cppblog/format-specifiers-checking/

but they are not all enabled by default.

and I can enable the custom checks in VS too

https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values?view=msvc-170

Maybe I can try and then we revisit this PR.

audetto:
Postponed to after various string replacements.

PR #1014, opened 2022-01-02
Correct Bmp creation on Linux

audetto:
1. Header was not packed properly
2. static assert did not work on gcc / clang
3. added `FrameBase::Video_GetScreenShotFolder` to let frontends customise where to save the screenshots
4. added the size to the bitmap header: to make it more compliant


tomcw:
Looks good, thanks! (merged)

PR #1027, opened 2022-01-28
writes to PIC strobe should not be interpreted as a data byte

sh95014:
Only send writes to address $C090 (although the code is generalized in case the PCI is inserted into a different slot) to the printer. PrintShop was also writing to the strobe ($C092) which should not be interpreted as a data byte.

Tested with PrintShop configured for a "Epson MX-, FX-, RX- (80 or 100)" connected to "Apple II Parallel Interface" in slot 1.

This still does not cleanly pass the bytes to an Epson printer. You'll also want to set `g_bFilterUnprintable` to `false` because that filters out the ESC character that the printer needs, and `g_bPrinterAppend` to `true` because the default setting has a bad habit of resetting the printout.

### DIFF
--- a/source/ParallelPrinter.cpp
+++ b/source/ParallelPrinter.cpp
@@ -142,21 +142,24 @@ static BYTE __stdcall PrintStatus(WORD, WORD, BYTE, BYTE, ULONG)
 }
 
 //===========================================================================
-static BYTE __stdcall PrintTransmit(WORD, WORD, BYTE, BYTE value, ULONG)
+static BYTE __stdcall PrintTransmit(WORD, WORD address, BYTE, BYTE value, ULONG)
 {
 	if (!CheckPrint())
 		return 0;
 
-	BYTE c = value & 0x7F;
+    if ((address & 0xF) == 0) {
+        // load output port (i.e., $C090)
+        BYTE c = value & 0x7F;
 
-	if (IsPravets(GetApple2Type()))
-	{
-		if (g_bConvertEncoding)
-			c = GetPravets().ConvertToPrinterChar(value);
-	}
+        if (IsPravets(GetApple2Type()))
+        {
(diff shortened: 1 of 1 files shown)


source/ParallelPrinter.cpp
```diff
@@ -142,21 +142,24 @@ static BYTE __stdcall PrintStatus(WORD, WORD, BYTE, BYTE, ULONG)
...
 
 //===========================================================================
-static BYTE __stdcall PrintTransmit(WORD, WORD, BYTE, BYTE value, ULONG)
+static BYTE __stdcall PrintTransmit(WORD, WORD address, BYTE, BYTE value, ULONG)
 {
 	if (!CheckPrint())
 		return 0;
 
-	BYTE c = value & 0x7F;
+    if ((address & 0xF) == 0) {
```
tomcw, line 150: Can you move the open-brace to a newline.

tomcw: FYI, see the coding conventions doc here:
https://github.com/AppleWin/AppleWin/blob/master/docs/CodingConventions.txt

sicklittlemonkey: My 2 cents. I always try to minimise a diff. When you come back later (blame command, etc) you only want to see lines that changed functionality, not indented whitespace etc.

In this case we can just do:
```
if ((address & 0xF) != 0)
    return 0;
```

tomcw: @sh95014 - can you make the change that @sicklittlemonkey suggested, and then I can merge this PR. Thanks.

sh95014: Yup, will do. Was just traveling and couldn't get to it yet.

tomcw:
Thanks @sh95014 - merged to master.

PR #1030, opened 2022-02-05
Add undefined int8_t type for VS2008.

ralph-irving:
Fixes Debugger_Disassembler.cpp build failure at struct FAC_t exponent field definition.


tomcw:
Thanks - but you just got beaten to this by the PR #1029 which I've just merged to master.

PR #1031, opened 2022-02-06
Refactor OutputDebugString() usage

kiyolee:
Add `OutputDebugFormat()` to simplify the common combination of `sprintf()` and `OutputDebugString()`.
Use `_vsnprintf_s()` behind the scene to prevent buffer overflow.

Other changes:
- Add `SIZE_T_FMT` to properly print size_t variables
- `va_end()` should be called following standard
- Fix to support UNICODE properly in `Log.cpp`
- Call `OutputDebugStringA()` directly for cases that cannot be UNICODE


tomcw:
Hi @kiyolee - it seems like a good idea to combine sprintf() + OutputDebugString() into a single function.
I think my preference is to just combine by replacing them with the existing LogOutput() function, and not adding this new function.

Nick / @sicklittlemonkey: do you have a view?

kiyolee:
`LogOutput()` is indeed functionally the same as my new `OutputDebugFormat()`.
The difference is `LogOutput()` is either ANSI or UNICODE depends on compile switch.
My `OutputDebugFormat()` as C++ template supports both uses at the same time.
I can easily turn `LogOutput()` into more generic version, essentially replacing with my suggested implementation, and replace all calls to `OutputDebugFormat()` to `LogOutput()`.

audetto:
> * Add `SIZE_T_FMT` to properly print size_t variables

I like this.
See https://github.com/AppleWin/AppleWin/pull/1012 where I proposed to add info for the compiler to check the printf formats vs the actual data.

It is currently blocked because I don't know what works in VS2008.
Does `SIZE_T_FMT` work there?

kiyolee:
@audetto `SIZE_T_FMT` is conditionally defined to work for all VS and gcc/clang (except those that are too old to be a problem).

sicklittlemonkey:
I'm all for unifying these and making it simpler/better. But this is still pretty ugly:
```
    char szDebug[100];
    OutputDebugFormat(szDebug, "HDD Entrypoint: $%04X\n", PC);
```

How about if we standardise, have:
void LogOutput(LPCTSTR format, ...) // Uses a default buffer size
void LogOutput(int buffer_size, LPCTSTR format, ...) // Use the given buffer size

kiyolee:
@sicklittlemonkey Agree with your suggestion. I'll update my PR to do that unless somebody objects.

audetto:
> void LogOutput(int buffer_size, LPCTSTR format, ...) // Use the given buffer size

Does this add a lot?
It is still a compile time decision, so why not update the default buffer directly?

I fear these cases will go unnoticed unless something is logged (to the console?) about buffer being too small.

Alternatively, without a newer version of `snprintf`, one can still iteratively increase the buffer size, and if it stored in a static `string` or `vector`, then this effectively becomes a O(1) operation.

audetto:
Does it need to template on the `char` type?
I have seen `char`, `TCHAR`, and `wchar`, but are they really meant to be used or just a typing accident?

kiyolee:
@audetto I can asure you those types used in my PR are all intentional. The problem was `OutputDebugString()` are called with both `TCHAR` and `char` buffers in various places. So a single `TCHAR` version is not accurate. My template would work whether you call it with `char` or `wchar_t` buffer. Whatever `TCHAR` is compiled to (depending being `ANSI` or `UNICODE`), my template would deduce the correct Windows API to use.

kiyolee:
If `TCHAR` is used consistently and universally everywhere, I wouldn't put in the template in the first place. However, I think `TCHAR` and the associated expected practice of writing generic code that can be compiled and work for **both** `ANSI` and `UNICODE` is old Windows crap. In practice, it is just too difficult. More often than not, developers are more naturally stuck to the `ANSI` way and the `UNICODE` side is likely and simply broken. Say for example, if one uses `sizeof` of a `TCHAR` buffer, there is good chance of being wrong. Additionally, if the software is meant to be ported to other platform, all these `TCHAR` stuff would just become unnecessary burden.

audetto:
> @audetto I can asure you those types used in my PR are all intentional. The problem was `OutputDebugString()` are called with both `TCHAR` and `char` buffers in various places. So a single `TCHAR` version is not accurate. My template would work whether you call it with `char` or `wchar_t` buffer. Whatever `TCHAR` is compiled to (depending being `ANSI` or `UNICODE`), my template would deduce the correct Windows API to use.

I know your PR is intentional, I wonder if the code really needs all these types?

Can we not just forget about all this stuff and use `char` everywhere?
The only place where unicode are actually used, they don't work: https://github.com/AppleWin/AppleWin/pull/917

tomcw:
>Can we not just forget about all this stuff and use `char` everywhere?

This sounds good to me - ie. get rid of `TCHAR`, `wchar_t`.
And this will reduce the complexity of this PR, meaning no templates too.

kiyolee:
Already done that. Please check the latest commits.

kiyolee:
I have been using VS for decades but I still missed to learn that `_vsnprintf()` (VS2013 or before) can indeed return required buffer size but *only* if it is called with buffer being NULL and length 0. An extra call to `_vsnprintf()` is still better than growing buffer I suppose.

tomcw:
@kiyolee - can you fix Log.cpp and Log.h so that the diff doesn't think that every line has changed.
I guess it's "line-endings" related: local is Windows-style, but remote is Unix-style?

IE. here's my setting when I install Git on Windows...

image

tomcw:
Otherwise this PR looks fine to me.
@audetto and @sicklittlemonkey: if you disagree then now is your last chance to comment!

kiyolee:
@tomcw Sure I can fix that. Give me a min.

kiyolee:
I prefer "Checkout as-is, commit as-is" so I think the two files are still stored as dos format (that's what I got upon clone).
Perhaps they haven't been changed for a long time. I thought I could take the opportunity to commit as unix format.

audetto:
I checked the code in linux and 2 things needs addressing

1. `_countof`: the rest of the code just uses `sizeof`. can this be use here too?
2. one cannot use a `va_list` twice, so something like the following is required.

```
  va_list va_copied;
  va_copy(va_copied, va);

  int len = vsnprintf(buf, _countof(buf), format, va);
  if (len < 0)
  {
    va_end(va_copied);
    return std::string(); // Error.
  }
  else if (size_t(len) < _countof(buf))
  {
    va_end(va_copied);
    return std::string(buf, size_t(len)); // No overflow.
  }
  else
  {
    // Overflow, need bigger buffer.
    std::string s(size_t(len) + 1, '\0');
    len = vsnprintf(&(*s.begin()), s.length() + 1, format, va);
    va_end(va_copied);
    return std::string(s.data(), size_t(len));
  }
```

audetto:
I checked it on linux and 2 things needs addressing

1. `_countof`: everywhere else `sizeof` is used. Can we use it here too?
2. a `va_list` can only be used once, so it must be copied for the 2nd time

```
  va_list va_copied;
  va_copy(va_copied, va);
  int len = vsnprintf(buf, _countof(buf), format, va_copied);
  va_end(va_copied);

  if (len < 0)
  {
    return std::string(); // Error.
  }
  else if (size_t(len) < _countof(buf))
  {
    return std::string(buf, size_t(len)); // No overflow.
  }
  else
  {
    // Overflow, need bigger buffer.
    std::string s(size_t(len), '\0');
    len = vsnprintf(&(*s.begin()), s.length() + 1, format, va);
    return s;
  }
```

kiyolee:
@audetto Good catch with `va_copy()`. Didn't aware that before. Let me fix that. `_countof()` is more accurate but unfortunately it is MSVC specific. Given `TCHAR` is no longer a concern though, `_countof()` and `sizeof()` are the same here. I can change that.

tomcw:
@audetto - how is this looking now?
Once I get the ok from you, then I will merge it in.

audetto:
It looks ok.

I will probably submit a patch to include linux changes to `stdafx.h` now.
And a very minor thing, most of the code uses `OutputDebugString` with only 3 occurrences of `OutputDebugStringA` one of which in `Log.cpp` is new.

You can merge I will propose these 2 changes later.

tomcw:
OK, thanks all - now merged into master.

Issue #268, opened 2015-02-17
Replace all sprintf()/wsprintf() calls with _snprintf()


tomcw:
I got a crash in `CmdTraceFile(int)` when the sprintf of the pathname exceeded the buffer of 80 characters.

Ideally all sprintf's should be changed to `_snprintf()`.

There are also calls to `wsprintf()`, which is not wide sprintf, but instead an obsolete Windows API call. Again this should be replaced with `_snprintf()`. NB. [MSDN says](https://msdn.microsoft.com/en-us/library/windows/desktop/ms647550(v=vs.85).aspx) "Do not use."

tomcw:
>Again this should be replaced with `_snprintf()`. 

Why the underscore version of `snprintf()`?

From MSDN [here](https://msdn.microsoft.com/en-us/library/2ts7cx93.aspx), it says:

>The snprintf function and the _snprintf family of functions format and store count or fewer characters in buffer. The snprintf function always stores a terminating null character, truncating the output if necessary. The _snprintf family of functions only appends a terminating null character if the formatted string length is strictly less than count characters.

And:

>Ensure that format is not a user-defined string. Because the _snprintf functions do not guarantee NULL termination—in particular, when the return value is count—make sure that they are followed by code that adds the null terminator.

Better to use the always null-terminating `snprintf()` versions.

tomcw:
In PR #674, _snprintf() got replaced by StringCbPrintf(), from safestr.h. I'll change the title of this issue to reflect this.

tomcw:
From safestr.h:
```
STDAPI
StringCbPrintf(
    __out_bcount(cbDest) LPTSTR  pszDest,
    __in size_t  cbDest,
    __in __format_string LPCTSTR pszFormat,
    ...
    );

Routine Description:

    This routine is a safer version of the C built-in function 'sprintf'.
    The size of the destination buffer (in bytes) is a parameter and
    this function will not write past the end of this buffer and it will
    ALWAYS null terminate the destination buffer (unless it is zero length).

    This function returns a hresult, and not a pointer.  It returns
    S_OK if the string was printed without truncation and null terminated,
    otherwise it will return a failure code. In failure cases it will return
    a truncated version of the ideal result.
```

tomcw:
NB. PR #1041 removed all `wsprintf()` and `StringCbPrintf()` calls.
There are still lots of `sprintf()` calls - mainly in the debugger.

PR #1032, opened 2022-02-06
Replace StringCbPrintf() with StrFormat()

kiyolee:
There is no need to enter buffer length explicitly for `_sntprintf_s()` which is less error prone.
The calls clearly indicate output string will be truncated (if required) instead of buffer overflow.
The calls are made to be consistently UNICODE capable.

### DIFF
--- a/source/Configuration/About.cpp
+++ b/source/Configuration/About.cpp
@@ -65,13 +65,12 @@ static INT_PTR CALLBACK DlgProcAbout(HWND hWnd, UINT message, WPARAM wparam, LPA
 	case WM_INITDIALOG:
 		{
 			HICON hIcon = LoadIcon(GetFrame().g_hInstance, TEXT("APPLEWIN_ICON"));
-			SendDlgItemMessage(hWnd, IDC_APPLEWIN_ICON, STM_SETIMAGE, IMAGE_ICON, (LPARAM)hIcon);
+			SendDlgItemMessage(hWnd, IDC_APPLEWIN_ICON, STM_SETIMAGE, IMAGE_ICON, reinterpret_cast<LPARAM>(hIcon));
 
-			TCHAR szAppleWinVersion[50];
-			StringCbPrintf(szAppleWinVersion, 50, "AppleWin v%s", VERSIONSTRING);
-			SendDlgItemMessage(hWnd, IDC_APPLEWIN_VERSION, WM_SETTEXT, 0, (LPARAM)szAppleWinVersion);
+			std::string strAppleWinVersion = "AppleWin v" + g_VERSIONSTRING;
+			SendDlgItemMessage(hWnd, IDC_APPLEWIN_VERSION, WM_SETTEXT, 0, reinterpret_cast<LPARAM>(strAppleWinVersion.c_str()));
 
-			SendDlgItemMessage(hWnd, IDC_GPL_TEXT, WM_SETTEXT, 0, (LPARAM)g_szGPL);
+			SendDlgItemMessage(hWnd, IDC_GPL_TEXT, WM_SETTEXT, 0, reinterpret_cast<LPARAM>(g_szGPL));
 		}
 		break;
 	}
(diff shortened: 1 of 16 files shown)


tomcw:
I think both `StringCbPrintf()` and `_sntprintf_s()` are Microsoft specific functions.
btw. see #268 where there was some work to replace `sprintf()` with the safer `StringCbPrintf()`.

@audetto - what do you think about this PR? How does it impact the work on your Linux port of AppleWin?

kiyolee:
I think `StringCbPrintf()` is more old school for supplementing older C runtime with Windows API.
`_sntprintf_s()` is more generic, i.e. less Windows centric, and already available at least since VS2008 C runtime.
`StringCbPrintf()` does not have the C++ template support that can derive buffer length automatically.
It also does not support opting out truncation if so needed.
For porting to other platform, it is better to model `_sntprintf_s()` than `StringCbPrintf()`.

kiyolee:
Ok, honestly, I am now confused. MS doc does mention StringCbPrintf() (and other stuff from `strsafe.h`) are intended to replace C/C++ runtime functions. But I am not completely buying that. Losing the buffer length deduction from C++ template just makes calling the class of functions more error-prone. How is that an improvement? And `strsafe.h` seems to be more about targeting C code. The doc does not mention the newer C/C++ safer functions with suffix `_s` (which I use here) and how those functions are relating to `strsafe.h`.

audetto:
It is because the function is templated and get the buffer size directly.

But, I personally dislike all these functions. I would not use any. In my opinion they just encourage sloppy coding.
It is beyond belief that C++ has never provided anything decent to format a string (https://en.cppreference.com/w/cpp/utility/format/format)

My plan (contingent on dropping VS2008, and if `std::format` is not available) was

- functions are not speed critical, so a 2X slowdown + `malloc` is acceptable
- use `snprintf(NULL, 0, ...)` to process it once, get the necessary size, allocate a `std::string`, reprocess it

but

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170

I dont think this feature (C99) is available in VS2008.

`Beginning with the UCRT in Visual Studio 2015 and Windows 10, snprintf is no longer identical to _snprintf. The snprintf function behavior is now C99 standard conformant.`

So, if the PR means I need to add yet another wrapper for sprintf_v6, so be it...

kiyolee:
Indeed, life would be much easier if support for anything before VS2015 is dropped. The feature that `snprintf()` returns the required size in case of overflow leads to a general pattern I use more often and that is to call snprintf() first with a reasonable sized buffer that can cover most of the use cases and only in case of overflow then `snprintf()` is called again with a malloc()'ed buffer. Alternatively, there is `asprintf()` on Linux that you can model from.

audetto:
I checked it on linux and 2 things needs addressing

1. `_countof`: everywhere else `sizeof` is used. Can we use it here too?
2. a `va_list` can only be used once, so it must be copied for the 2nd time

```
  va_list va_copied;
  va_copy(va_copied, va);

  int len = vsnprintf(buf, _countof(buf), format, va);
  if (len < 0)
  {
    va_end(va_copied);
    return std::string(); // Error.
  }
  else if (size_t(len) < _countof(buf))
  {
    va_end(va_copied);
    return std::string(buf, size_t(len)); // No overflow.
  }
  else
  {
    // Overflow, need bigger buffer.
    std::string s(size_t(len) + 1, '\0');
    len = vsnprintf(&(*s.begin()), s.length() + 1, format, va_copied);
    va_end(va_copied);
    return std::string(s.data(), size_t(len));
  }
```

audetto:
Are you going to replace some of these with StrFormat?

kiyolee:
I think I can rework this.

kiyolee:
Life is so much easier with `StrFormat()`. See if this is better.

audetto:
It is.
Soon we can replace the 1st argument of `FrameMessageBox` with a `std::string`.

tomcw:
This look ok from my point of view.
@audetto - sounds like you are ok too?

kiyolee:
Please merge then. I have a PR mostly ready to simplify all the Console print functions in Debugger.

tomcw:
@kiyolee - thanks for this - I've merged to master.

PR #1035, opened 2022-02-12
make DiskGetFullPathName() public...

sh95014:
...because the macOS port needs it for the disk image browser feature. The feature reads the disk image file and displays a directory listing, and should eventually let the user preview the file.

### DIFF
--- a/source/Disk.h
+++ b/source/Disk.h
@@ -139,6 +139,7 @@ public:
 	void FlushCurrentTrack(const int drive);
 
 	const std::string & GetFullDiskFilename(const int drive);
+	const std::string & DiskGetFullPathName(const int drive);
 	const std::string & GetFullName(const int drive);
 	const std::string & GetBaseName(const int drive);
 	void GetFilenameAndPathForSaveState(std::string& filename, std::string& path);
@@ -192,7 +193,6 @@ private:
 	void AllocTrack(const int drive, const UINT minSize=NIBBLES_PER_TRACK);
 	void ReadTrack(const int drive, ULONG uExecutedCycles);
 	void WriteTrack(const int drive);
-	const std::string & DiskGetFullPathName(const int drive);
 	void ResetLogicStateSequencer(void);
 	UINT GetBitCellDelta(const ULONG uExecutedCycles);
 	void UpdateBitStreamPosition(FloppyDisk& floppy, const ULONG bitCellDelta);


tomcw:
Hi @sh95014,

re. macOS port
Can you explain your plan? Is this is collaboration with @audetto (and his Linux port)?
Also where is this work located? (Can you share the repo link)

Thanks.

sh95014:
I actually didn't have much of a plan, but [mysticreddit suggested](https://www.reddit.com/r/apple2/comments/sbq74i/comment/hwhaamd/) that I eventually submit a PR to bring the macOS port back to the main trunk, so this is just a small change to remove one of the conflicts I had. It would be impossible in practice to pull in the macOS port without also pulling in @audetto's Linux port, and I don't know what the plan is on that front either. But I'm open to whatever arrangement you folks prefer, including keeping the macOS port in my fork.

My code is at https://github.com/sh95014/AppleWin

tomcw:
Great to see a native macOS port of AppleWin! :-)

One problem with merging this into the current AppleWin's repo is there's nothing to prevent this change being _accidentally_ reverted in future. What I mean is that this repo is purely for Windows, so someone may submit a PR to revert this since it's not required for Windows and there's other no build targets that require this (in the main AppleWin repo).

Perhaps you (or @audetto) can come up with an idea to prevent this?
EG, perhaps a mock macOS/Linux target that is a dependency for the AppleWin build, which is explicit about calling this function?

tomcw:
>But more broadly, is there actually any interest in merging the Linux/Raspberry Pi port and/or the macOS port back here at all? I can go either way but it does kinda affect how I go about things downstream.

Firstly, there is certainly interest in re-architecting AppleWin such that the OS-specific code (eg. Windows) and frontend is separated out from the core emulation parts. @audetto is doing a great job in tackling this.

In addition there is interest from some in the team (and obviously @audetto too) in a Linux port, but I don't know if there is similar interest for macOS (there isn't from me at the moment).

One problem with having non-Windows platform(s) in the main codebase is maintenance. We'd want a strong commitment that someone would be willing to maintain this other platform for the long term, otherwise the (other platform) code will become stale over time.

For now, it's probably best to continue maintaining your own fork "out-of-tree", then perhaps as the mainline re-arch of AppleWin progresses there will come a good time to merge. But I'd be ok with a mock macOS platform (ie. written in Win32) as a unit-test build dependency, so that any changes you submit as PRs don't get accidentally reverted later (as I commented above).

tomcw:
>The same function for HD is public, even if nothing outside the class needs it.

Yes, true... and I want consistency between the classes `Disk2InterfaceCard` & `HarddiskInterfaceCard`.

>I would vote for making it public and monitor how many more cases turn up, in which case Tom's suggestion sounds a good idea.

OK, let's continue to monitor things.

Michaelangel007:
@sh95014 Thanks for this PR.  I'm mysticreddit over on reddit.
 
 @tomcw Thanks for taking point on this.
 
 @audetto Thanks for your work on cleaning up the codebase to be more platform neutral.

PR #1036, opened 2022-02-12
SY6522: initialise card.

audetto:
and avoid usage of uninitialised values.

### DIFF
--- a/source/6522.h
+++ b/source/6522.h
@@ -7,6 +7,7 @@ public:
 	{
 		for (UINT i = 0; i < kNumTimersPer6522; i++)
 			m_syncEvent[i] = NULL;
+		Reset(true);
 	}
 
 	~SY6522(void)


tomcw:
Thanks! (merged)

audetto:
Hi I did not realise the fact that this code is called from

https://github.com/AppleWin/AppleWin/blob/83e56924f735a1bfc6b443a7f2861924ec94f921/source/Mockingboard.cpp#L144

and this needs to initialise the critical section.

otherwise, I would have just initialised the memory and left the rest unchanged.
shall I split `Reset` in 2 instead? and call the state only in the constructor?

audetto:
BTW, in linux I completely ignore critical sections. What are they used for?

Other than the serial card, is there any usage?

tomcw:
@audetto - no need to do anything, as I fixed the SY6522.Reset() function to explicity call the CriticalSection create function.

>BTW, in linux I completely ignore critical sections. What are they used for?

A few things, but in this case to ensure consistency when set/clearing IRQ sources in: `g_bmIRQ` from multiple threads.

>Other than the serial card, is there any usage?

Only the serial card uses a 2nd thread to set/clear the 6502 IRQ.
Every other hardware device set/clears IRQs from the main thread, ie. 6522, SSI263 (for speech), Mouse.

audetto:
Understood. Thanks

PR #1039, opened 2022-02-16
Remove unnecessary static std::string variables

kiyolee:
The static variables are kind of weird that got recomputed upon every call.
The functions can simply return the wanted string as value.


tomcw:
Hmm... I got a handful of build errors, which I fixed in 43b9df2.

PR #1041, opened 2022-02-18
Replace sprintf_s() and wsprintf() with StrFormat()

kiyolee:
This PR deals with all the `sprintf_s()` and `wsprintf()` calls and replaces with equivalent `StrFormat()`.
This leads to some possible improvements to SerialComms and `GetSymbol()`/`FormatAddress()` which I separate into individual commits to help review. (I can do if these deserve some separate PRs)
The remaining bigger lot of `sprintf()` will be deal with in another PR.

### DIFF
--- a/source/Debugger/Debug.cpp
+++ b/source/Debugger/Debug.cpp
@@ -2270,9 +2270,8 @@ void _CmdColorGet( const int iScheme, const int iColor )
 	}
 	else
 	{
-		TCHAR sText[ CONSOLE_WIDTH ];
-		wsprintf( sText, "Color: %d\nOut of range!", iColor );
-		GetFrame().FrameMessageBox(sText, TEXT("ERROR"), MB_OK );
+		std::string strText = StrFormat( "Color: %d\nOut of range!", iColor );
+		GetFrame().FrameMessageBox(strText.c_str(), "ERROR", MB_OK);
 	}
 }
 
@@ -5774,7 +5773,6 @@ Update_t CmdOutputCalc (int nArgs)
 		return Help_Arg_1( CMD_OUTPUT_CALC );
 
 	WORD nAddress = g_aArgs[1].nValue;
-	TCHAR sText [ CONSOLE_WIDTH ];
 
 	bool bHi = false;
 	bool bLo = false;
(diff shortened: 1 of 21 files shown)


source/Debugger/Debug.cpp
```diff
@@ -8420,33 +8411,33 @@ void DebugContinueStepping(const bool bCallerWillUpdateDisplay/*=false*/)
 
 		if (regs.pc == g_nDebugStepUntil || g_bDebugBreakpointHit)
 		{
-			char szStopMessage[CONSOLE_WIDTH];
-			const char* pszStopReason = szStopMessage;
+			std::string strStopMessage;
```
tomcw, line 8414: Minor thing: you can get rid of `strStopMessage`

tomcw:
I've looked in some detail at the non-debugger changes, and they appear OK.
I've skimmed the debugger changes - and again they appear OK.

If I were doing a thorough review then I should first checkout @kiyolee's branch, and check all code paths... but the changes are many and pervasive!

I think if there are some subtle issues, then they can just be fixed in the main branch (ie. post merge).
And it will be worth doing a Coverity static analysis scan, and see what new defects that reports.

So I will merge to master (then do the Coverity scan).

tomcw:
NB. Coverity didn't report any new defects with any of the changes in this PR.

tomcw:
btw, @kiyolee - thanks for your work on all of these string-formatting related changes.

Michaelangel007 (commented):
Looks fine.

source/Debugger/Debugger_Help.cpp
```diff
@@ -735,96 +733,78 @@ Update_t CmdHelpSpecific (int nArgs)
 //		if (nFound && (! bAllCommands) && (! bCategory))
 		if (nFound && (! bAllCommands) && bDisplayCategory)
 		{
-			char sCategory[ CONSOLE_WIDTH ];
+			const char* pszCategory = "";
```
Michaelangel007, line 736: Please rename `pszCategory` to `psCategory`.  The `z` indicator is redundant.  All we need to know is that:

* This is a pointer
* This is a string

Michaelangel007 (commented):
Other then the one var name, looks fine.

source/Debugger/Debugger_Help.cpp
kiyolee: https://github.com/AppleWin/AppleWin/pull/1057
Submitted the PR to fix this.

Michaelangel007:
+1 Ship it!

PR #1047, opened 2022-02-26
Add Uthernet II and make Uthernet 1 a Card

audetto:
This PR adds support for the Uthernet II card to AppleWin, and at the same time converts the TFE to a Uthernet 1 card.

The Uthernet 1 code has been moved, but it is otherwise unchanged (*).

The AW settings tab has been extended and a few corner cases where settings were lost have been fixed.

Uthernet II emulation: not every single feature is currently supported as I could not find examples to test

- IPRAW: not implemented (see https://github.com/A2osX/A2osX/issues/163)
- PPPoE (no idea what it does)
- some server side cases of TCP and UDP mode
- save state (currently it only saves the *presence* of the card)

Tested on:

- https://github.com/oliverschmidt/contiki
- https://github.com/cc65/ip65
- https://github.com/A2osX/A2osX
- https://github.com/KrisKennaway/ii-vision

(*) the PCAP code has been refactored as a `NetworkBackend` in order to ease alternative implementations (libslirp in Linux)

Still TODO

- full save state (to the degree where this is possible)
- further clean up (i.e. deletion) of TFE/PCAP code (all the string manipulations can go)
- `Card` interface really needs to include `Destroy` and `Reset` as virtual methods
- SerialCard and Uthernet have similar semantics: open a `device`, but their behaviour is different (former does not require a reboot, latter does) and they should all handle the interface name in the `Card` constructor

But I think this is a good milestone and worth merging.


tomcw:
I've not looked at this yet - but it sounds exciting!
btw, @bobbimanners' [Emai//er](https://github.com/bobbimanners/emailler) also supports Uthernet II for the //e.

tomcw:
I cloned your branch, and got both U1 and U2 working using Contiki:
- IPCONFIG: Request IP address
- WEBBROWS: Browse INFO.CERN.CH

So great work! :-)

I know save-state isn't fully functional yet, but I created a save-state file at the "Request IP address", but on reloading then when I press RETURN (ie. to request the ip address), then it just sits there saying "Requesting...". If I press TAB and select "Cancel" then re-run IPCONFIG, then I can successfully request an IP addr. Anyway, not a big deal at the moment, as I'm sure it's just that the W5100 state isn't being reset when I load a save-state.

A few requests:
- can you rename "Backend.cpp/h" and to "NetworkBackend.cpp/h", ie. to keep the filename consistent with the class name.
- for the renamed TFE.cpp to Uthernet1.cpp, can you restore the original `tfe.c` GPL comment header (eg. "tfe.c - TFE ("The final ethernet") emulation", "Written by Spiro..." etc.
  - if you want: add a _second_ comment block after this to describe your changes.
  - eg. see VidHD.cpp - you could say:
    - "Emulate an Uthernet1 card (adapted from VICE's TFE support)"
- for the 3 new .cpp files (Backend.cpp, PCapBackend.cpp and Uthernet2.cpp) can you add the "boilerplate" AppleWin GPL comment header.
  - eg. see Card.cpp - just change the names (Tom, Michael) for yours here and use: "Copyright (C) 2022"

audetto:
Save state not done yet.

tomcw:
One mistake... can you move the GPL comment header from Uthernet1.h to Uthernet1.cpp?

Otherwise this all look good to merge.
Don't worry about save-state just yet. That can afterwards.

Thanks.

audetto:
> I've not looked at this yet - but it sounds exciting! btw, @bobbimanners' [Emai//er](https://github.com/bobbimanners/emailler) also supports Uthernet II for the //e.

Very comprehensive suite, but so far failed to configure it properly.
On the good side, it uses IP65 which I tested for other clients, but I will continue to configure it to ensure correctness.

tomcw:
Great @audetto! Really good to get this Uthernet II card finally supported!
(Now merged to master)

tomcw:
I forgot to try building the old VS2008 solution, and there are a few unsupported C++11 things, which would need fixing if there is really a need to continue to keep the VS2008 support going:
- std::shared_ptr is C++11, so would need to be changed to std::tr1::shared_ptr
- Uthernet2.cpp does a `#include <cstdint>` - just remove this (not needed for Windows)
- Uthernet2.cpp uses C++11 range-based for loops, ie. `for (Socket &socket : mySockets)` which would need to be replaced with this ugliness:
  - `for (std::vector<Socket>::iterator it = mySockets.begin(); it != mySockets.end(); ++it)`
- Uthernet2.cpp uses C++11 `std::vector<Ty>::data()`, which would need to be replaced with this:
  - &myMemory[0], etc
- Uthernet2.cpp uses C++11 `nullptr` - replace with `NULL`

I will raise a new issue and get opinions from other project members, but I've been releasing with VS2019_v141_xp for nearly a year now, so probably time to retire VS2008.

audetto:
ops....

PR #1048, opened 2022-02-26
Replace sprintf() with StrFormat()

kiyolee:
After this, the remaining sprintf() usages *should* all be within the debugger area.


tomcw:
Can you remove the tfe.cpp changes (but tfearch.cpp is fine to remain), as this will clash with some file renaming (and refactoring) that @audetto is doing in another PR.

After that other PR lands, then you can apply this final patch to the newly named file.

Or perhaps just wait. As if a "git mv" was done on tfe.cpp, then the merge maybe able to resolve it successfully.

kiyolee:
Changes for `tfe.cpp` has been removed.

kiyolee:
Updated the PR to incorporate latest changes. I have put back the original changes for `tfe.cpp` (now in `Uthernet1.cpp`). Please check.

tomcw:
Thanks - merged it now.

PR #1049, opened 2022-02-27
Various *Name() functions can simply return const std::string&

kiyolee:
This saves a bunch of unnecessary string copying.

### DIFF
--- a/source/CPU.cpp
+++ b/source/CPU.cpp
@@ -797,7 +797,7 @@ void CpuNmiDeassert(eIRQSRC Device)
 #define SS_YAML_VALUE_6502 "6502"
 #define SS_YAML_VALUE_65C02 "65C02"
 
-static std::string CpuGetSnapshotStructName(void)
+static const std::string& CpuGetSnapshotStructName(void)
 {
 	static const std::string name("CPU");
 	return name;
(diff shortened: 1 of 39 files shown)


tomcw:
This all looks fine, but a minor point, can you add back in the `void` keywords to all the functions' definitions/declarations.
EG. `static const std::string& CpuGetSnapshotStructName(`**`void`**`)`

This will retain the style used by those files.
Admittedly this style isn't consistent within all .cpp files, but whenever I refactor functions or .cpp files, that's what I've been adding.

kiyolee:
I am fine to put those back in. But I think without `void` is more modern C++ style. In C++, `()` strictly means `(void)` whereas in C, `()` is equivalent to `(...)`.

tomcw:
Thanks - yes, put "void" back for now.

We can review the style of using void (or not) another time, involving the other project members.
If it's decided to drop "void" then it can be done wholesale, and documented in the "CodingConventions.txt" file.

kiyolee:
`void` argument lists have been restored.

tomcw:
@kiyolee - after a recent merge, tfe.cpp/h got renamed to Uthernet1.cpp/h.
So can you modify this PR to target those files instead?
Also Uthernet2.cpp/h can be similarly targeted.
Thanks.

kiyolee:
Updated the PR to incorporate latest changes. Please check.

tomcw:
Merged into master - thanks.

PR #1051, opened 2022-02-28
Fix VS2008 build

kiyolee:
I guess this may raise the question again if VS2008 support is still worthy.
If yes, check this.


kiyolee:
Assuming VS2008 support is to be dropped. Close this for now.

PR #1053, opened 2022-03-05
Use clear() instread of erase(begin(), end())

kiyolee:
Just cleaner code using `X.clear()` instead of `X.erase(X.begin(), X.end())`.


tomcw:
Thanks - merged it in.

PR #1054, opened 2022-03-05
Uthernet II: implement save state.

audetto:
This seems to work, but I need some more testing.

### DIFF
--- a/source/SaveState.cpp
+++ b/source/SaveState.cpp
@@ -465,6 +465,7 @@ void Snapshot_LoadState()
 		return;
 	}
 
+	LogFileOutput("Loading Save-State from %s\n", g_strSaveStatePathname.c_str());
 	Snapshot_LoadState_v2();
 }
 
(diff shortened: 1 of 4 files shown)


tomcw:
Re. this FIXME comment in `LoadSnapshot()`
```
    yamlLoadHelper.LoadBool(SS_YAML_KEY_ENABLED); // FIXME: what is the point of this?
```

I think this is probably legacy from the old TFE (ie. Uthernet1) code where the implementation wasn't a `Card`, so the static "TFE" code needed a way to know if it was enabled or not.

So I think you can remove it from the **Uthernet2** code for `LoadSnapshot()` and `SaveSnapshot()`.

NB. We can't just remove it from the **Uthernet1** code, since there it has been released in AppleWin versions. If you decide to remove this, then I can make the change for Uthernet1 code (ie. I'll bump the version and support both cases: with & without SS_YAML_KEY_ENABLED)... or feel free to make this change yourself.

tomcw:
Thanks - merged to master.

Issue #1050, opened 2022-02-27
Deprecate support for building with VS2008


tomcw:
Since April 2021 (10 months ago) I've been building releases with VS2019 v141_xp, which has been fine.

@Michaelangel007 , @sicklittlemonkey , @audetto:
Is it time to retire VS2008 support?

Nick: in the past you've said: "I presume that means it will be ok to use modern C++ features."
Basically C++11 things like range-based for loops, std::shared_ptr, `nullptr` etc. seem fine.
But perhaps some initial limits on _gratuitous_ use of `auto` and lambdas?

EG, this PR seems to take C++11 Initializer Lists too far:
https://github.com/AppleWin/AppleWin/pull/634/files

Can someone suggest some sensible limits?

sicklittlemonkey:
Whether it would be good to be able to use modern C++ is probably a matter of taste. ; - )
Since these days I don't get to use C++ at work much (or contribute to AppleWin!) I'm a bit out of touch.
I would be interested in the opinions of the more active recent contributors.

There are some established style guides we could adopt or copy from. E.g.
https://google.github.io/styleguide/cppguide.html

This one is more philosophical about being dragged kicking and screaming into the future.
https://isocpp.github.io/CppCoreGuidelines
This latter one appeals to me in terms of modernizing the codebase.
It might line up well with all the great refactoring you guys have been doing recently.

audetto:
To me the reasons to move away from 2008 are

- hard to test. not easy to find the compiler any longer.
- allow to use a few good improvements to C++ (some of which are in https://github.com/AppleWin/AppleWin/pull/1051)

The good pattern for me include

1. unique_ptr
2. range based loop
3. static_assert
4. vector initialiser list
5. default value for member variables

kiyolee:
I do this all for fun and deprecating VS2008 or not imposes different kinds of challenge and fun. However, as a software project moving forward, deprecating VS2008 makes more sense. With latest C++ features, a lot of code can be modernised and simplified. Ideally this should also reduce the barrier for new contributors. I would even suggest to make C++14 (default for VS2019) the minimum requirement.

tomcw:
btw, the v141 toolset is used to build the AppleWin release.

This VC++ v14.1 is from Visual Studio 2017 (v15.0)... but is still shipped with VS2019.

It has some level of support for C++11/14/17 (but not complete) - I can't find a good description of what is / isn't supported.
[...]

PR #1055, opened 2022-03-06
Fix warnings: forcing integer or pointer to bool

kiyolee:


tomcw:
I guess this just relates to VS2008?
It's very likely we'll deprecate the VS2008 support (#1050) - and if that's the decision then this PR isn't required.

kiyolee:
Ah, just double checked, the kind of warnings no longer happens since VS2017 though.
Right, assuming VS2017 would become the minimum requirement, this is not required.

kiyolee:
Assumming VS2008 support is to be dropped. Close this for now.

Issue #868, opened 2020-11-12
Update CodingConventions.txt


tomcw:
Hi @Michaelangel007,

[CodingConventions.txt](https://github.com/AppleWin/AppleWin/blob/master/docs/CodingConventions.txt) could do with an update!

My AppleWin style has deviated from this old Hungarian-heavy convention, and is more "lower camel case" now, generally without any Hungarian prefixes. See this [recent cpp file](https://github.com/AppleWin/AppleWin/blob/master/source/SynchronousEventManager.cpp), where I just use prefixes 'p' (pointer) and 'm_' (member) and not much else!

Obviously the whole codebase is very mature with a range of styles/conventions accrued over the years. I'm not suggesting we change any of the existing code. Just that we update this CodingConventions.txt document to reflect the current convention.

I'm proposing changing it to just be:
```
Prefixes:
g_	: global
m_	: member
p	: pointer
```

And then move everything else to the `Legacy` sub-section.

Also remove the part about "Don't go Hungarian-mad..."

We should also add a note to say to adopt the coding convention of any module (or function) you are modifying.

What do you think?

Michaelangel007:
@tomcw Hey Tom, sorry for the late reply.

If you want to remove the underscore `_` from the prefixes that is perfectly fine.

In my day-to-day graphics / shader job I have a more personal stringent adherence. For example since I deal with math types all the time I usually also have these prefixes:

```
m matrix
v vector
q quaternion
t non-POD
```

Now since AppleWin is NOT math heavy I originally left them out as the Coding Convention back then as the proposed standard was already a bit of a "burden".

Personally, after ~14 years since joining AppleWin, I'm now an even bigger believer in keeping the mini-Hungarian notation "as-is." One of the things I see polluting modern "Crap++" is the abuse of `auto` where the reader can no longer _eassily_ tell the types -- having the burden pushed onto the compiler by the writer -- the reader now has to waste even more time decoding types. I strongly believe a _minimal_ character prefix helps the reader. Either extreme, none, or the other -- the insane verbosity of "pure" Hungarian -- is a hindrance.

Here is the updated listed without the underscores:

```
a	: array
b	: bool
e	: enum variable
g 	: global
h	: handle
i	: iterator (eg. UINT, STL-iterator)
m	: STL map
m	: member
n	: int
p	: pointer
r	: reference
s	: string
sgp	: singleton
u	: unsigned int
v	: STL vector

Tags:
_e	: named enum definitions
[...]

PR #1057, opened 2022-03-06
Rename 'pszCategory' to 'pCategory'

kiyolee:


tomcw:
In Debugger_Help.cpp, the style for naming vars of type `const char*` is either of the style `pSrc` or `format` (or this one case of `pszCategory`).

So even though the CodingConventions.txt talks about prefixes `p` and `s`, I don't think prefix `ps` is in keeping with the style of Debugger_Help.cpp file. IMO, better to rename to `pCategory`.

btw, I did a search for "psz" across the whole AppleWin project and found 164 instances (cross 25 files).

kiyolee:
In general I personally don't like variable prefix (except `p` to mean pointer) or even hate Hungarian notation.
I thought `psz` was kind of the de facto for Windows and assumed it was used here as we are dealing with Windows software (at least still for now).
So I would :+1: `pCategory` too.
If you are ok, I can expand the PR to replace all `psz` to simply `p.

tomcw:
I'm no fan of Hungarian these days either, and when I write an entirely new .cpp file (eg. LanguageCard.cpp, VidHD.cpp), then there's just the `m_` prefix (+ a small number of `p` prefixes).

But the codebase is mature, and has had many different authors write various parts in their own styles. Hence the point in the code conventions stating:

>As a general rule and for consistency, adopt the coding convention/style of any module (or function) you are modifying.

Refactoring the whole codebase to update to my current style today would be a huge change. If we were to go this route, then we'd need consensus from other project members and agree a style too... the later surely being difficult and a degree of "bike shedding" too.

From my POV, I prefer to spend my limited available time focussing on architectural improvements.

tomcw:
>If you are ok, I can expand the PR to replace all `psz` to simply `p`.

I think best to leave the code as-is for now.
My preference would be no prefix at all for these strings, but since this `psz` is so pervasive, then doing that would almost definitely go against this coding conventions rule:

>As a general rule and for consistency, adopt the coding convention/style of any module (or function) you are modifying.

IE. the newly refactored variable names wouldn't be correct relative to the style of that .cpp file.

kiyolee:
Then please merge this.
And I have a question regarding the upcoming big lot of replacing `sprintf()` in debugger.
I have added a fair amount of new `std::string` variables, I have currently added the prefix `str` there.
Is the prefix `str` preferrable then?
Again, personally I don't use that. Using simple prefix `s` seems confusing with `const char*`.

tomcw:
>And I have a question regarding the upcoming big lot of replacing `sprintf()` in debugger.
I have added a fair amount of new `std::string` variables, I have currently added the prefix `str` there.
Is the prefix `str` preferrable then?

I searched the whole project for "std::string", and just looking at the Debugger related files, there's a mixed bag:
- std::string g_aSourceFileName;
- std::string g_sFileNameConfig
- std::string strAddressBuf;
- std::string const* pTarget = NULL;
- std::string stopReason = "Unknown!";

(I think this last one is mine, and that's my preference, as can be seen in the non-debugger .cpp code.)

I guess some of these came about when previous refactors changed `char*` into `std::string`, but the variable name wasn't changed.

Let me defer to @Michaelangel007 to comment on this, since it's all (mainly) his code.

tomcw:
>Then please merge this.

AFAICT in the Debugger code, there's no `ps` prefix for `char*` variables - just `p`, `s` or no prefix.
So for this specific PR, again let's see what @Michaelangel007 prefers.

tomcw:
Just re-reviewing this old(ish) PR.
Above I said:
>IMO, better to rename to `pCategory`.

I'll accept this PR if you change to that.

tomcw:
OK, merged it.

Michaelangel007:
> I guess some of these came about when previous refactors changed `char*` into `std::string`, but the variable name wasn't changed.

Sorry for the late reply.

Tom, YES, you are correct for why we have quite a mixed back of naming.  The underlying types have changed but not the variables names.  (This isn't a design flaw of my pseudo-Hungarian notation, it is a feature.)

i.e.
`std::string g_aSourceFileName` is obviously wrong as the `a` stands for `array`.  The proper name (& type) should be either:

* char g_aSourceFileName[];

or

* std::string g_sSourceFileName;

Since we're abandoning old-style C's char and replacing it new-style C++ std::string the `s` prefix is the correct one.

I probably should make a variable name cleanup pass to see what is outstanding.

Michaelangel007:
@Tom Here are some background on why I use my minimal Hungarian notation prefixes:

I never use `psz` as that is **bloated** Hungarian notation.  I prefer a much more sane **minimal** Hungarian notation.  (Well as sane as Hungarian can be.) 

/Oblg french accent:_There is ze method to ze madness!_ 

The bloated `psz` is hideous because:

* the `z` is just noise.  I don't _need_ to know the internals of a string being zero-terminated in a "high" level language.
*  The `s` can also be redundant -- it depends on if I have other types within scope. If not, then I drop it.  If I do then I keep it.
* The only part we really care about is the **pointer** notation so I insist on the `p`.  Some places use a trailing `p` -- which drives me insane as it is too easy to miss.  I hate LISP-style naming.

More details on how other prefixes pop up:

* `s` is a string.  See note above.
* Occasionally I will prefix variables with a `t` which stands for `the`.  This is rare but when I have more then one string type and I want to maintain column naming alignment. i.e. `tCell` `nCell, and `iCell`.
* `ps` is a pointer to string.  I tend to be lazy and dropping the `s`, which is fine, since the most important bit is knowing it is a **pointer** as explained above.
* `rs`  is reference to string.  Again dropping the `s` is fine, as knowing it is a **reference** is the most important bit. i.e. `rLHS` vs `rRHS`
* For member variables I don't use the `m_`.  The `m` is redundant and just adds visual clutter.  I've been slowly migrating away from even included the leading underscore the same way I've been dropping s for string.  `FontConfig_t` has it while `DisasmLine_t` does not.
* `g_` is a global.  I could have used `g` but two characters is a "discreate "Does this NEED to be a global" without being in your face about it.  A single char `g` is too easy to miss.

You can see an example in `ArgsCook` where I use `i`, `n`, and `p` for Iterator, Total, and Pointer respectively.

```
	int    nArg = nArgs;
	int    iArg = 1;
	Arg_t *pArg = NULL; 
```

Another good example is `ClearArg()`

```
	pArg->sArg[0] = 0;
	pArg->nArgLen = 0;
	pArg->bSymbol = false;
	pArg->eDevice = NUM_DEVICES; // none
	pArg->eToken  = NO_TOKEN   ; // none
	pArg->bType   = TYPE_STRING;
	pArg->nValue  = 0;
```

I feel a single character is a good compromise between minimizing verbose naming and communicating types.

Hope this helps explains why I invented and stick to my minimal Hungarian notation.

tomcw:
Thanks Michael, I think what you say above is consistent with previous discussions, eg. #868, where we concluded:

>* _For consistency, adopt the coding convention of any module (or function) you are modifying._
>
>So unless someone refactors the entire C++ debugger file, then they should be sticking to your style in the debugger.

Michaelangel007:
Yes.

I should probably add a debugger coding convention since we are having more people cleanup inconsistent formatting.

tomcw:
Yes, probably a good idea, so that we can point at that to explain to two styles/conventions in the code base.

PR #1060, opened 2022-03-12
Debugger: replace sprintf() part 1

kiyolee:
- Add MemoryTextFile_t::PushLineFormat()
- Replace some sprintf() with PushLineFormat()


tomcw:
OK, I've merged this.

btw, I'm sure you spotted it, but `CmdBookmarkLoad()` and `CmdBreakpointLoad()` are not accessible (ie. commands `BMLOAD` and `BPLOAD` are commented out), and also they are not actually implemented either (the functions are empty!).

PR #1064, opened 2022-03-14
Add ByteToHexStr() and WordToHexStr()

kiyolee:
- Simplify common StrFormat(), especially in Debugger (changes upcoming)
- Add helpers StrAppendByteAsHex() and StrAppendWordAsHex()
- Add helpers StrBufferAppendByteAsHex() and StrBufferAppendWordAsHex() for plain string buffer

### DIFF
--- a/source/LanguageCard.cpp
+++ b/source/LanguageCard.cpp
@@ -443,7 +443,7 @@ bool Saturn128K::LoadSnapshot(YamlLoadHelper& yamlLoadHelper, UINT version)
 		}
 
 		// "Memory Bankxx"
-		std::string memName = GetSnapshotMemStructName() + StrFormat("%02X", uBank);
+		std::string memName = GetSnapshotMemStructName() + ByteToHexStr(uBank);
 
 		if (!yamlLoadHelper.GetSubMap(memName))
 			throw std::runtime_error("Memory: Missing map name: " + memName);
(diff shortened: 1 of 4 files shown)


audetto:
I see this going to include `StrAppendDWordAsHex` and `StrAppendWord64AsHex`,

maybe it is one of the cases where templates help...

kiyolee:
Well, I don't see anywhere here printing 32-bit or 64-bit quantities and I do doubt the need of it.
Using pointer may generalize the algorithm, but usually for small quantities direct char array lookup is more inline friendly.
Some modern compilers may be already smart enough to do loop unrolling these days though.
Pointer also introduces the requirement to deal with Endianness differences. That is tedious.
However, assumming 8-bit and 16-bit quantities are main concern here, my code here is probably good enough.

audetto:
I was referring to this

https://github.com/AppleWin/AppleWin/blob/master/source/YamlHelper.h#L221-L227

kiyolee:
I see.
But for numeric presentation of values, it is still easier to avoid Endianness consideration.
Btw I do prefer using `Int(8|16|32|64)` and `UInt(8|16|32|64)` (`UInt` is more common than `Uint` (lowercase `i`) in Windows APIs) as in your code reference instead of `Byte`, `Word`, `DWord`, and `QWord` (definitely not `Word64`).
Just as I prefer `uint(8|16|32|64)_t` in C++ code instead of `BYTE`, `WORD`, `DWORD`, `QWORD` which are kind of Windows specific.
This is probably another topic split people half and half.

tomcw:
This looks fine to me.

One thing: I don't think these new functions need to be declared `inline` - it's not like these string functions are speed critical. But I'm okay to stick with how you've done it.

tomcw:
Thanks - merged to master.

Issue #518, opened 2017-12-16
Support Uthernet II card


tomcw:
Uthernet II uses the WIZnet W5100 for hardware TCP/IP stack (the original Uthernet used the cs8900a ethernet controller) . (Datasheet [here](https://www.sparkfun.com/datasheets/DevTools/Arduino/W5100_Datasheet_v1_1_6.pdf).)

http://a2retrosystems.com/products.htm:
>Apple II Ethernet Module - This product is based on the Wiznet W5100 chip. It has an embedded hardware IP stack. Contiki, IP65, Marinetti and ADTPro have been updated to work with the new card.

So far no requests for supporting this card, but there's this comment on http://a2retrosystems.com:
>You may also want to go and grab a copy of the latest Applewin 
 and check out the Uthernet I emulation.
>
>There is no Uthernet II emulation :(

Anyway, I'm creating this issue as a placeholder for potential future work.

patrickbst:
I will second the desire for U2 emulation, since that is the currently available card and what many might have in their real apples.

tomcw:
Streaming demo using Uthernet II:
https://github.com/KrisKennaway/ii-vision


And posting on csa2, [Streaming video](https://groups.google.com/forum/#!topic/comp.sys.apple2/higTXSzAVIA):

>Something caught my eye when I was reading the manual - the W5100 has 16K of onboard memory for TCP send/receive buffers, and the Apple II interface is via an I/O address that auto-increments the address pointer on the W5100.

>This means that iterating through the received TCP data is very cheap on the Apple side if all you care about is reading in sequence.

>Actually this is the fastest possible way for the 6502 to read data (without DMA) since we’re just reading a single memory address repeatedly, and we can read up to 8k at a time this way before we have to deal with socket buffer management.

audetto:
Hi
missed this issue.
Anyway, I wrote an emulation for the Uthernet II card which reuses all the tfe-pcap code as Uthernet I.

The card supports 4 type of sockets: MACRAW, IPRAW, UDP and TCP.

MACRAW works almost the same as Uthernet I, and one sends Ethernet II frames via libpcap.

The other types are more complicated to emulate, but would be easier to use for the 6502 as they abstract away part of the low level network mechanics.

Currently

- Contiki uses MACRAW
[...]

PR #1069, opened 2022-03-20
Uthernet II: implement IPRAW sockets.

audetto:
I think the emulation for IPRAW sockets is complete (except potentially some errata).

Missing are some more extensive test cases, the only existing one is A2OSX.
But even this will make it easier to test it and validate further.

### DIFF
--- a/AppleWinExpress2019.vcxproj
+++ b/AppleWinExpress2019.vcxproj
@@ -114,6 +114,7 @@
     <ClInclude Include="source\StrFormat.h" />
     <ClInclude Include="source\SynchronousEventManager.h" />
     <ClInclude Include="source\Tape.h" />
+    <ClInclude Include="source\Tfe\IPRaw.h" />
     <ClInclude Include="source\Tfe\NetworkBackend.h" />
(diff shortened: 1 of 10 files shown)


tomcw:
Hi Andrea - this sounds really good!

I can see from #518 that A2osX doesn't really support IPRAW mode.
So how are you testing this?

Also what is this new library `iphlpapi.lib` ?

audetto:
The last TEST from https://github.com/A2osX/A2osX/tree/master/.Floppies partially supports it. There are currently both hardware and software issues.

But, I can generate enough packets targeting the local network and the internet to write the emulation and run some of it, until the (A2osX) software issues kicks in. I see the packets send and received correctly.

This is why I say 90%. I think the support is good enough to be used and tested at a wider audience which will eventually lead to a better emulation too.

I need to use `SendARP` https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-sendarp which has been available since Windows 2000 (in `iphlpapi.dll`)

tomcw:
Thanks Andrea - I've merged it in.

PR #1072, opened 2022-03-25
Debugger: Format changes only - consistent space after keywords

kiyolee:
Consistently adding space after keywords `if`, `for`, `switch` and `while`.
I have counted, there are more statements having the space than not.
This is only for the Debugger area. Inconsistency still exists in other areas.

### DIFF
--- a/source/Debugger/Debug.cpp
+++ b/source/Debugger/Debug.cpp
@@ -793,7 +793,7 @@ Update_t CmdProfile (int nArgs)
 				int   nLine = g_nProfileLine;
 				int   iLine;
 				
-				for( iLine = 0; iLine < nLine; iLine++ )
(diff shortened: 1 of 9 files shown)


tomcw:
OK, merged this one.

btw, I'm doing a heavy refactor to Mockingboard.cpp (so that it behaves like a card), and I'm applying a similar "space after keywords" tidy-up to that file. So if you are tempted to submit a similar consistency pull-request for **non-debugger files**, then best to exclude Mockingboard.cpp (so that I don't get a merge conflict when I come to submit my changes).

kiyolee:
No worries. I am still trying to completely removing `sprintf()` in debugger. These format changes just get into the way.

Michaelangel007:
Personally I'm not a big fan of spaces after keywords such as `if` but we probably should add it to the Coding Standard.

PR #1073, opened 2022-03-25
Fix to use logical "&&" for boolean variables

kiyolee:
Since both variables are boolean, VC detected and suggested that the intention should be using logical AND `&&` instead of bitwise AND `&`.


tomcw:
OK, merged it.

PR #1076, opened 2022-03-27
Debugger: Format changes only

kiyolee:
- Fix some inconsistent indentations
- Add some curly brackets to clarify structure

PR #1081, opened 2022-04-02
Debugger 2.9.1.14

Michaelangel007:
Small patch to make the debugger's markup of hex bytes in Applesoft's FAC format be nicer.

Also added preliminary support for A2_BASIC.A2D to markup all the Floating-point Constants in Applesoft BASIC.

Waiting on #1076 before accepting since there will be merge conflicts.


tomcw:
btw, @Michaelangel007, #1076 has been merged.

PR #1082, opened 2022-04-04
Debugger: replace prefix 'str' with 's'

kiyolee:
While the memory is still fresh, let me clean up the wrong prefix I put in through some earlier PRs.


Michaelangel007 (approved):
Looks good!

tomcw:
Merged.

PR #1083, opened 2022-04-04
Debugger: Remove unnecessary alias pointer

kiyolee:
pBmp is never changed and always points to bmp.
I think the code can simple use bmp directly.
Also initialize the bmp structure properly.


Michaelangel007 (approved):
Looks good.

tomcw:
Merged.

PR #1090, opened 2022-04-17
Debugger: replace sprintf() part 4 - Debugger_Display

kiyolee:
- Change `ColorizeSpecialChar()` to return `std::string`


source/Debugger/Debugger_Display.cpp
```diff
@@ -945,15 +940,15 @@ void SetupColorsHiLoBits ( bool bHighBit, bool bCtrlBit,
 
 // To flush out color bugs... swap: iAsciBackground & iHighBackground
 //===========================================================================
-char ColorizeSpecialChar( char * sText, BYTE nData, const MemoryView_e iView,
-		const int iAsciBackground /*= 0           */, const int iTextForeground /*= FG_DISASM_CHAR */,
-		const int iHighBackground /*= BG_INFO_CHAR*/, const int iHighForeground /*= FG_INFO_CHAR_HI*/,
-		const int iCtrlBackground /*= BG_INFO_CHAR*/, const int iCtrlForeground /*= FG_INFO_CHAR_LO*/ )
+static std::string ColorizeSpecialChar( BYTE nData, const MemoryView_e iView,
+	const int iTextBackground = BG_INFO, const int iTextForeground = FG_DISASM_CHAR,
```
Michaelangel007, line 944: Please fix line 944 so the `const iTextForeground` is column aligned.
i.e.
```
	const int iTextBackground = BG_INFO     , const int iTextForeground = FG_DISASM_CHAR,
	const int iHighBackground = BG_INFO_CHAR, const int iHighForeground = FG_INFO_CHAR_HI,
	const int iCtrlBackground = BG_INFO_CHAR, const int iCtrlForeground = FG_INFO_CHAR_LO )
```

source/Debugger/Debugger_Display.cpp
```diff
@@ -2438,18 +2409,18 @@ void _DrawSoftSwitchLanguageCardBank( RECT & rect, const int iBankDisplay, int b
...
-		if (GetCurrentExpansionMemType() == CT_Saturn128K)  { sText[0] = 's'; iActiveBank = GetCardMgr().GetLanguageCard()->GetActiveBank(); }
+		char chMemType = '?'; // Default to RAMWORKS
+		if (GetCurrentExpansionMemType() == CT_RamWorksIII) { chMemType = 'r'; iActiveBank = GetRamWorksActiveBank(); }
+		if (GetCurrentExpansionMemType() == CT_Saturn128K)  { chMemType = 's'; iActiveBank = GetCardMgr().GetLanguageCard()->GetActiveBank(); }
 
 		if (iActiveBank >= 0)
 		{
 			DebuggerSetColorFG( DebuggerGetColor( FG_INFO_REG )); // light blue
-			PrintTextCursorX( sText, rect );
+			const char szMemType[2] = { chMemType, '\0' };
```
Michaelangel007, line 2419: Please rename to `sMemType`.  We don't need the extra `z` clutter.

source/Debugger/Debugger_Display.cpp
```diff
@@ -2438,18 +2409,18 @@ void _DrawSoftSwitchLanguageCardBank( RECT & rect, const int iBankDisplay, int b
 		rect.right  += 4*w;
 
 		int iActiveBank = -1;
-		char sText[ 4 ] = "?"; // Default to RAMWORKS
-		if (GetCurrentExpansionMemType() == CT_RamWorksIII) { sText[0] = 'r'; iActiveBank = GetRamWorksActiveBank(); }
-		if (GetCurrentExpansionMemType() == CT_Saturn128K)  { sText[0] = 's'; iActiveBank = GetCardMgr().GetLanguageCard()->GetActiveBank(); }
+		char chMemType = '?'; // Default to RAMWORKS
```
Michaelangel007, line 2412: Thanks for catching and renaming the non-descript `sText`!

I'm not sure if we need `ch`.  Let's rename this to just use `c` (character) prefix so it matches the single letter `s` (string) for symmetry.
i.e.
`chMemType` -> `cMemType`

source/Debugger/Debugger_Display.cpp
```diff
@@ -2538,39 +2509,32 @@ void DrawSoftSwitches( int iSoftSwitch )
...
-	PrintTextCursorY( sText, rect );
+	PrintTextCursorY( !VideoGetSWPAGE2() ? "PAGE 1 / -" : "PAGE - / 2", rect );
 	
 	// $C056 / $C057 LORES/HIRES = HIRESOFF/HIRESON = SW.LORES/SW.HIRES
 	// LO / HIRES
 	// LO / -----
 	// -- / HIRES
-	sprintf( sText, !VideoGetSWHIRES() ? "LO /-- RES" : "---/HI RES" );
-	PrintTextCursorY( sText, rect );
+	PrintTextCursorY( !VideoGetSWHIRES() ? "LO /-- RES" : "---/HI RES", rect );
```
Michaelangel007, line 2527: While we're here, let's do some more minor cleanup for lines 2527-2537:
```
        PrintTextCursorY( !VideoGetSWHIRES()     ? "LO /-- RES"  : "---/HI RES" , rect ); PrintTextCursorY( "", rect );
	PrintTextCursorY( !VideoGetSW80COL()     ? "40 / -- COL" : "-- / 80 COL", rect ); // Extended soft switches
	PrintTextCursorY( VideoGetSWAltCharSet() ? "ASCII/-----" : "-----/MOUSE", rect );
	PrintTextCursorY( !VideoGetSWDHIRES()    ? "HGR / ----"  : "--- / DHGR" , rect ); // 280/560 HGR
```

source/Debugger/Debugger_Display.cpp
```diff
@@ -3061,16 +3004,12 @@ void DrawSubWindow_Data (Update_t bUpdate)
...
-	_ASSERT( CONSOLE_WIDTH > WINDOW_DATA_BYTES_PER_LINE );
 
-	char sOpcodes  [ CONSOLE_WIDTH ] = "";
-	char sImmediate[ 4 ]; // 'c'
+	_ASSERT( CONSOLE_WIDTH > (WINDOW_DATA_BYTES_PER_LINE * 3));
 
-	const int nDefaultFontWidth = 7; // g_aFontConfig[FONT_DISASM_DEFAULT]._nFontWidth or g_nFontWidthAvg
-	int X_OPCODE      =  6                    * nDefaultFontWidth;
-	int X_CHAR        = (6 + (nMaxOpcodes*3)) * nDefaultFontWidth;
+	const int nDefaultFontWidth	= 7; // g_aFontConfig[FONT_DISASM_DEFAULT]._nFontWidth or g_nFontWidthAvg
```
Michaelangel007, line 3010: These `=` equal signs should be columned aligned. They are but it is too hard to tell with this PR. Please replace TABs with Spaces.
```
	const int nDefaultFontWidth = 7; // g_aFontConfig[FONT_DISASM_DEFAULT]._nFontWidth or g_nFontWidthAvg
	const int X_OPCOD           =  6                    * nDefaultFontWidth;
	const int X_CHAR            = (6 + (nMaxOpcodes*3)) * nDefaultFontWidth;
```

Michaelangel007 (changes requested):
Looks good overall.  Some minor cleanup needed but this is coming along very nicely.

source/Debugger/Debugger_Display.cpp
kiyolee: Fixed in new commit.

source/Debugger/Debugger_Display.cpp
kiyolee: Fixed in new commit.

source/Debugger/Debugger_Display.cpp
kiyolee: Fixed in new commit.

source/Debugger/Debugger_Display.cpp
kiyolee: Fixed in new commit.

Michaelangel007 (approved):
Thanks! Looks good.

tomcw:
@kiyolee - I'm getting 2 warnings now that this PR has been merged to master:

```
1>D:\Apple][\AppleWin\git\master\AppleWin\source\Debugger\Debugger_Display.cpp(1886,27): warning C4244: 'argument': conversion from 'WORD' to 'uint8_t', possible loss of data
1>D:\Apple][\AppleWin\git\master\AppleWin\source\Debugger\Debugger_Display.cpp(2168,28): warning C4244: 'argument': conversion from 'WORD' to 'uint8_t', possible loss of data
```

kiyolee:
I do notice the couple of warnings. There are other similar size mismatching type warnings anyway, especially for 64-bit build. I thought these can be kept for to be addressed in the future. Of course simple casting can silence the warnings but I try to avoid simple casting only for the sake of suppressing warnings. That could hide potential problems. Let me see if these are safe to be just suppressed.

kiyolee:
https://github.com/AppleWin/AppleWin/pull/1094
I think this is the better way to address the couple of warnings.

PR #1091, opened 2022-04-17
Debugger: replace sprintf() part 5 - Debugger_DisassemblerData

kiyolee:
- Change `_GetAutoSymbolName()` to return `std::string`


tomcw:
Merged it.

PR #1094, opened 2022-04-23
Fix a couple of warnings from PR#1090

kiyolee:
1. The only call site of DrawFlags() is taking only a BYTE. The function signature should change to prevent upsizing BYTE to WORD and hence the warning.
2. nStackDepth could in theory go over 255, in that case, some special indication should be printed. With the condition checked, casting it to BYTE is now safe.


tomcw:
Stack depth can't every be more than 255.

So this code...
```
	if (PARAM_REG_SP == iSource)
	{
		WORD nStackDepth = _6502_STACK_END - nValue;
```

should change to:

```
	if (PARAM_REG_SP == iSource)
	{
		BYTE nStackDepth = _6502_STACK_END - (_6502_STACK_BEGIN + (nValue & 0xFF));
```

tomcw:
Thanks - now merged.

PR #1097, opened 2022-04-30
Uthernet II: add virtual DNS feature

audetto:
This PR adds a new feature to the Uthernet II card (not found on real hardware).

**Virtual DNS**: the card offers an interface for DNS resolution, allowing Apple ][ software to run **without** raw sockets.
And as a consequence, `pcap` is not needed, allowing operation on non-Ethernet networks.

There is a description of this feature at: https://github.com/a2retrosystems/uthernet2/wiki/Virtual-W5100-with-DNS
An example of an application using this feature is https://github.com/oliverschmidt/A2Stream

Non-DNS related changes

1. A new save state version for Uthernet II has been added.
2. Long standing issue of where the Uthernet flags are read has been resolved, They are now read in the constructor of the cards.
3. Ethernet configuration has been changed. `pcap` is not compulsory any longer, but without, only U II software dns-aware has any chance to work. This is mentioned in the configuration dialog.
4. Split `Config.h` -> `Config.cpp` as this files has become big and heavy.


audetto:
And ensure npcap's usage is always protected from inside `tfearch.cpp`.

This means as well that `SS_YAML_KEY_CANNOT_USE` from Uthernet 1 SaveState is ignored.

tomcw:
@audetto - in DNS.cpp, can you add the license header at the start of the file (ie. same as you did for Uthernet2.cpp).

resource/Applewin.rc
```diff
@@ -206,19 +206,22 @@ BEGIN
...
     LTEXT           "Ethernet",IDC_TFE_SETTINGS_ENABLE_T,9,7,30,8
     COMBOBOX        IDC_TFE_SETTINGS_ENABLE,45,5,60,80,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    CONTROL         "Virtual DNS",IDC_CHECK_TFE_VIRTUAL_DNS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,150,7,50,8
     LTEXT           "Interface",IDC_TFE_SETTINGS_INTERFACE_T,9,24,30,8
     COMBOBOX        IDC_TFE_SETTINGS_INTERFACE,45,22,210,80,CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
     LTEXT           "",IDC_TFE_SETTINGS_INTERFACE_NAME,9,44,250,8
     LTEXT           "",IDC_TFE_SETTINGS_INTERFACE_DESC,9,60,250,8
-    DEFPUSHBUTTON   "Ok",IDOK,20,75,50,14
-    PUSHBUTTON      "Cancel",IDCANCEL,80,75,50,14
+    GROUPBOX        "Npacp",IDC_STATIC,5,75,260,55
```
tomcw, line 221: There's a minor typo here: should be "Npcap"

source/W5100.h
```diff
@@ -46,6 +46,10 @@
 #define W5100_SN_MR_IPRAW         0x03
 #define W5100_SN_MR_MACRAW        0x04
 #define W5100_SN_MR_PPPOE         0x05
+#define W5100_SN_VIRTUAL_DNS      0x08
```
tomcw, line 49: Can you add a comment to say that this doesn't exist on a real W5100.

tomcw (changes requested):

resource/Applewin.rc
audetto: changed

source/W5100.h
audetto: done

tomcw (approved):

tomcw:
@audetto - thanks for this work - it's now merged into the mainline.

tomcw:
Test both a wired (my desktop) and a wireless (my laptop) connection - both working fine.

audetto:
Cool.
Do you think we could / should change the default to TRUE for this feature?

tomcw:
ie. should we make "Virtual DNS" the default when "Uthernet II" is selected?

I quickly tested Contiki (webbrowser) and A2Stream v1.0, and both worked fine with "Virtual DNS" checked.
NB. Only at A2Stream v1.2 was DNS offloading support added.

So for a better end-user experience then it seems like a good thing to make "Virtual DNS=TRUE" the default setting.
Are there any reasons not to do this?

tomcw:
@audetto, @oliverschmidt:
New AppleWin release with this DNS offload support: https://github.com/AppleWin/AppleWin/releases/tag/v1.30.10.0

sicklittlemonkey:
Awesome work guys. What a great release!

oliverschmidt:
Thanks for the release 😊

Here's the promised video - https://youtu.be/5oddV-wMyqg

PR #1100, opened 2022-05-14
Uthernet II: 2 fixes and minor changes

audetto:
Bug fix:

1. UDP: it would not take into account the size of the header (6 bytes) when computing the amount of data that can be read from the peer (so potentially reading 6 bytes too many, which would not fit)
2. IPRAW: The header should contain the `source` of the packet rather than the `destination` (which is useless as it is our IP address). Doc is not very clear, but for UDP (which they refer to) they use the peer IP (so the `source` of the packet), or destination of the socket.

Changes:

1. Make Virtual DNS enabled by default

### DIFF
--- a/source/Uthernet2.cpp
+++ b/source/Uthernet2.cpp
@@ -42,7 +42,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 // when the connect() calls is ongoing
 //
 // in the checks below we allow both in all cases
-// (myErrno == SOCK_EINPROGRESS || myErrno == SOCK_EWOULDBLOCK)
+// (errno == SOCK_EINPROGRESS || errno == SOCK_EWOULDBLOCK)
 // this works, bu we could instead define 2 functions and check only the correct one
 #ifdef _MSC_VER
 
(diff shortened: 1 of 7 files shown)


source/Uthernet2.cpp
```diff
@@ -1599,7 +1648,12 @@ void Uthernet2::SetRegistryVirtualDNS(UINT slot, const bool enabled)
 bool Uthernet2::GetRegistryVirtualDNS(UINT slot)
 {
     const std::string regSection = RegGetConfigSlotSection(slot);
-    DWORD enabled = 0;
-    RegLoadValue(regSection.c_str(), REGVALUE_UTHERNET_VIRTUAL_DNS, TRUE, &enabled, 0);
+
+    // be default VirtualDNS is enabled
```
tomcw, line 1652: Minor thing... the comment here doesn't make sense: should it say "The default VirtualDNS is enabled" ?

source/Uthernet2.h
```diff
@@ -44,6 +45,11 @@ struct Socket
     Socket();
 
     ~Socket();
+
+private:
+    socket_t myFD;
+    uint8_t mySocketStatus;  // sn_sr
```
tomcw, line 51: What is does this comment "sn_sr" mean?
Is it the old variable name?

audetto: Changed

source/Uthernet2.cpp
sicklittlemonkey: Did you mean "By"? It makes more sense:
// By default, VirtualDNS is enabled
// as it is backward compatible

sicklittlemonkey:
I commented on the commit, but it was outdated, so maybe nobody noticed.

Did you mean "By"? It makes more sense:
// By default, VirtualDNS is enabled
// as it is backward compatible

Cheers,
Nick.

tomcw:
Thanks Nick - yes, I agree, and I've updated it.

Issue #1113, opened 2022-06-08
Uthernet II regs repeat in DEVICE SELECT I/O space


tomcw:
From Slack / #emulators, @oliverschmidt said:

>actually those addresses are repeated 4 times (as you'd expect). The documentation says to use 0x4-0x7 in oder to stay compatible with potential future enhanced UII versions. However, when you emulate the UII as-is, then you might want to have them repeated. I seem to remember that there was/is some UII software not caring about the docs and using 0x0-0x3...

>Only A0 and A1 are connected. Therefore 0x00 == 0x04 == 0x08 == 0x0C.

But in `IO_C0()`, we only support regs 0x04..0x07:
https://github.com/AppleWin/AppleWin/blob/6a934bcfef78e14f3a0e5f2da8db2231024dfe38/source/Uthernet2.cpp#L1402-L1410

More accurate to zero base these 4 registers and in `IO_C0()` mask with 0x03 (instead of 0x0F):
https://github.com/AppleWin/AppleWin/blob/6a934bcfef78e14f3a0e5f2da8db2231024dfe38/source/W5100.h#L5

audetto:
The change seems easy enough.
But I could not find anything in the manual: http://dserver.macgui.com/Uthernet%20II%20manual%20draft.pdf

Is it undocumented hardware behavior?

oliverschmidt:
The documentation is for the user - not for the implementor of an emulation ;-)

As I mentioned in the Slack channel: The hard decodes only A0 and A1, A2 and A3 are ignored.

audetto:
Do you mind sharing a link to the slack channel?

oliverschmidt:
Here's the direct linke to the conversation: https://apple2infinitum.slack.com/archives/CABEM8JFK/p1654532874669579

audetto:
Do I need to be invited? There is no way to create an account.

oliverschmidt:
Sorry, I wasn't aware of that. You should have received an invitation now.

PR #1115, opened 2022-06-11
Uthernet II: the real hardware only decodes A0 and A1, so we introduc…

audetto:
…e a 0x03 mask.

Fixes: https://github.com/AppleWin/AppleWin/issues/1113


tomcw:
Thanks - merged it in.

Issue #1110, opened 2022-06-04
Possible bug in .WOZ & Quarter Tracks


burniouf:
if you do:

_**.9				lda D2.CurrentQTrack
				cmp	D2.TargetQTrack
				bne .3

				lsr						CS if X,Y on

				jsr D2.Wait25600usec

				lda IO.D2.Ph0Off,y
				
				bcc .90

				lda IO.D2.Ph0Off,x
				clc

.90				rts**_

the head is supposed to stay between X and Y phases, within Applewin, it moves back to Half track at X

tomcw:
I think it depends on the state of the adjacent phase magnets.

Can you attach a save-state file (.yaml) just before this code is run.

NB. you can do this precisely from within the AppleWin debugger - ie. F11 (Save state) will take effect from the current opcode you are stopped at.

burniouf:
[YAML.zip](https://github.com/AppleWin/AppleWin/files/8840040/YAML.zip)

burniouf:
BEFORE is before lda IO.D2.Ph0Off,y
BETWEEN is after lda IO.D2.Ph0Off,y
AFTER is after lda IO.D2.Ph0Off,x

burniouf:
BEFORE
        Unit0:
          Drive Connected: true
          Phase: 34
          Phase (precise): 34.500000
          Head Window: 0x0
          Last Stepper Cycle: 0x000000009DEE5C03
          Motor On Cycle: 0x000000009D87D40F
          Spinning: 1000000
BETWEEN
       Unit0:
          Drive Connected: true
          Phase: 34
          Phase (precise): 34.000000
          Head Window: 0x0
          Last Stepper Cycle: 0x000000009DEF0967
          Motor On Cycle: 0x000000009D87D40F
          Spinning: 1000000
AFTER
        Unit0:
          Drive Connected: true
          Phase: 34
          Phase (precise): 34.000000
          Head Window: 0x1
          Last Stepper Cycle: 0x000000009DEF096D
          Motor On Cycle: 0x000000009D87D40F
          Spinning: 1000000

tomcw:
It looks like a bug, since rapidly turning of magnet phases 3 & 2 should not affect the position.

From Sather (UTA2e 9-13):
>"...turn two adjacent phases on ... This is referred to as stepping to a quarter track. The two phases must be turned off in rapid sequence if one of them is phase 1, the drive is unmodified, and data is to be written on the quarter track."

[...]

PR #1116, opened 2022-06-12
Two adjacent magnets off in quick succession

tomcw:
PR for #1110.

If 2 _adjacent_ stepper control OFF accesses occur within 10 cycles, then update the magnet states, but don't move the cog. So the track/phase doesn't change.

Developer info: all Control Stepper access are now deferred by 10 cycles, which allows detecting of the above condition. The SyncEvent functionality is used to allow this very fine level of detection, without impacting performance - ie. it's the same functionality used for precise interrupts from Mockingboard(6522) & Mouse(VBl) cards.

### DIFF
--- a/source/Disk.cpp
+++ b/source/Disk.cpp
@@ -36,6 +36,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 #include "Interface.h"
 #include "Core.h"
+#include "CardManager.h"
 #include "CPU.h"
 #include "DiskImage.h"
 #include "Log.h"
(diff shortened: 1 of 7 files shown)


tomcw:
@sicklittlemonkey, Hi Nick - can I get you to cast your eye over this issue (#1110) and possible fix?

btw, the 10 cycle number is just based on the A2osX issue, I'm sure I could increase it, but keeping it narrow like this will restrict the fix to just this Apple II stepper code. Put another way: if you are positioned at a quarter track (with 2 adjacent magnets on), then when turning just 1 magnet off, how long does it take for the magnetic attraction to move the cog to a half-track position?

source/Disk.cpp
```diff
@@ -484,10 +492,58 @@ void __stdcall Disk2InterfaceCard::ControlStepper(WORD, WORD address, BYTE, BYTE
...
+void Disk2InterfaceCard::ControlStepperDeferred(bool rapidMagnetChange, WORD nextAddress)
+{
+	m_deferredStepperEvent = false;
+	const WORD address = m_deferredStepperAddress;
+
+	FloppyDrive* pDrive = &m_floppyDrive[m_currDrive];
+	FloppyDisk* pFloppy = &pDrive->m_disk;
 
 	// check for any stepping effect from a magnet
 	// - move only when the magnet opposite the cog is off
```
sicklittlemonkey, line 549: You might want to adjust the comment about momentum and timing, as that's what's being emulated now. : - )

sicklittlemonkey (approved):
I only had time for a couple of quick skims over the new code. I knew the event system would come in handy, but didn't have this in mind! Very nice. BTW, RWTS uses a tiny little table for the head settling time delay. Anyway, if it works then it looks good!

tomcw:
Thanks Nick. I'll update the comment, and there's also a code-path that needs a small fix.
I've been testing extensively with regular .dsk, but also .woz (eg. spira-track and other quarter track games).

PR #1162, opened 2023-01-07
[WinMM] Fix joysticks not detected : first & second devices are not always at index 0 & 1

fabricecaruso:
We can't trust JOYSTICKID1 (0) & JOYSTICKID2 (1) indexes with WinMM. ( my 1st device is @ index 4 )
So it's mostly not working & not detecting joysticks.

Detect JOYSTICK_1 & JOYSTICK_2 real indexes instead.

### DIFF
--- a/source/Joystick.cpp
+++ b/source/Joystick.cpp
@@ -51,6 +51,9 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 enum {DEVICE_NONE=0, DEVICE_JOYSTICK, DEVICE_KEYBOARD, DEVICE_MOUSE, DEVICE_JOYSTICK_THUMBSTICK2};
 
+int JOYSTICK_1 = -1;
+int JOYSTICK_2 = -1;
+
 // Indexed by joytype[n]
 static const DWORD joyinfo[6] =	{	DEVICE_NONE,
 									DEVICE_JOYSTICK,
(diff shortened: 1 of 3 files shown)


tomcw:
Thanks for this PR.

> my 1st device is @ index 4

So what is at indices 0,1,2 and 3?

And what can I do to test this on my system? (ie. how can I move my device to an index that isn't 0 or 1?)

In terms of coding style, I'm keen to get rid of global variables (eg. `int JOYSTICK_1 = -1;` so declare it `static`), and instead use getter/setting functions when read/writing that variable from other C++ files. Can you change to use eg. `int GetJoystick1(void)` and `int GetJoystick2(void)`.

Note: I will update the Coding Conventions doc to include this point.

fabricecaruso:
`So what is at indices 0,1,2 and 3?`
I don't really know !! It's just nothing. 

I presume it's related to the fact that I tested many controllers with my machine in the past, and there is probably something stored from my previous devices somewhere. My 8bitDo is always @ index 4, and my xbox controller is always @ index 6...
So it's probably a 'controller driver' index.

Anyway : It's not a connected device index for sure !

I can't find any official documentation from Microsoft about this, but all working samples I found on the net were doing like I did by looping joyGetNumDevs & testing joyGetDevCaps/joyGetPos.

Windows mysteries....

xotmatrix:
As I understand it, which isn't terribly well, there is no predicting which slot an XInput controller will occupy when it is connected. XInput supports four XInput controllers and reserves slots 0-3 for them. DirectInput controller can occupy the next group of slots (at least 8 are supported). Software usually prompts the user or scans for an expected input to decide which controller is in use.

source/Joystick.cpp
```diff
@@ -51,6 +51,9 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 enum {DEVICE_NONE=0, DEVICE_JOYSTICK, DEVICE_KEYBOARD, DEVICE_MOUSE, DEVICE_JOYSTICK_THUMBSTICK2};
 
+int JOYSTICK_1 = -1;
+int JOYSTICK_2 = -1;
+
```
tomcw, line 56: Make `JOYSTICK_1` and `JOYSTICK_2` static.
Add getter/setter functions, eg. `int GetJoystick1(void)` and `int GetJoystick2(void)`.
Use them when read/writing these variables from other C++ files.

tomcw:
Thank you. PR tweaked + merged to master.

PR #1177, opened 2023-02-05
Mockingboard: fix out of bounds access.

audetto:


tomcw:
Thanks - don't know how I missed that!

PR #1184, opened 2023-02-19
Properly catch installation of multiple Z80 cards.

ThorstenBr:
Corrected the check if multiple Z80 cards are installed. The check was already there, but didn't work - due to incomplete copy&pasting - and a minor goof with the class name...


tomcw:
How does that line `class m_pZ80Card* m_pZ80Card;` even make it through the compiler?!
Coverity - static analysis tool - didn't complain either!

Anyway, thanks for raising this. (One advantage of OSS = more eyes on the code.)

tomcw:
Now merged into the main line.

PR #1203, opened 2023-03-26
Debugger step over

Michaelangel007:
New behavior showing the stack on failure/success. Detect stack state on failure after `JSR` (3 possible states).

1. Failure: RTS address on stack sans extra junk
2_step_over_0x1f2

2. Failure: RTS address on stack with extra junk
1_step_over_0x1f1

3. Failure: RTS address not found on stack (!)
3_step_over_0x1f3

4. Successful step over, like 2, but with increased max execution steps.
4_step_over

### DIFF
--- a/source/Debugger/Debugger_Assembler.cpp
+++ b/source/Debugger/Debugger_Assembler.cpp
@@ -468,7 +468,11 @@ WORD _6502_GetStackReturnAddress ()
 	return nAddress;
 }
 
-// NOTE: nStack is both an input and output
+// NOTES: nStack is both an input and output;
+//        If nStack is 0x1FF it WILL return overflow 0x200.  Current callers are:
+//          _6502_FindStackReturnAddress(), and
+//          _6502_GetStackReturnAddress()
+//       which DON'T return this overflow stack value to previous callers.
 //===========================================================================
 WORD _6502_PeekStackReturnAddress (WORD & nStack)
 {
--- a/source/Debugger/Debugger_Assembler.cpp
+++ b/source/Debugger/Debugger_Assembler.cpp
@@ -464,11 +464,12 @@ bool _6502_CalcRelativeOffset ( int nOpcode, int nBaseAddress, int nTargetAddres
 	return false;
 }
 
-// Return stack offset if the address is on stack, else 0 if not found
+// Return stack offset if the address is on stack, else -1 if not found
 //===========================================================================
 int _6502_FindStackReturnAddress (const WORD & nAddress)
 {
 	WORD nStack = regs.sp;
+	int  nDepth = -1; // not found
 	nStack++;
 
 	while (nStack <= (_6502_STACK_END - 1))
@@ -480,10 +481,13 @@ int _6502_FindStackReturnAddress (const WORD & nAddress)
 		nReturnAddress++;
 
 		if (nReturnAddress == nAddress)
-			return (nStack-1 - regs.sp);
+		{
+			nDepth = (nStack - 2 - regs.sp);
+			return nDepth;
+		}
 	}
 
-	return 0; // not found
+	return nDepth;
 }
 
 //===========================================================================
--- a/source/Debugger/Debugger_Assembler.cpp
+++ b/source/Debugger/Debugger_Assembler.cpp
@@ -587,18 +587,20 @@ void _6502_GetOpcodeOpmodeOpbyte (int & iOpcode_, int & iOpmode_, int & nOpbyte_
 	iOpcode_ = _6502_GetOpmodeOpbyte( regs.pc, iOpmode_, nOpbyte_ );
 }
 
+// NOTE: If the stack pointer is <= 01 before a JSR the stack will wrap around.
+// Use bWrapAround to access stack memory in this case.
 //===========================================================================
-bool _6502_GetStackReturnAddress (WORD & nAddress_)
+bool _6502_GetStackReturnAddress (WORD & nAddress_, bool bWrapAround /* = false */)
 {
 	unsigned nStack = regs.sp;
 	nStack++;
 
-	if (nStack <= (_6502_STACK_END - 1))
+	if (bWrapAround || nStack <= (_6502_STACK_END - 1)) // JSR always pushes 2 bytes.  If no wrap around then the stack pointer must be <= 0x1FE
 	{
-		nAddress_ = (unsigned)*(LPBYTE)(mem + nStack);
+		nAddress_ = (unsigned)*(LPBYTE)(mem + 0x100 + (nStack & 0xFF));
 		nStack++;
 		
-		nAddress_ += ((unsigned)*(LPBYTE)(mem + nStack)) << 8;
+		nAddress_ += ((unsigned)*(LPBYTE)(mem + 0x100 + (nStack & 0xFF))) << 8;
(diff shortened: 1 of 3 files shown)


source/Debugger/Debugger_Assembler.cpp
```diff
@@ -464,6 +464,27 @@ bool _6502_CalcRelativeOffset ( int nOpcode, int nBaseAddress, int nTargetAddres
 	return false;
 }
 
+// Return stack offset if the address is on stack, else 0 if not found
+//===========================================================================
+int _6502_FindStackReturnAddress (const WORD & nAddress)
+{
+	unsigned nStack = regs.sp;
```
tomcw, line 471: "unsigned" is missing a type... assume `WORD`
And in 2 other occurrences below...

source/Debugger/Debugger_Assembler.cpp
```diff
@@ -464,6 +464,27 @@ bool _6502_CalcRelativeOffset ( int nOpcode, int nBaseAddress, int nTargetAddres
...
+	nStack++;
+
+	while (nStack <= (_6502_STACK_END - 1))
+	{
+		WORD nReturnAddress = (unsigned)*(LPBYTE)(mem + nStack);
+		nStack++;
```
tomcw, line 477: Need to handle the wraparound case for this `nStack++`, eg:
`nStack = ((nStack+1)&0xff)+0x100;`

source/Debugger/Debugger_Assembler.cpp
```diff
@@ -464,6 +464,27 @@ bool _6502_CalcRelativeOffset ( int nOpcode, int nBaseAddress, int nTargetAddres
...
+
+		nReturnAddress += ((unsigned)*(LPBYTE)(mem + nStack)) << 8;
+		nReturnAddress++;
+
+		if (nReturnAddress ==nAddress)
```
tomcw, line 482: Missing a space after `==`.

source/Debugger/Debugger_Assembler.cpp
```diff
@@ -464,6 +464,27 @@ bool _6502_CalcRelativeOffset ( int nOpcode, int nBaseAddress, int nTargetAddres
...
+			return (nStack-1 - regs.sp);
+	}
+
+	return 0; // not found
```
tomcw, line 486: I haven't fully traced the logic, but isn't a value of 0(zero) a valid stack offset?
So maybe return -1 instead for this "not found" case.

tomcw (changes requested):

source/Debugger/Debugger_Assembler.cpp
Michaelangel007: Fixed f8533d5c

source/Debugger/Debugger_Assembler.cpp
Michaelangel007: Fixed 29796e4e

source/Debugger/Debugger_Assembler.cpp
Michaelangel007: Interesting edge case, when sp = 0x101.  Need to track down why this assert is triggering: `_ASSERT(g_nAppMode == MODE_RUNNING || g_nAppMode == MODE_STEPPING);`

source/Debugger/Debugger_Assembler.cpp
Michaelangel007: Thanks for the catch.  Fixed with f8bdedcc

source/Debugger/Debugger_Assembler.cpp
Michaelangel007: Fixed with commit c49569fa

Michaelangel007:
@tomcw Should be all fixed now. I've also clarified how to repro/test the old behavior.

tomcw (changes requested):

source/Debugger/Debugger_Assembler.cpp
```diff
@@ -464,11 +464,12 @@ bool _6502_CalcRelativeOffset ( int nOpcode, int nBaseAddress, int nTargetAddres
 	return false;
 }
 
-// Return stack offset if the address is on stack, else 0 if not found
+// Return stack offset if the address is on stack, else -1 if not found
 //===========================================================================
 int _6502_FindStackReturnAddress (const WORD & nAddress)
```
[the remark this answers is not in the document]
Michaelangel007: Thanks for catching that old habit pass-by-ref.

I've cleaned up all the stack return address with `_6502_PeekStackReturnAddress()`.

Also, `_6502_GetStackReturnAddress()` no longer returns a redundant bool but the actual return address.  This caused a minor cascade cleanup in `CmdStepOut()` and `CmdCursorJumpRetAddr()`.

Fixed with commit 55f863d1

Michaelangel007:
@tomcw What's the protocol for `Resolve Conversation`  ? The person who opened the PR or the reviewer?  TIA.

source/Debugger/Debugger_Assembler.cpp
```diff
@@ -426,8 +426,60 @@ Fx	BEQ r  SBC (d),Y  sbc (z)  ---  ---      SBC d,X  INC z,X  ---  SED  SBC a,Y
...
+	WORD nAddress = _6502_PeekStackReturnAddress(nStack);
+	return nAddress;
+}
+
+// NOTE: nStack is both an input and output
+//===========================================================================
+WORD _6502_PeekStackReturnAddress (WORD & nStack)
+{
+	WORD   nAddress;
+	       nAddress  = ((unsigned) *(LPBYTE)(mem + 0x100 + (nStack & 0xFF))     ); nStack++;
```
tomcw, line 476: Since `nStack` is an input and output, then you need to wrap it correctly:
`nStack = 0x100 + ((nStack+1) & 0xff);`

Hmm, actually since _currently_ it's only used afterwards by `_6502_FindStackReturnAddress()`, then for this specific case when it overflows (ie. IN: nStack = 0x1ff, OUT: nStack = 0x200), then you use `nStack` afterwards here:

```
nDepth = (nStack - 2 - regs.sp);
```

So for this overflow case: nDepth = 0x200 - 2 - regs.sp = 0x1fe - regs.sp... so perhaps fine?
... and nDepth = 0x200 - 2 - 0x1fe = 0.

So looks to be fine.

---

At the other extreme... IN: nStack=0x100, OUT: nStack=0x101

So for this case: nDepth = 0x101 - 2 - regs.sp = 0x0ff - regs.sp... which doesn't look right to me, since regs.sp is >= 0x100, so nDepth is a negative value.

OK, ignore me for this case, since `nStack = regs.sp + 1`, so as a minimum IN: nStack=0x101, OUT: nStack=0x102
... and nDepth = 0x102 - 2 - 0x100 = 0.

---

In conclusion it all seems fine.

Could you add a comment after this line "// NOTE: nStack is both an input and output", just saying that the overflow for the case when nStack=0x1ff and returns with 0x200 is fine, since the calling function accounts for this.

Michaelangel007: Added note with commit 64b4fb77

tomcw: Thanks Michael - all looks fine now.

PR #1209, opened 2023-04-09
Update BMP import from GIMP 2.10 with X8 R8 G8 B8

Michaelangel007:
This patch makes the debugger `NTSC LOAD` properly import Windows BMP files exported from GIMP again.   (Found while investigating issue #357 and re-verifying the NTSC palette loading in post 1.26.)

# Technical Details

Apparently GIMP changed the way they save BMPs now.

There are multiple version of a Windows BMP file.

Struct Size 

* 0x28 version 3 (and prior?)
* 0x38 version 4 which has Red, Green, and Blue channel masks. (Wish hardware and graphics people would just pick ONE bloody RGBA standard for once but I digress...)

The original `.bmp` files `Default_NTSC_Palette.bmp`, `White_NTSC_Palette.bmp`, `Black_NTSC_Palette.bmp`, `Tweaked_Palette.bmp` on my page http://michael.peopleofhonoronly.com/dev/applewin/ntsc/ have different Red, Green, Blue masks.  Re-exporting them as `.bmp` will cause the debugger to fail to load these due to GIMP 2.10 changing the Red, Green, Blue masks which was causing an incorrect swizzling.

### DIFF
--- a/docs/Debugger_Changelog.txt
+++ b/docs/Debugger_Changelog.txt
@@ -1,4 +1,16 @@
 /*
+2.9.1.18 Fixed: NTSC LOAD was failing to import BMPs exported from GIMP.
+    To export a 64x256 bitmap from GIMP:
+      1. File, Export As...
+      2. Use a file extension of .bmp
+      3. Export Image as BMP
+         Compatibility Options
+           [x] Do not write color space information
+         Advanced Options
+           32 Bits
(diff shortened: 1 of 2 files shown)


tomcw:
LGTM!

tomcw (approved):

PR #1212, opened 2023-04-15
Debugger DISK INFO QoL

Michaelangel007:
The debugger's `DISK INFO` was lacking readability. Small QoL patch.

Repro:

    <F7>
    BPX C65E
    <F7>
    Boot any .woz (floppy) disk
    DISK INFO

Before:
disk_info_before

After:
disk_info_after

Also alphabetized the DiskGetCurrent*() helper functions.


Michaelangel007:
@tomcw The last number is the Shift Register.  Would it make more sense to display this with the shift operator as a mnemonic?

    Reading << 03

tomcw:
I prefer leaving the text as "bitOffset" (instead of just "Offset"), otherwise it's ambiguous (ie. is it a byte or bit offset?)

tomcw:
2. If you look at the strings that `Disk2InterfaceCard::GetCurrentState()` can return, then I think it's only useful to return the shift register for `Writing` and `Reading` states.

3. > Would it make more sense to display this with the shift operator as a visual mnemonic? i.e. `Reading << 03`

Probably fine as-is (so without the "<<" operator).

4. You have done work in another PR to truncate the line at 78 chars. But `Disk2InterfaceCard::GetCurrentState()` can return these longer strings: `Off (spinning)` and `Writing (write protected)`. Worth checking what the line looks like for these 2 cases.

Michaelangel007:
* Split the disk status string into a 2nd line (which I'm not really happy with) but it always guaranteed to fit.
* Colorized the disk status string.
* Added 3 shift symobls:
  * `<<` reading
  * `>>` writing
  * `>|` write protected
* Lastly, mirrored m_shiftReg for non `.WOZ` disks so you can see the last nibble read/wrote.

image

Michaelangel007:
@tomcw Made another pass.  I'm pretty happy with this now.  I would have preferred a "short" disk status string to fit on one line but I'll live with a descriptive 2-line output since the colorized output is way nicer to read now.

source/Disk.cpp
```diff
@@ -130,24 +132,44 @@ std::string Disk2InterfaceCard::GetCurrentPhaseString(void)
 	return FormatIntFracString(m_floppyDrive[m_currDrive].m_phasePrecise, true);
 }
 
-LPCTSTR Disk2InterfaceCard::GetCurrentState(void)
+LPCTSTR Disk2InterfaceCard::GetCurrentState(Disk_Status_e *pDiskState)
```
tomcw, line 135: If you change this function's arg to a reference: `LPCTSTR Disk2InterfaceCard::GetCurrentState(Disk_Status_e& diskState)`, then you can get rid of all the `if(pDiskState)` checks.

Michaelangel007: Good idea. Fixed.

Michaelangel007: Switching to a reference allows us to table drive the strings.  See commit [69bb9f9](https://github.com/AppleWin/AppleWin/pull/1212/commits/69bb9f9fff3ed1d3790f0bf17a29d8fdea587d94).

Michaelangel007:
* Refactored output to be cleaner and table driven.
* Added undocumented support for 1 line output.  Will add `CONFIG DISK #` later.

disk_info_one_line

Michaelangel007:
Should be good to go now.

source/Windows/WinFrame.cpp
```diff
@@ -68,8 +68,9 @@ static const DWORD g_aDiskFullScreenColorsLED[ NUM_DISK_STATUS ] =
 	RGB(  0,  0,  0), // DISK_STATUS_OFF   BLACK
 	RGB(  0,255,  0), // DISK_STATUS_READ  GREEN
 	RGB(255,  0,  0), // DISK_STATUS_WRITE RED
-	RGB(255,128,  0)  // DISK_STATUS_PROT  ORANGE
-//	RGB(  0,  0,255)  // DISK_STATUS_PROT  -blue-
+	RGB(255,128,  0), // DISK_STATUS_PROT  ORANGE
+	RGB(  0,  0,255), // DISK_STATUS_EMPTY -blue-
+	RGB(  0,1280,128)  // DISK_STATUS_SPIN  -cyan-
```
tomcw, line 73: Just noticed this typo... 1280 should be 128!

Michaelangel007: Ack!  Thank for the catch, fixed.

tomcw (approved):

tomcw:
Thanks Michael. I've merged it into the main line now.
(I meant to remove some of the minor commit comments, but hit the "Commit" button accidentally!)

PR #1214, opened 2023-04-17
RIFF: fix sizes.

audetto:
Previously `g_dwTotalNumberOfBytesWritten` was the number of sample-bytes written, while it was used as total number of bytes written (i.e. the file size).


audetto:
The easiest way to see it, is to create a 0-length wav file.
The previous code would underflow.

tomcw:
Thanks @audetto - merged into main line.

PR #1218, opened 2023-04-29
Fix a linux/gcc issue where sizeof(wchar_t)==4.

audetto:
`std::fill_n` works everywhere.


source/Speaker.cpp
```diff
@@ -39,6 +39,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #include "Riff.h"
 
 #include "Debugger/Debug.h"	// For DWORD extbench
+#include <algorithm>
```
tomcw, line 42: You shouldn't need this, since it's already including by stdafx.h.

audetto: Remove

tomcw:
Merged it.

PR #1222, opened 2023-05-02
Mockingboard: do not play sound unless a card is present.

audetto:
This change improves

1. save some CPU cycles if no card is present
2. help debugging IDirectSoundBuffer underruns

The are NO underruns, but with this patch, when the read pointer overtakes the write point, it is a real underrun.


tomcw:
Merged it.

PR #1225, opened 2023-05-06
Avoid a potential memory leak if RamWorks memory size were configurable.

audetto:
And add an accessor.

This is not an issue in AW, but it happens if one allows to configure RW size.
The check for NULL is enough to detect exactly which ones were allocated.

### DIFF
--- a/source/Memory.cpp
+++ b/source/Memory.cpp
@@ -1296,7 +1296,7 @@ void MemDestroy()
 	delete [] pCxRomPeripheral;
 
 #ifdef RAMWORKS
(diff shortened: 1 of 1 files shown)


tomcw:
You've added a new accessor: `UINT GetRamWorksMemorySize()`
But from the changes it doesn't look like it's ever called. Is it needed?

audetto:
You are right, not needed in AW.
I will remove it and keep it in my code.

tomcw:
Merged it.

PR #1229, opened 2023-05-30
Debugger df qol

Michaelangel007:
@tomcw Some small Quality of Life for the debugger `df` command.

Before:
before

After:
after


tomcw:
Thanks Michael - looks good.

PR #1230, opened 2023-05-30
Allow alternate directory for built-in symbols

sh95014:
needed because macOS apps are a bundle (tree of directories) and resources are packaged somewhere within, not necessarily in the same directory as the executable.

probably doesn't deal with the case of user-supplied symbols, but I haven't gotten that far yet...

https://github.com/sh95014/AppleWin/issues/12

### DIFF
--- a/source/Debugger/Debugger_Symbols.cpp
+++ b/source/Debugger/Debugger_Symbols.cpp
@@ -751,13 +751,13 @@ Update_t CmdSymbolsLoad (int nArgs)
 	{
 		sFileName = g_sProgramDir + g_sFileNameSymbols[ iSymbolTable ];
 		nSymbols = ParseSymbolTable( sFileName, (SymbolTable_Index_e) iSymbolTable );
-        
-        // Try optional alternate location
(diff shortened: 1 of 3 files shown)


source/Debugger/Debugger_Symbols.cpp
```diff
@@ -749,8 +749,15 @@ Update_t CmdSymbolsLoad (int nArgs)
 	// Debugger will call us with 0 args on startup as a way to pre-load symbol tables
 	if (! nArgs)
 	{
-		sFileName += g_sFileNameSymbols[ iSymbolTable ];
+		sFileName = g_sProgramDir + g_sFileNameSymbols[ iSymbolTable ];
 		nSymbols = ParseSymbolTable( sFileName, (SymbolTable_Index_e) iSymbolTable );
+        
+        // Try optional alternate location
+        if (nSymbols == 0 && !g_sBuiltinSymbolsDir.empty())
```
Michaelangel007: Pleas add a set of parenthesis around `nSymbols == 0`.

Other then that looks good.

Michaelangel007 (approved):

PR #1233, opened 2023-06-02
remove reference to processor status register

sh95014:
The "R P \<xx\>" command doesn't actually work:

```
>r p ff
Category: CPU
R, Set register
 Usage: <reg> <value | expression | symbol>
  Where <reg> is one of: A X Y PC SP 
 See also: OPERATORS
 Examples:
  R PC RESET + 1
  R PC $FC58
  R A  A1
  R A  $A1
  R A  #A1
```

so this PR removes the reference in the help doc.


Michaelangel007:
That's actually a bug.  Will investigate tomorrow.

Michaelangel007:
Thanks for catching this!

Closing since invalid.

PR #1239, opened 2023-06-07
Remove some StringCb functions and fix assertion.

audetto:


source/Disk.cpp
```diff
@@ -243,14 +243,7 @@ void Disk2InterfaceCard::SaveLastDiskImage(const int drive)
...
 
-	TCHAR szPathName[MAX_PATH];
-	StringCbCopy(szPathName, MAX_PATH, pathName.c_str());
-	TCHAR* slash = _tcsrchr(szPathName, PATH_SEPARATOR);
-	if (slash != NULL)
-	{
-		slash[1] = '\0';
-		RegSaveString(REG_PREFS, REGVALUE_PREF_START_DIR, 1, szPathName);
-	}
+	RegSaveString(REG_PREFS, REGVALUE_PREF_START_DIR, 1, pathName + PATH_SEPARATOR);
```
audetto, line 246: wait a minute, this is not quite right.
let me revisit it.

audetto: Fixed in the new commit.

tomcw:
Thanks... and merged.

Issue #1237, opened 2023-06-03
Joystick axis shaping


xotmatrix:
A recent change (#1162) fixed the long-standing joystick detection problems I've had, so many, many thanks for fixing that. But it appears we need to look into axis shaping since XInput analog sticks are physically bound to a circle. This makes reaching the Apple paddle/joystick "corners" impossible.

![AppleWin_2023-06-03_14-03-52](https://github.com/AppleWin/AppleWin/assets/14209329/1620935a-8ffb-466b-aa10-cbf821e7b5f9)

Also, thank you for making building AppleWin with VS2022 completely painless. Building AppleWin is something I had been putting off forever. I'm not a dynamite C++ programmer but I may be able to offer better help in the future.

audetto:
Yes, see for a previous discussion here: https://github.com/AppleWin/AppleWin/issues/429

In Linux, it uses a remap from circle to square: https://github.com/audetto/AppleWin/blob/944b84574ac788dfa9a66bea5e524d053cac1118/source/linux/paddle.cpp#L60

xotmatrix:
Shoot, I somehow missed that issue. In any event, I think it needed to be surfaced ahead of the next release's improved controller detection.

> remap from circle to square
That's an interesting solution. I've tried different schemes but I always seem to go back to multiplying by sqrt(2).

tomcw:
@xotmatrix 
>In any event, I think it needed to be surfaced ahead of the next release's improved controller detection.

Yes, it'll be in the next release (whenever that happens... but soon I'd hope).

tomcw:
Fixed in [AppleWin 1.30.15.0](https://github.com/AppleWin/AppleWin/releases/tag/v1.30.15.0), so closing this issue.

Issue #429, opened 2017-06-11
Joystick: cannot reach corners


audetto:
Hi,

I'm using AppleWin 1.26.2.4 via Wine and I've realised I struggle to move the joystick in the corners, especially TopRight and Top Left.

For instance, this means I cant run when I play Karateka.

I've tried to use COLOR DEMOSOFT from the Apple DOS 3.3 disk and indeed I cannot cover the whole screen.

I am using a XBOX 360 controller, and it looks like the control is a circle and not a square like the Apple 2 Joystick.

Or maybe it is an artifact introduced by Wine?

![color demosoft](https://user-images.githubusercontent.com/212523/27012742-385f52ec-4ece-11e7-9539-fb63f304fca2.png)

tomcw:
Using a controller with analogue thumbsticks (in my case a PS4 Dualshock 4) I get the same result as you. I'm running natively on Win7-64.

I've tested with 1.25.0.4 and the latest from git master (ie. after Nick's latest changes in this area). Both behave the same - ie. as reported in this issue.

A link to COLOR DEMOSOFT is [here](https://mirrors.apple2.org.za/ftp.apple.asimov.net/images/productivity/graphics/misc/color-demosoft.dsk).

A simpler repro is from AppleSoft run this simple code:
```
10 ?PDL(0),PDL(1): GOTO 10
```
...you can't get both values to be simultaneously 0.

NB. Using a mouse as a joystick works fine - full coverage in all corners & everywhere else on the screen. Also cursor keys and a USB digital joypad work "fine" (albeit the granularity is only left/centre/right, ie: 0/128/255).

audetto:
I was thinking that one can remap the circle to the square.

If x and y are in the unit circle centred at 0,0

C = min(x^2 / y^2, y^2 / x^2)
D = Sqrt(1 + C)

D is basically the length of the line from 0,0 passing by x,y till the edge of the square.

x*D and y*D will cover the unit square.
I have not tried it and cannot say how much distortion it introduces.

sicklittlemonkey:
Yes, joystick programming has some interesting topics like deadzones and
squaring the circle.

Here are some good links. (I like pictures that help me understand things.)

http://theinstructionlimit.com/squaring-the-thumbsticks
http://blog.hypersect.com/interpreting-analog-sticks/

Cheers,
Nick.

audetto:
The first one has the same formula I posted above, just with too many ifs.
[...]

PR #1242, opened 2023-06-18
Support 'squaring the circle' for analog sticks

tomcw:
PR for #429 & #1237

### DIFF
--- a/source/Joystick.cpp
+++ b/source/Joystick.cpp
@@ -703,10 +703,18 @@ void JoyResetPosition(ULONG nExecutedCycles)
 {
 	CpuCalcCycles(nExecutedCycles);
 
-	if(joyinfo[joytype[0]] == DEVICE_JOYSTICK)
+	bool isJoystick[2] = { false, false };
+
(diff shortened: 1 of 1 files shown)


tomcw:
Hi @audetto - here's a PR that integrates your code into the AppleWin mainline.

I want to give you a chance to comment before I merge to the mainline, in case you spot any issue.

audetto:
I won't be able to debug, but I don't see any issue.

What I am a bit concerned about is the following 287 thing.
Have you checked the 2 adjustments still work correctly together?

audetto:
Actually, it can only get better because `coeff>=1`

tomcw:
>Have you checked the 2 adjustments still work correctly together?

They are independent: the first one (square the circle) can give a value of 255 now for both x & y (when in the bottom right position). The the second adjustment (`if (pdlPos >= 255) pdlPos = 287`) will just set x & y both to 287. I tested against the games that need this 287 position, and all work fine (and better than before).

Anyway, it seems like this is fine from your POV, so I'll go ahead and merge it. Thanks.

PR #1249, opened 2023-08-28
Serial comms: compilation in Linux

audetto:
Some initial steps to use `SerialComms` in linux.
Current goal is to enable the TCP mode.


tomcw:
Tested against the 2 serial-TCP save-states in AppleWin-Test: `SSC-TCP-IN#2.aws.yaml` and `SSC-TCP-tx-rx-tcp.aws.yaml`
Merged to master.

PR #1250, opened 2023-08-30
WIN64: Use correct IReferenceClock definition from <strmif.h>

kiyolee:
See code comments for more details.

### DIFF
--- a/source/SoundCore.cpp
+++ b/source/SoundCore.cpp
@@ -646,7 +646,8 @@ void SoundCore_SetErrorMax(const int nErrorMax)
 
 //=============================================================================
 
-static DWORD g_dwAdviseToken;
+// Use DWORD_PTR according to IReferenceClock from <strmif.h>.
+static DWORD_PTR g_pdwAdviseCookie = 0; // Not really used as pointer.
 static IReferenceClock *g_pRefClock = NULL;
 static HANDLE g_hSemaphore = NULL;
 static bool g_bRefClockTimerActive = false;
@@ -713,7 +714,9 @@ void SysClk_StartTimerUsec(DWORD dwUsecPeriod)
 		return;
 	}
 
-	if (g_pRefClock->AdvisePeriodic(rtNow, rtPeriod, g_hSemaphore, &g_dwAdviseToken) != S_OK)
+	// IReferenceClock from <strmif.h> (origin <axcore.idl>) is "oddly" defined to use HSEMAPHORE.
+	static_assert(sizeof(HSEMAPHORE) == sizeof(HANDLE), "must be same size");
+	if (g_pRefClock->AdvisePeriodic(rtNow, rtPeriod, (HSEMAPHORE)g_hSemaphore, &g_pdwAdviseCookie) != S_OK)
 	{
 		fprintf(stderr, "Error creating timer\n");
 		_ASSERT(0);
@@ -729,7 +732,7 @@ void SysClk_StopTimer()
 	if(!g_bRefClockTimerActive)
 		return;
 
-	if (g_pRefClock->Unadvise(g_dwAdviseToken) != S_OK)
+	if (g_pRefClock->Unadvise(g_pdwAdviseCookie) != S_OK)
 	{
 		fprintf(stderr, "Error deleting timer\n");
 		_ASSERT(0);
--- a/source/StdAfx.h
+++ b/source/StdAfx.h
@@ -3,6 +3,11 @@
 #include <tchar.h>
 
 #include <crtdbg.h>
+// <strmif.h> has the correct IReferenceClock definition that works for both x86 and x64,
+// whereas the alternative definition in <dsound.h> is incorrect for x64. (out of
+// maintenance perhaps)
+// <strmif.h> *must* be included before <dsound.h> for x64 to work.
+#include <strmif.h>
 #include <dsound.h>
 #include <dshow.h>
 


audetto:
Do you have any link to a Microsoft doc about this?

inexorabletash:
Drive-by comment: The parameter type is defined as `DWORD_PTR` not `DWORD` 

https://learn.microsoft.com/en-us/windows/win32/api/strmif/nf-strmif-ireferenceclock-adviseperiodic

Why not just correct the type of `g_dwAdviseToken` to be `DWORD_PTR` ? If that type is not available in AppleWin for some reason (not pulling in all the windows headers), then defining a platform-specific type in one place seems like a more scoped/maintainable change. 

Casting the parameter of Unadvise() as introduced in this PR seems incorrect, as that also takes a `DWORD_PTR`. 

But again, just a drive-by comment. I use AppleWin occasionally, but don't contribute to the code. Good luck!

tomcw:
>Casting the parameter of `Unadvise()` as introduced in this PR seems incorrect, as that also takes a `DWORD_PTR`.

For `Unadvise()` the MS docs are wrong:
https://learn.microsoft.com/en-us/windows/win32/api/strmif/nf-strmif-ireferenceclock-unadvise

```
HRESULT Unadvise(
  [in] DWORD_PTR dwAdviseCookie
);
```

And here's the interface from dsound.h:

```
DECLARE_INTERFACE_(IReferenceClock, IUnknown)
{
    // IUnknown methods
    STDMETHOD(QueryInterface)       (THIS_ _In_ REFIID, _Outptr_ LPVOID*) PURE;
    STDMETHOD_(ULONG,AddRef)        (THIS) PURE;
    STDMETHOD_(ULONG,Release)       (THIS) PURE;

    // IReferenceClock methods
    STDMETHOD(GetTime)              (THIS_ _Out_ REFERENCE_TIME *pTime) PURE;
    STDMETHOD(AdviseTime)           (THIS_ REFERENCE_TIME rtBaseTime, REFERENCE_TIME rtStreamTime,
                                           HANDLE hEvent, _Out_ LPDWORD pdwAdviseCookie) PURE;
    STDMETHOD(AdvisePeriodic)       (THIS_ REFERENCE_TIME rtStartTime, REFERENCE_TIME rtPeriodTime,
                                           HANDLE hSemaphore, _Out_ LPDWORD pdwAdviseCookie) PURE;
    STDMETHOD(Unadvise)             (THIS_ DWORD dwAdviseCookie) PURE;
};
```

Note `Unadvise()` takes a DWORD, not a DWORD_PTR.


But as @audetto says above about the original PR comment:
>Do you have any link to a Microsoft doc about this?

kiyolee:
No. Can't find any doc about the bug.
But I can see that in debugger, 64-bit build was crashing that `g_pRefClock` became corrupted after `AdvisePeriodic()` returned.
I did the experiment defining `g_dwAdviseCookie` as `DWORD[2]` and set to `{-1,-1}` before calling `AdvisePeriodic()`, both `DWORD` would be assigned. If the API is not buggy, `DWORD[1]` should not be touched.
I think either that `unadvise` doc from Microsoft was wrong or the header was wrong.
The header does imply `AdviseCookie` is `DWORD`, and in that case, I very much suspect the 64-bit implementation has a bug assigning the quantity as 64-bit.
If `AdviseCookie` is meant to be `DWORD_PTR`, the bug is then not a bug, `DWORD_PTR` is indeed 64-bit for x64, 32-bit for x86.
But if you check the value returned in `g_dwAdviseCookie`, the value does not look like any pointer. It is simply numeric value like 1 or 2.

audetto:
There is at least another example in github

https://github.com/Harlinn/Harlinn.Windows/blob/59f1a42f273f80860fe9b79ef1d28ecd2b124364/Harlinn.Windows/HWStrmif.h#L437

Is it relevant?

kiyolee:
At least that's an consistent implementation.

tomcw:
dsound.lib (which I assume is where this `IReferenceClock` API lives) is a 32-bit library, so there must be some shim code to go from x64 to 32-bit. Hmm... actually dsound.lib is only 4KB in size(!) so it's not this.

Stepping into `AdvisePeriodic()` from a regular 32-bit build of AppleWin:
```
	quartz.dll!CAMSchedule::AddAdvisePacket(__int64 const &,__int64 const &,void *,int)	Unknown
 	quartz.dll!CBaseReferenceClock::AdvisePeriodic(__int64,__int64,unsigned long,unsigned long *)	Unknown
 	AppleWin.exe!SysClk_StartTimerUsec(unsigned long dwUsecPeriod) Line 716	C++
 	AppleWin.exe!ContinueExecution() Line 210	C++
```

quartz.dll is the DirectShow runtime.

I have quartz.dll in C:\Windows\System32\ and C:\Windows\SysWOW64\
And many (44!) of these, eg. in C:\Windows\winsxs\amd64_microsoft-windows-directshow-core_31bf3856ad364e35_6.1.7600.16385_none_0265298807973233
(Where [SxS](https://en.wikipedia.org/wiki/Side-by-side_assembly) / side-by-side is to avoid DLL Versioning hell)

It's [WoW64](https://en.wikipedia.org/wiki/WoW64) that allows running 32-bit apps on a 64-bit system.

Perhaps an idea is to create a simple x64 project that just calls this API (eg. `AdvisePeriodic()`), then we can step at the assembly level to see what's going wrong.

kiyolee:
You can just build a 64-bit version of AppleWin from my repo.
I can readily debug into that particular spot. That's how I notice the problem and come up a possible solution.

kiyolee:
Just debugged into that spot.
VS does show this signature in stack:
```
>	quartz.dll!CBaseReferenceClock::AdvisePeriodic(__int64,__int64,unsigned __int64,unsigned __int64 *)	Unknown
```
So it clearly shows that it is expecting a pointer to some 64-bit quantity and that is certainly different from what `dsound.h` says.

kiyolee:
Note that `dsound.lib` being small is not of much significance. lib files for dll are usually small as they are just entry points reference (redirection). This kind of lib files pretty much contains no read code and hence small.

tomcw:
And pulling the evidence together...

From my 32-bit callstack:
quartz.dll!CBaseReferenceClock::AdvisePeriodic(__int64,__int64,unsigned long,unsigned long *)

From your 64-bit callstack:
quartz.dll!CBaseReferenceClock::AdvisePeriodic(__int64,__int64,unsigned __int64,unsigned __int64 *)

kiyolee:
I think now I see where the confusion comes from.
Apparently `IReferenceClock` is defined differently in `dsound.h` (using `DWORD` as `AdviseCookie`) and `strmif.h` (using `DWORD_PTR` as `AdviseCookie`).
Behind the scene, I suspect for 64-bit `IReferenceClock` is meant to be the one from `strmif.h`.
But current code only include `dsound.h` that leads to wrong definition of `IReferenceClock` for x64.

kiyolee:
https://github.com/kiyolee/applewin/commit/437810262036ec13e10f75b91684a98a05229f89
This is an alternative fix that seems to work.
By including `strmif.h` before `dsound.h` for x64, the definition of `IReferenceClock` is then correct.
This fix looks a bit cleaner and less hacky, except the `HSEMAPHORE` casting.
If this is prefered, I can update this PR to use the above commit.

kiyolee:
Or we could always include `strmif.h` for both x86 and x64 and use DWORD_PTR for both builds. Might be simpler.

kiyolee:
https://github.com/kiyolee/applewin/commit/cfb05b8a52e3ef146c3e49ba0c0748f91293b88a
Including `strmif.h` does seem to work for both x86 and x64 and the change can be indeed simpler. See above for reference.
Of course there needs more comments to explain the problem.
Dig a bit more at the ms doc site, there seems no more reference of `dsound.h` (DirectSound), and `strmif.h` (DirectShow) is the only place talking about `IReferenceClock`.
`strmif.h` does change a little amoungst various versions of Windows SDK, little update shows up in SDK for Windows 11.
Whereas `dsound.h` has not been changed at all for a very long time.
That implies `strmif.h` is more current and can be relied on. That matches what ms doc site is showing.
But the use of `HSEMAPHORE` is a bit peculiar.
`HSEMAPHORE` comes from `axcore.idl` where the definition of `IReferenceClock` comes from.
Use of `HSEMAPHORE` seems pretty uncommon amongst the Windows SDKs. (It appears in `winddi.h` which flows back into one of these DirectThing.)
I believe `HANDLE` is what we use for semaphore since WinNT.
If I understand the SDK headers correctly, `HSEMAPHORE` is legacy from Win95 or before. (I vaguely remember OS/2 had that too.)
So the summary of this story is that `dsound.h` is having the wrong `IReferenceClock` definition for x64 for years (decades?) whereas `strmif.h` has the correct definition for both x86 and x64 but referring to legacy (very old) Windows API definition of `HSEMAPHORE`.
Conclusion: I will update this PR to use `strmif.h` directly unless somebody objects.

kiyolee:
@tomcw Updated the PR to use `IReferenceClock` definition from `strmif.h` instead of `dsound.h`. Please have a look.

tomcw:
Thanks - Merged into master branch.

PR #1269, opened 2024-01-04
Uthernet2: use fcntl for better compatibility with MacOS.

audetto:
This PR changes the way a socket is marked as "non-blocking" in linux.

The call to `fcntl` is more portable and available in macOS too.

Linux offers an easier alternative (an argument to `socket`), which does not exist in macOS.
The final result is the same, with no special `#ifdef` required for macOS (which at the moment does not work).

At the same time, an error is reported on all platform if the "non-blocking" flag fails.

### DIFF
--- a/source/Uthernet2.cpp
+++ b/source/Uthernet2.cpp
@@ -72,17 +72,10 @@ typedef int socklen_t;
 #include <arpa/inet.h>
 #include <poll.h>
 #include <unistd.h>
+#include <fcntl.h>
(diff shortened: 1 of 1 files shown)


tomcw:
Sorry for the delay. I will try to verify this soon.

tomcw:
I tested this using A2Stream 1.4 with "Virtual DNS" enabled in AppleWin (over WiFi on my PC).

PR #1270, opened 2024-01-04
Minor bugs in SSI263

audetto:
Fix an assertion that would never trigger.
Avoid assertion (or segfault) when running with `-no-mb`.


tomcw:
Thanks. Merged it.

Issue #1264, opened 2023-12-19
Harddisc driver does not report size back in X/Y on ProDOS status call


peteri:
The ProDOS driver for the HD doesn't report back the block size in the X (low) and Y (high) registers on a ProDOS status call. This would need some modification to the HDD firmware 6502 code to allow for this (if the c++ code just tweaked the registers directly).

Also there does not seem to be a mechanism for having a hard disc (or smartdrv) file that does not auto grow to (although maybe I missed this)

Looking at the other ProDOS disk driver issues there does seem to be some appetitie for rewriting this code, so I thought I'd raise this first.

I noticed this as an issue when trying to get the Cirtech CP/M plus code running in AppleWin, I can certainly change the loader BIOS for CP/M to use the value of from the slot ROM if the status call returns a value that isn't $28 and the slot ROM isn't $0000 as per the ProDOS 2.4.2 disk driver documentation.

Note the size at $CxFC seems to be wrong for a 32Mb image? Formatting a new disk with ProDOS 2.4.2 shows the size as being 32767 blocks (which is 16MB) although the code happily works with a 32MB image (i.e. Total Replay shows the disk size as 65535 blocks)

Are you open to a patch for these issues? Or is there a plan to add better support for SmartDrv with all of the extended calls?

peteri:
I can supply some disk images for CP/M bootable in AppleWin if you'd like and the source code is available. (The BIOS supports both SmartDRV & ProDOS)

tomcw:
Hi @peteri, I'm without a good connection at the moment. Let me reply properly in a few days' time.

In general supplying a patch is fine, but it would be good for you to read the notes on submitting AppleWin pull requests (the link is off the main page), and also for you to define the "spec" of your patch first.

peteri:
Okay, I think I need to have a good hard look at the code to be honest along with some of the other issues raised, make a plan then share it. 

My thirty year old ProDOS 1.0 programmers workbench doesn't even mention returning back stuff in X/Y for the status call (it is definitely very sparse).

The other annoyance is that a lot of the Apple Technotes say "removed please see the //c or IIGS tech ref" which misses out a lot of context for older programs that might rely on a specific behaviour.

I'll see if I get something together for the start of 2024..

burniouf:
https://archive.org/download/Apple_ProDOS_8_Technical_Reference_Manual/Apple_ProDOS_8_Technical_Reference_Manual.pdf

[...]

Issue #1033, opened 2022-02-08
Hacked AppleWin for slot 5 boot (for pascal) Help?


dhlav:
Hi, I'm quite new to all the github stuff, but I have checked out the latest and modified it so it sets the hard drive to slot 5 instead of slot 7. Used this for pascal setup/booting (pascal wont boot slot 7). Works good in testing here. I'd like to at least contribute this hacked version back maybe as a branch or whatever. Can anyone give me a clue how to get it back up? Pull request? I might need a simple step by step. Don't want to hurt the repo. Thanks. Dave Hlavac.

audetto:
Fork the repo, and create a branch with your new code.
If and when you think it is clean enough, create a PR to merge it here.

dhlav:
Ok, many thanks. FYI, this is just a hack job substituting 5 for 7, and you can't have disk card ii on 5 either. Just to get me going and a basis for others, if needed. I've already forked it. Haven't pushed anything to my repo yet. I will try and set this up and do it the ~proper~ way with a config as noted via cmd line to start. Shouldn't be too crazy. Someone already set the hd firmware to be slot independent. One thing to note is the hd firmware card has a fixed 32767 reported block count. So formatting under pascal always makes the size 32767. If you format/use a different size outside of applewin/emulated pascal formatter, the images work and show the size correctly. Just an fyi.

tomcw:
Probably this needs adding for command line support:
- -s5 hdc
- -s7 hdc
- -s5h1 \<pathname\>
- -s5h2 \<pathname\>

where 'hdc' = hard disk controller.

btw, it was me that made the hd firmware slot independent (#996), with the intention of adding this slot 5 support at some stage.

>One thing to note is the hd firmware card has a fixed 32767 reported block count. So formatting under pascal always makes the size 32767. If you format/use a different size outside of applewin/emulated pascal formatter, the images work and show the size correctly.

How else would this be reported? IE, yes, the firmware has a fixed 32767, but what are alternative ways to support this?

dhlav:
It's just in the pascal formatter, so somehow it's reading the disk size as that (probably from unitstatus/status). I'll check and see if it can be worked around/supported, but since only the first n blocks of an image is all that's needed for pascal, can probably make images externally. You can take the 800k pascal 3.5 floppy image and works/reports just fine, until you format it. :)

tomcw:
[...]

PR #1272, opened 2024-01-14
ProDOS HDD Driver return size in X/Y for status call

peteri:
ProDOS HDD Driver modified to support returning the volume size in the X/Y registers. 
New command line option to force a size for autoexpanding use.
Fixes issue #1264 and #1033 for Pascal 1.3


tomcw:
Thanks @peteri - I will try to take a look in the coming days.

tomcw:
I have a concern about the HDD firmware and backwards compatiblity:

1. Is it back-compat? ie. the 16-bit value at $C7FC (number of blocks in device)
   - Some s/w must be reading this value to show the device size.
   - Or perhaps all the s/w that reads this value knows that if it's 0x0000 then it needs to issue a STATUS command?

3. Saving the state when the PC=$C7nn (using AppleWin built with the old firmware) and then loading the state (using AppleWin with the new firmware) could result in undefined behaviour, as the running code may've changed.
   - NB. there is some detection for this case in `HarddiskInterfaceCard::LoadSnapshot()`
   - So we could just bump `kUNIT_VERSION` to 4, and the version check in `HarddiskInterfaceCard::LoadSnapshot()` to 3, to throw the error "HDD card: 6502 is running old HDD firmware".
   - But this isn't a great solution, as it prevents some older save-states from being loaded (even if the chance of a snapshot occurring when the 6502 is executing the HDD firmware is small)
   - to do (for me): consider persisting the HDD controller firmware to the state file? (And not just for slot 7, but for other slots that use this card.)

My preference is to retain the current HDD firmware in AppleWin as a fallback measure (eg. selectable via a command line option).

Can you put your new firmware into a new folder "Firmware/HDDv2".
- In Applewin.rc: add a new line: `IDR_HDDRVR_V2_FW           FIRMWARE                "Hddrvr-v2.bin"`
- In resource.h: add a new `#define IDR_HDDRVR_V2_FW 155` (just add to the end after IDR_BASE64A_VIDEO_ROM)
- And change AppleWin to use this new HDD firmware
- In Harddisk.cpp: bump `kUNIT_VERSION` to 4 (but leave the version check in `HarddiskInterfaceCard::LoadSnapshot()` at 2).
- I will make the changes to AppleWin to allow using the old firmware - so no need to update your PR to do this.

tomcw:
A minor thing about the cmd line switch: "-harddisknumblocks"

Can you add a setter function to Harddisk.h, eg. `void SetUserNumBlocks(UINT numBlocks)` and a new private member `m_userNumBlocks`.
This way, the `HarddiskInterfaceCard` class doesn't need to include CmdLine.h, and doesn't need to access `g_cmdLine`.

Then you'll need to query every slot if it's of type `CT_GenericHDD` - so just extend this loop to do this too:
https://github.com/AppleWin/AppleWin/blob/fcd216bb45e7640bbd3c0c66758bc48e65a38195/source/Windows/AppleWin.cpp#L801

Note the Config UI also allows HDC cards to be inserted.
So in `CPropertySheetHelper::SetSlot()` for cards of type `CT_GenericHDD`, we should also call this setter function with the cmd line specified value.

peteri:
Thanks for the comments, I hadn't thought about the save state so I'll get the points you've raised cleaned up.

The block size being zero does mean that code should call the status to get the size and this is the functionality in the original Profile hard disk ROM from Feb 1984 (see page 43 in https://www.bitsavers.org/pdf/apple/disk/profile/appleII_interface/AII_Profile_Boot_Prom_198402.pdf ) so I'd expect most software to make the call.

tomcw:
> ...so I'd expect most software to make the call.

Set a watchpoint on $C7FC-FD with `bpm c7fc,2`

#### copy_ii+_v9.1_s1.dsk:
- "Y" / Verify option: breaks at 91CD
  - if (C7FC).w=0x0000 then: LDA #0; JSR A1F9
  - function at 91CB returns num blocks in Y:X

```
91CB:A0 FC                LDY #$FC
91CD:B1 E9                LDA (HGR.SH+1),Y
91CF:AA                   TAX
91D0:C8                   INY
91D1:B1 E9                LDA (HGR.SH+1),Y
91D3:A8                   TAY
91D4:D0 09                BNE $91DF
91D6:E0 00                CPX #$00
91D8:D0 05                BNE $91DF
91DA:A9 00                LDA #$00
91DC:20 F9 A1             JSR $A1F9
91DF:18                   CLC
91E0:60                   RTS
```

So yes, this software from 1990 does make the call.

peteri:
I've just had a quick look at the oldest version of ProDOS I can find online (1.0 Nov 83) and it has a similar size check in `FILER` so I'm happy (at $4250-$429c).

peteri:
Just a quick comment to say I've not forgotten about this, just been a bit busy at work. I'm hoping to get the suggestions incorporated over the weekend.

peteri:
Okay, got a bit of time.... 
I'm not entirely convinced the change to CPropertySheetHelper::SetSlot is needed as adding or removing the controller causes a restart of AppleWin, is there an edge case I'm not aware of?

tomcw:
This PR looks good.

>I'm not entirely convinced the change to CPropertySheetHelper::SetSlot is needed as adding or removing the controller causes a restart of AppleWin,

Yes, you're I right - so that code in `SetSlot()` can go.

Thanks.

peteri:
Reverted the change to the PropertySheetHelper....

Thanks for being so patient with my changes.

tomcw:
My notes:

1. When growing an image (ie. appending a block to the end), then `CImageBase::WriteBlock()` updates the `ImageInfo` object with the new `uImageSize`. So subsequent calls to `GetImageSizeInBlocks()` will return the _new_ size (unless `m_userNumBlocks` has been set).

2. `GetImageSizeInBlocks()` returns `numberOfBlocks` by just dividing by `HD_BLOCK_SIZE` (512), when perhaps it should round up (ie. add 511) before dividing. Unlikely to be an issue since the image size should (must?) always be a multiple of 512.

Neither are a concern, but just making a note here.

tomcw:
@peteri - thanks for your PR. It's now merged into the main line.

tomcw:
@peteri - here's a new [AppleWin 1.30.18.0](https://github.com/AppleWin/AppleWin/releases/tag/v1.30.18.0) release that includes your PR work.

Issue #1285, opened 2024-03-29
CPU emulation: AppleWin crashes when doing a 16-bit read from 6502 address 0xFFFF


tomcw:
Repro from debugger:
- out c083  ; set LC to HIGH RAM mode
- fffe: cd 00 20 ; CMP $2000
- 2000: ea
- r a=ea
- r pc=fffe
- \<SPACE\> to step

This will cause AppleWin to crash! ...because `mem+64K` is unmapped.
(Correct result is to set Z=1, and not crash :-)

A simple fix for this is to `mmap` (or the equivalent under Windows) so that virtual addresses `mem` and `mem+64K` both point to the same physical 64K "page".

tomcw:
In PR #1286, I've gone for the "poor workaround" which will work on pre-Windows 10 OS's.

audetto:
Disclaimer: I have not debugged anything.

I was thinking at the implications of this and how to implement it in a system without sophisticated `mmap` abilities.

1. Single byte reads can be solved with `& $FFFF`
2. Multi byte could be turned into many single byte reads
3. BTW, do they wrap only at `$FFFF` or at every `$FF`?
4. How does this interact with the idea from https://github.com/AppleWin/AppleWin/issues/795#issue-630380206

If multi byte reads are allowed and don't wrap at every page, the idea of multiple small physical pages and an extra indirection is dead.

So the question for me is: should multi byte reads be allowed?

tomcw:
>2. Multi byte could be turned into many single byte reads

That's correct, but the issues are:
a) having to rewrite some of the 6502 emulation code to switch from a 16-bit read to 2x 8-bit reads (+ write tests for all this)
b) expect some performance loss

So it's a lot of work _just_ for this `$FFFF` edge-case (which in 20+ years nobody has ever raised).

The great thing about this "circular buffer" solution is that there's no need to do (a) or be impacted by (b).
Therefore it's a cheap solution with zero impact on existing code.

>3. BTW, do they wrap only at `$FFFF` or at every `$FF`?

Just at `$FFFF`.

>4. How does this interact with the idea from https://github.com/AppleWin/AppleWin/issues/795#issue-630380206

That's another idea...

In addition to your point-2 (turn multi-byte into many single-byte reads) this idea is about doing all byte reads through page pointers (just as AppleWin currently does for byte writes). NB. currently AppleWin maintains a flat, contiguous 64KiB `mem` buffer for any read (except for the I/O area at $Cnnn).

[...]

PR #1286, opened 2024-03-29
Fix for GH1285: 2x adjacent 64K pages mapped to same physical page

tomcw:
Fix for #1285

### DIFF
--- a/source/Memory.cpp
+++ b/source/Memory.cpp
@@ -1537,7 +1537,7 @@ static LPBYTE AllocMemImage(void)
 
 	// Allocate memory for 'memimage' (and the alias 'mem')
 	// . Setup so we have 2 consecutive virtual 64K pages pointing to the same physical 64K page.
-	// . This is a fix (and optimisation) for 6502 opcodes that do a 16-bit at 6502 address $FFFF. (GH#1285)
+	// . This is a fix (and optimisation) for 6502 opcodes that do a 16-bit read at 6502 address $FFFF. (GH#1285)
 	SYSTEM_INFO info;
 	GetSystemInfo(&info);
 	bool res = info.dwAllocationGranularity <= _6502_MEM_LEN;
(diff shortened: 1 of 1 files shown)


tomcw:
@audetto - this is very much a Windows-only implementation.

I will wrap this in a `#ifdef _MSC_VER` so it's only done under Windows, but it'd be good to get a Linux equivalent, eg. using `mmap`.

audetto:
I think this will do for the linux version

https://abhinavag.medium.com/a-fast-circular-ring-buffer-4d102ef4d4a3

source/Memory.cpp
```diff
@@ -1509,12 +1509,106 @@ bool MemIsAddrCodeMemory(const USHORT addr)
...
+static LPBYTE AllocMemImage(void)
+{
+	LPBYTE baseAddr = NULL;
+
+	// Allocate memory for 'memimage' (and the alias 'mem')
+	// . Setup so we have 2 consecutive virtual 64K pages pointing to the same physical 64K page.
+	// . This is a fix (and optimisation) for 6502 opcodes that do a 16-bit at 6502 address $FFFF. (GH#1285)
+	SYSTEM_INFO info;
+	GetSystemInfo(&info);
+	bool res = info.dwAllocationGranularity <= _6502_MEM_LEN;
```
sicklittlemonkey, line 1547: Very "clever" code. ; - )
Might as well be explicit here with info.dwAllocationGranularity == _6502_MEM_LEN, as I think < would fail.

tomcw: OK, I've made that change.

tomcw:
@sicklittlemonkey - it's a shame that the page alloc granularity on Windows is 64KiB and not 4KiB, otherwise we could probably do some neat things with the Language Card bank switching and not have to memcpy() regions around.

source/Memory.cpp
```diff
@@ -242,7 +241,12 @@ static LPBYTE	RWpages[kMaxExMemoryBanks];		// pointers to RW memory banks
 static const UINT kNumAnnunciators = 4;
 static bool g_Annunciator[kNumAnnunciators] = {};
 
+#ifdef _MSC_VER
+static HANDLE g_hMemImage = NULL;	// NB. When not initialised, this handle is NULL (not INVALID_HANDLE_VALUE)
+#endif
+
 BYTE __stdcall IO_Annunciator(WORD programcounter, WORD address, BYTE write, BYTE value, ULONG nCycles);
+void FreeMemImage(void);
```
audetto, line 249: this fwd declaration needs `static`

tomcw: Thanks - I've added this.

sicklittlemonkey:
> @sicklittlemonkey - it's a shame that the page alloc granularity on Windows is 64KiB and not 4KiB, otherwise we could probably do some neat things with the Language Card bank switching and not have to memcpy() regions around.

@tomcw: Good point. You probably already knew this, but it's been a long time since I wondered about something that low-level on a modern platform, and I assumed the reason would be "Any change might break something" (e.g. your code, and this probably *is* a reason) but the [historical reason](https://devblogs.microsoft.com/oldnewthing/20031008-00/?p=42223) is interesting. One can usually count on the indefatigable Raymond Chen. ([Bonus info](https://devblogs.microsoft.com/oldnewthing/20210510-00/?p=105200).)

tomcw:
Ha! Really interesting articles - thanks for digging those out & sharing!

PR #1287, opened 2024-03-30
Implement mmap circular buffer for linux.

audetto:


tomcw:
Thanks @audetto.

Since your PR is a PR on top of mine, then how do I merge this into the main line?
EG. do I first merge my PR (#1286), then secondly merge this one?

Also I thought you abstracted the Linux code away in your fork of AppleWin (for Linux) - ie. I know you did this for Win32 API functions so that I didn't need to change any of the main line AppleWin code. Should we move both of these OS-specific functions into dedicated OS (or platform) specific .cpp files?

audetto:
You could merge this first, which will modify your branch, and when you merge your PR they will both land in master. Or I can re raise another one after you have merged yours.

The structure of the linux code is very simple and I would prefer to avoid the retry and loops necessary in windows.

tomcw:
>You could merge this first, which will modify your branch, and when you merge your PR they will both land in master. 

Thanks - I took this option.

>The structure of the linux code is very simple and I would prefer to avoid the retry and loops necessary in windows.

I agree, but perhaps we should have a Windows-specific .cpp file (to include this and other Windows-specific code), and similarly one for Linux?
So in this case, the functions `AllocMemImage()` and `FreeMemImage()` are relocated into their respective Windows / Linux .cpp files.

PR #1309, opened 2024-06-01
Fix for corrupt 't' glyph

xotmatrix:
This fixes a missing ascender pixel in the 't' glyph used for Apple //e emulation and the debugger.


xotmatrix:
I had some difficulty creating the `CHARSET4.BMP` in a format compatible with the emulator. Paint.NET made quite a mess but Aesprite worked like a charm. I made one manual change in the file header afterward to more closely match the original. The original does not include resolution (pixels/meter) information. I removed it from the new file to match. The only differences now are the 12 changed pixels. Likewise the binary content of `Debug_Font.bmp` differs by only one RGB pixel.

tomcw:
As you say, we should use the original video rom, and not rely on a bitmap.
I will accept your PR, since the `Debug_Font.bmp` contains extra chars that don't exist in the video rom.

tomcw:
Anyway, thanks for submitting this PR!

Issue #1207, opened 2023-04-03
Add support for Smartport hard disk controller


tomcw:
Spun off from #1186.

@burniouf asked:

wouldn't it be possible to add a way to select "Block Mode" $Cn07 = $3C with 2 devices so that Apple ][+ could boot and "Smartport Mode" $Cn07 = $00 with more than 2 devices so that Slot 7 being seen as a SmartPort device by ProDOS and Enh //e could boot ?

tomcw:
In #1201, @burniouf gave these Smartport references:

Apple IIgs Firmware Reference manual Chapter 7 (with extended smarport calls $4x) :

https://mirrors.apple2.org.za/Apple%20II%20Documentation%20Project/Computers/Apple%20II/Apple%20IIGS/Manuals/Apple%20IIGS%20Firmware%20Reference.pdf

Apple //c reference manual (2nd edition) Chapter 6 :

https://mirrors.apple2.org.za/Apple%20II%20Documentation%20Project/Computers/Apple%20II/Apple%20IIc/Manuals/Apple%20IIc%20Technical%20Reference%20Manual.pdf

Also, Apple //c firmware listing will tell you whats beyond Smartport API

Apple SCSI card ROM listing (extra call feature) may also show you how was supposed to be handled targeting CHAR devices on SCSI bus like Tape drive....

tomcw:
The current AppleWin hard disk controller (hdc) firmware ([here](https://github.com/AppleWin/AppleWin/blob/master/firmware/HDD/hddrvr.a65)) is mainly ProDOS compliant (for II, II+, //e, Enh //e), and supports 2 units per hdc.

It has a Smartport entrypoint at $Cs0D, but since the ID bytes don't identify as a Smartport (ie. $Cs07 = $3C, not $00), then I guess this Smartport entrypoint (and associated support) is redundant.

AppleWin could just auto-switch $Cs07=$00 for an Enh //e, but as @burniouf says, the firmware shouldn't do this because:
>About switching from Cn07 = $3C <-> $00....should be more an option...putting a $00 device in a ][+ is absolutely possible in real hardawre.....the only difference is that you cannot autoboot on it....

tomcw:
Probably best to write a new hdc-smartport firmware for the Enh //e (and newer //c?) models.
- but switchable so that it can be tested against older models.
- (and a new firmware means that the old cruft can be dropped, eg. old DOSMaster entrypoint)

[...]

PR #1314, opened 2024-06-15
Support HDC SmartPort firmware

tomcw:
(From #1207)


tomcw:
For the SmartPort status command, I'm using a f/w version of 130.18 (for AppleWin 1.30.18.x), eg:

STABLE 32MB

Also the Device ID string uses the last 2 chars to show the decimal device number (so anything over 99 won't fit).
(Other options are encode as hex, or shorten the base part of the string: "AppleWin SP D#")

tomcw:
Using command line: -s7h3 "image3.hdv" -s7h4 "image4.hdv"

4 HDDs

PR #1327, opened 2024-09-21
Added support for Hebrew charset

gmegidish:


gmegidish:
Realized I can support switching between main charset and alt charset, to allow support of lowercase english.

PR #1328, opened 2024-09-21
Add support for Hebrew charset

gmegidish:
This pull-request adds support for Hebrew character generator for Apple IIe clone.

Israeli Apple IIe had a custom char-gen rom with a toggle to switch between original Apple IIe with lower-case, and the custom hebrew one.

I have implemented this as a Clone called "Hebrew", and it's using the original Apple IIe rom. 

I've seen roms that are exactly like Apple's, but change "Apple ][" banner to a store name.

Keyboard mapper also has been implemented, which maps the ascii lowercase keyboard to the hebrew layout.


xotmatrix:
It is possible to use this character set using the `-videorom` command line option with the 8K Video ROM you uploaded to archive.org. I'm not familiar with the Hebrew variant and don't know if there are other changes necessary.
https://archive.org/details/AppleIIeHebrewVideoROM

gmegidish:
That's correct. -videorom will get you the font (not sure about the alt-f10 toggle), but a hebrew keyboard has a layout. So a "a" needs to be translated to "y", but uppercase stays the same. Uppercase A stays the same. See keyboard.cpp for the mapping. Is keyboard mapping possible through command-line?

tomcw:
@gmegidish - it sounds like this is a stock European //e (ie. which includes the video ROM rocker switch) and only the video ROM has been replaced with this Hebrew video ROM. So this isn't a clone per se.

And as you state, sometimes the computer ROM was changed:
>I've seen roms that are exactly like Apple's, but change "Apple ][" banner to a store name.

Do you know which company manufactured it? And how widely used was it?

Ideally AppleWin would have a way to configure the video ROM (ie. country: Canada, France, Germany, Spain, UK, US, etc) from the main UI, and then this Hebrew video Rom could just be added there as "Israel". Let me consider this approach.

gmegidish:
@tomcw A person I know dumped one of the Israeli clones that were going around. I myself grew up with a clone apple2e with Hebrew charset toggle. I have these roms and I can incorporate these as clone.

What do you prefer?
1. Implement hebrew as a keyboard layout. I think this feature is missing, but with -videorom and -keyboard mapping, you could achieve the same functionality.
2. Identify the Israeli clone used and use its rom with this pull request?

I can do either. We just uncovered some Israeli floppies (educational). Up until now, I don't think anybody in the Israeli Apple2 community had floppies that would require keyboard layout and charset.

Thanks!

tomcw:
@gmegidish - you are right, the keyboard mapping shows that it was more than just a change to the video rom.

If there are other rom(s) that are different, eg. the 16K Apple //e rom ($C100 - $FFFF) then that can tip the balance in favour of this being a clone, and therefore implemented as such.

gmegidish:
@tomcw You're right. Let's sack this pull-request, and I'll create another pull-request instead that along with -videorom, you can get hebrew support. There were a few clones in Israel in 80s, I grew up with one, but I don't think it was any different than Apple II with 48KB and a custom ROM that replaced the "Apple ][" logo on boot. I'm closing this pull-request and I'll create another one with keyboard mapping command line parameter.

Issue #105, opened 2014-05-23
Videx 80 column card in slot 3 for Apple][ and ][+


tomcw:
**BerliOS Feature Request #4880**

**Date:** 2009-November-13 22:40
**Submitted By:** tomch
**Category:** General Functionality
**Assigned To:** tomch
**Priority:** 5
# 

Date: 2009-Nov-13 22:40
Sender: tomch
Logged In: YES 
user_id=25188
Browser: Mozilla/5.0 (Windows; U; Windows NT 5.1; en-GB; rv:1.9.0.7) Gecko/2009021910
Firefox/3.0.7 (.NET CLR 3.5.30729)

Requested by Stanislav Georgiev (3/11/2009)

tomcw:
**Email from Ju-Hyun Lee:** (14 Feb 2015)

Today, I would like to ask one thing if you or someone who is developing AppleWin these days. It is about Videx video card support on Apple II+ mode. I know IIe has 80 column mode and it is the standard, and videx is not the standard. However, in Korea in my age people, we didn't have formal import of Apple II series from America so that most of 8bit PC kids used the illegal copy of Apple II+ together with Videx & CP/M (maybe they were the copied ones) to learn about computers. I was one of them. I enjoyed cp/m learning a lot of computer related stuff although I was not the computer major student: learned a lot of programming languages (PL/I, Forth, Fortran, Lisp, Pascal, C, Algol, etc.) and many useful software (dBase II, Word Star, Multiplan, etc.). I think my base of computing knowledge was almost established with Apple II. Only one thing I missed with CP/M environment compared to Apple Soft Basic was graphics. I knew I could use GBASIC (by Microsoft) can handle couple of Apple Basic's graphic commands. However, in CP/M, I moved to Turbo Pascal (V3.0) as my major programming language so that I didn't want to use interpreter based programming any more.

Couple days ago, I visited Korean Apple community and found an old computer magazine's article (Korean). It is about Apple II CP/M Turbo Pascal Graphics Toolbox. One Korean programmer made the library, implementing Apple Soft Basic's graphic functions by calling the Basic interpreter's routine by inline assembly. I tried this in AppleWin. LoRes functions work greatly. I was very surprised that I learned this graphics capability of Apple II CP/M Turbo Pascal, and was really happy with the demo program. However, the hires demo didn't work. I tried even MESS. Still not working.

[...]

PR #1338, opened 2024-11-08
Initial version for supporting the IMC-2001 (Taiwanese Apple ][ clone)

andreasbaumann:
* a new entry in the list of clones and a dedicated RUN button
* has differences in IMC2001.rom to support Apple ][e like lowercase characters in the monitor ROM
* the video ROM for 40-character more IMC2001-Video.rom is quite different (mainly to support lowercase ASCII characters in NORMAL mode)
* the keybaord act more like a Apple][e, decided to use the CAPS mode switch as for the Apple][e in the emulator
* TODO: support for special keyboard features like the separate HOME, DEL, UP, DOWN, LEFT, RIGHT block of keys. Also F1 to F12 keys which are hard-coded in the keyboard ROM (with a switch for APPLE and CP/M mode)


andreasbaumann:
Some pictures, documentation and background information is on https://www.andreasbaumann.cc/hardware/imc-2001/

source/Keyboard.cpp
```diff
@@ -223,6 +223,14 @@ void KeybQueueKeypress (WPARAM key, Keystroke_e bASCII)
 		if (key >= VK_LEFT && key <= VK_DELETE)
 		{
 			UINT model = IsCopamBase64A(GetApple2Type()) ? 2 : IS_APPLE2 ? 0 : 1;
+			if (IsIMC2001(GetApple2Type()))
+			{
+				// TODO: the IMC-2001 actually has black LEFT and RIGTH arrow keys which behave as
+				// on an Apple IIe, but it also has light arrow keys UP, DOWN, LEFT, RIGHT, which
+				// actually position the cursor on the screen, not sure how to emulate this.
+				// for now just doing the black LEFT and RIGHT keys
```
tomcw, line 231: Typo: "RIGTH"
Also the sentence is incomplete "which behave as ? on an Apple IIe"

source/Keyboard.cpp
```diff
@@ -450,7 +458,7 @@ BYTE KeybReadFlag (void)
 //===========================================================================
 void KeybToggleCapsLock ()
 {
-	if (!IS_APPLE2)
+	if (!IS_APPLE2 ||IsIMC2001(g_Apple2Type))
```
tomcw, line 461: Missing a space after "||"

source/NTSC_CharSet.cpp
```diff
@@ -209,6 +215,12 @@ static void userVideoRom2K(csbits_t csbits, const BYTE* pVideoRom, const eApple2
 					if (!(n & 0x01))
 						n = n ^ 0xfe;
 				}
+				else if (type == A2TYPE_IMC2001)
+				{
+					// for now just unconditionally negate the pattern for inverse (does
+					// this work for small letters too?)
```
tomcw, line 221: formatting: move the "(does" onto the 2nd comment line

tomcw (changes requested):
I've eye-balled this PR, and there are a few comments regarding typos & formatting that I'd like addressed.

tomcw:
Can you also add new info for the IMC-2001 in the "Clone" section in the Help html file:
https://github.com/AppleWin/AppleWin/blob/master/help/cfg-advanced.html

source/NTSC_CharSet.cpp
andreasbaumann: Hi, sorry about the typos.

tomcw: Thanks!

tomcw:
Hi @andreasbaumann,
Thanks for submitting this PR for this Apple II clone.

There are outstanding TODO items that you have identified. Do you plan to tackle any of these? Or do you think it's "good enough" to merge this PR into the main line now?

andreasbaumann:
It's usable as it is now. The trouble currently is, that the original machine is in mid-
restauration (every time I power it on, some new weird stuff happens). So, it's not
that easy to verify that everything is now the same as in the emulator. :-)

I could verify a few things:
- the inverse, flash modes and also the character under the cursor are wrong also
  on the original if you write small characters (it works in normal mode only), so I'll
  document accordingly and remove the TODO as this is the behaviour of the
  IMC-2001.
- the free floating cursor I can tackle, would be neat, but no other emulation mode
  seems to do it right now, so it might need some changes in the more core part of
  the screen handling. This can also be post-poned.
- a new TODO popped up: the language card has to be always enabled, as the two
  ROMS just contain the functionality of the language card and it can not be turned
  on and off (it's always on).

There is no particular hurry to merge it now, maybe let it "ripe" for another one or
two weeks. :-)

I also adapted the Linux port (audetto), Github makes it harder than necessary to
do patches in the original and the forked version of a project. There I just added
some values to the imlib, qt frontends and added the resources. I'll make a pull
request there later when this pull request here is done - also not in a big hurry here.

There is more stuff for future pull requests:
- the motherboard has a 6502 and a Z80 on it, sharing the same RAM. Thus it
  supports CP/M on board.
- It has a Vidtex Videoterm card for 80-column mode (with a hardware switch
  diretly in the video output). It works totally different than the 80-column mode on
  the later Apple models: the video ram is bank-switched in the $C000 area and
  manuals I find are not completely describing what the hardware is doing, there are
  differences I have to identify first.

tomcw:
>There is no particular hurry to merge it now, maybe let it "ripe" for another one or two weeks. :-)

Yes, this is completely fine - take your time (no rush or pressure! :-)

Let's use this PR issue to discuss how to achieve certain things when they don't quite fit into AppleWin's current structure (eg. extra keys, Videx-style 80 col video, etc).

andreasbaumann:
On the light-colored arrow keys: I just saw on
https://www.andreasbaumann.cc/hardware/imc-2001/manual_imc_2001_2.01.pdf, page 56 that

- `ESC @` is mapped to `HOME`
- `ESC A-D` to those light arrow keys.

And the escape sequences already work in the emulator and are the standard escape sequences
on an Apple ][.

So all the IMC-2001 keyboard controller does, is, it sends keyboard sequences when you press
the arrow keys or the HOME key.

tomcw:
>It has a Vidtex Videoterm card for 80-column mode (with a hardware switch
diretly in the video output). It works totally different than the 80-column mode on
the later Apple models: the video ram is bank-switched in the $C000 area and
manuals I find are not completely describing what the hardware is doing, there are
differences I have to identify first.

Sounds like it'll need AppleWin to support a Videx Videoterm (80 col) card.
There's already an issue requesting this: #105

PR #1346, opened 2024-12-02
check _WIN32 instead of _MSC_VER

Jamiras:
I'm trying to get the [libretro core](https://github.com/audetto/AppleWin/issues/44) compiling for Windows using msys2 (see https://github.com/audetto/AppleWin/issues/49). I've done this successfully via https://github.com/audetto/AppleWin/pull/198, but have been requested to move all changes to the files in the base repository into a PR in the base repository. This is the [first of two](https://github.com/audetto/AppleWin/pull/198#issuecomment-2509021847) PRs that will be made against the base repository.

The bulk of this PR is to change `#ifdef _MSC_VER` to `#ifdef _WIN32` as the mingw compiler does not define `_MSC_VER`. Additional changes will be noted in the PR itself.


source/Debugger/Debug.cpp
```diff
@@ -6538,7 +6538,9 @@ bool ParseAssemblyListing ( bool bBytesToMemory, bool bAddSymbols )
 		{
 			*p = 0;
 			//	sscanf( sLine, "%s %s %s %s %s %s %s %s", sAddr1, sByte1, sByte2, sByte3, sLineN, sLabel, sAsm, sParam );
-			sscanf( sLine, "%X", &nAddress );
+			unsigned nScanAddress; // helper variable in case sizeof(DWORD) != sizeof(unsigned)
+			sscanf( sLine, "%X", &nScanAddress );
+			nAddress = nScanAddress;
```
Jamiras, line 6543: Addresses compiler error:
```
 6542 | sscanf( sLine, "%X", &nAddress );
      |                 ~^   ~~~~~~~~~
      |                  |   |
      |                  |   DWORD* {aka long unsigned int*}
      |                  unsigned int*
      |                  %lX
```

source/StdAfx.h
```diff
@@ -1,5 +1,8 @@
-#ifdef _MSC_VER
+#ifdef _WIN32
 
+#ifdef __MINGW32__
+#define STRSAFE_NO_DEPRECATE
+#endif
```
Jamiras, line 5: addresses compiler error (many similar):
```
 error: 'strcpy_instead_use_StringCbCopyA_or_StringCchCopyA' was not declared in this scope
 7910 |         strcpy( pArg->sArg, g_aCommands[ g_iCommand ].m_sName );
      |         ^~~~~~
```

source/StdAfx.h
```diff
@@ -33,6 +36,7 @@ typedef UINT64 uint64_t;
 #include <ddraw.h>
 #include <htmlhelp.h>
 #include <assert.h>
+#include <winsock.h>
```
Jamiras, line 39: addresses compiler error (many similar):
```
 error: 'SOCKET' does not name a type
  112 |         SOCKET m_hCommListenSocket;
      |         ^~~~~~
```

source/StdAfx.h
```diff
@@ -77,11 +81,11 @@ typedef UINT64 uint64_t;
 #include <string>
 #include <vector>
 
-#include "windows.h"
+#include "linux/libwindows/windows.h"
```
Jamiras, line 84: makes it more explicit that a substitute windows header is being included in the non-windows build

source/Uthernet2.cpp
```diff
@@ -43,9 +43,11 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 // in the checks below we allow both in all cases
 // (errno == SOCK_EINPROGRESS || errno == SOCK_EWOULDBLOCK)
 // this works, bu we could instead define 2 functions and check only the correct one
-#ifdef _MSC_VER
+#ifdef _WIN32
 
+#ifndef _SSIZE_T_DEFINED
 typedef int ssize_t;
+#endif
```
Jamiras, line 50: addresses compiler error:
```
 error: conflicting declaration 'typedef int ssize_t'
   49 | typedef int ssize_t;
      |             ^~~~~~~
...
C:/msys64/mingw64/include/corecrt.h:45:35: note: previous declaration as 'typedef long long int ssize_t'
   45 | __MINGW_EXTENSION typedef __int64 ssize_t;
```

source/StdAfx.h
audetto: Don't do this. `windows.h` and "my" `windows.h` are the same since only one can be included at a time. It is managed by the makefile.

source/Debugger/Debug.cpp
audetto: What is the size on your compiler of `long unsigned int` ? 4 or 8?

source/StdAfx.h
Jamiras: But if the include paths aren't set up in the correct order, it could grab the wrong one. Now that it's looking for `_WIN32`, that _shouldn't_ happen, but this makes it clear that it's not the header provided by Microsoft.

source/Debugger/Debug.cpp
Jamiras: It appears to be 4. This doesn't fail:
```
			static_assert(sizeof(DWORD) == sizeof(unsigned int));
```
It's just complaining that `%X` isn't the right format for `long unsigned int` (which is what `DWORD` is typedef'd as).

I could silence the warning by doing:
```
			sscanf( sLine, "%X", (unsigned*)&nAddress );
``` 
But then if `nAddress` is ever more than 4 bytes, it would only fill the first four and not generate a compile error.

sicklittlemonkey:
This is really great work both you guys are doing!

source/StdAfx.h
Jamiras: I've changed this back, but left a comment making it clear that this isn't the Microsoft header.

source/Debugger/Debug.cpp
audetto: no longer needed. it is a `uint32_t` now.

audetto:
other than `nScanAddress` which is now obsolete, I have no objections to the rest.

source/Debugger/Debug.cpp
Jamiras: reverted

tomcw:
Thanks both - I've now commited this to the main line.

PR #1347, opened 2024-12-07
DWORD: introduce a macro for its correct printf format.

audetto:
DWORD is always 32 bits, but in Windows it is a "unsigned long", while in linux it must be a "unsigned int".

Same as we have done for `size_t` and Co.

### DIFF
--- a/source/Core.cpp
+++ b/source/Core.cpp
@@ -162,7 +162,7 @@ void LogFileTimeUntilFirstKeyRead(void)
 
 	DWORD dwTime = GetTickCount() - dwLogKeyReadTickStart;
 
-	LogFileOutput("Time from emulation reboot until first $C000 access: %d msec\n", dwTime);
+	LogFileOutput("Time from emulation reboot until first $C000 access: %" DWORD_T_FMT " msec\n", dwTime);
 
 	bLogKeyReadDone = true;
 }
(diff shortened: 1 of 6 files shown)


audetto:
To be crystal clear, in gcc 64 bit a `long` is 64 bit, so unfortunately there is no common `typedef` for `DWORD` (ok, unless Windows changed it to `int`...).

One could also drop all these types and move to new well defined `uint32_t` and similar...

tomcw:
>One could also drop all these types and move to new well defined `uint32_t` and similar...

Certainly for this PR, you could change the type from `DWORD` to `uint32_t`.

Are you also suggesting making this change throughout the entire AppleWin codebase?
(We could do this. It's just at the Win32 API that we'd need to cast to/from `DWORD`)

audetto:
Yes, I was suggesting to do it everywhere, unless it is directly related to a Win32 API. These cases, I would analyse separately.

audetto:
For these lonely xmas days... definitely not touching 3rd party code.

audetto:
Here it is:

https://github.com/AppleWin/AppleWin/pull/1350

I have replaced all DWORDs except the ones needed for Win32 API or which were anyway tightly coupled with it.

tomcw:
Now (after PR #1350) searching for "DWORD":
- Matching lines: 143 (was 454)
- Matching files: 30 (was 86)

audetto:
Thank you. The remaning ones are either necessary or very tightly coupled win WIN32 API.

PR #1350, opened 2024-12-19
Replace DWORD -> uint32_t.

audetto:
Some have been left where tightly coupled with the Win32 API.


audetto:
Yes, because there is still 1 place where we print a DWORD. But after Xmas I can get rid of it too.

tomcw:
Thanks - merged it.

PR #1352, opened 2024-12-27
add SoundBuffer interface

Jamiras:
Second PR for getting a [libretro core](https://github.com/audetto/AppleWin/issues/44) to compile in the msys2 environment.

This replaces the `LPDIRECTSOUNDBUFFER` pointer with a pure virtual class and an implementation that wraps the `LPDIRECTSOUNDBUFFER`, allowing other sound implementations without having to simulate the exact alignment of a `DIRECTSOUNDBUFFER` object.

Additionally, moves the `DSInit` and `DSUninit` calls out of the SoundCore.cpp file, allowing them to be appropriately reimplemented in a non-DirectSound environment.

Further changes could be made to eliminate the DirectSound-based flag defines, and functions could be renamed to not imply a dependence on DirectSound, but I tried to limit the scope of changes.

### DIFF
--- a/source/SoundBuffer.h
+++ b/source/SoundBuffer.h
@@ -1,0 +1,25 @@
+#pragma once
+
+class SoundBuffer
+{
+public:
+	virtual ~SoundBuffer() = default;
+
+	virtual HRESULT Init(DWORD dwFlags, DWORD dwBufferSize, DWORD nSampleRate, int nChannels, LPCSTR pDevName) = 0;
+	virtual HRESULT Release() = 0;
+
+	virtual HRESULT SetCurrentPosition(DWORD dwNewPosition) = 0;
+	virtual HRESULT GetCurrentPosition(LPDWORD lpdwCurrentPlayCursor, LPDWORD lpdwCurrentWriteCursor) = 0;
+
+	virtual HRESULT Lock(DWORD dwWriteCursor, DWORD dwWriteBytes, LPVOID* lplpvAudioPtr1, DWORD* lpdwAudioBytes1, LPVOID* lplpvAudioPtr2, DWORD* lpdwAudioBytes2, DWORD dwFlags) = 0;
+	virtual HRESULT Unlock(LPVOID lpvAudioPtr1, DWORD dwAudioBytes1, LPVOID lpvAudioPtr2, DWORD dwAudioBytes2) = 0;
+
+	virtual HRESULT Stop() = 0;
+	virtual HRESULT Play(DWORD dwReserved1, DWORD dwReserved2, DWORD dwFlags) = 0;
+
+	virtual HRESULT SetVolume(LONG lVolume) = 0;
+	virtual HRESULT GetVolume(LONG* lplVolume) = 0;
+
+	virtual HRESULT GetStatus(LPDWORD lpdwStatus) = 0;
+	virtual HRESULT Restore() = 0;
+};
--- a/source/SoundCore.cpp
+++ b/source/SoundCore.cpp
@@ -202,7 +202,7 @@ bool DSVoiceStop(PVOICE Voice)
 	HRESULT hr = Voice->lpDSBvoice->Stop();
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSStop failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
+		if(g_fh) fprintf(g_fh, "%s: DSStop failed (%08X)\n", Voice->name.c_str(), (uint32_t)hr);
 		return false;
 	}
 
@@ -227,7 +227,7 @@ bool DSZeroVoiceBuffer(PVOICE Voice, uint32_t dwBufferSize)
 	HRESULT hr = DSGetLock(Voice->lpDSBvoice, 0, 0, &pDSLockedBuffer, &dwDSLockedBufferSize, NULL, 0);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
+		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), (uint32_t)hr);
 		return false;
 	}
 
@@ -237,14 +237,14 @@ bool DSZeroVoiceBuffer(PVOICE Voice, uint32_t dwBufferSize)
 	hr = Voice->lpDSBvoice->Unlock((void*)pDSLockedBuffer, dwDSLockedBufferSize, NULL, 0);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
+		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), (uint32_t)hr);
 		return false;
 	}
 
 	hr = Voice->lpDSBvoice->Play(0,0,DSBPLAY_LOOPING);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSPlay failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
+		if(g_fh) fprintf(g_fh, "%s: DSPlay failed (%08X)\n", Voice->name.c_str(), (uint32_t)hr);
 		return false;
 	}
 
@@ -267,7 +267,7 @@ bool DSZeroVoiceWritableBuffer(PVOICE Voice, uint32_t dwBufferSize)
 							&pDSLockedBuffer1, &dwDSLockedBufferSize1);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
+		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), (uint32_t)hr);
 		return false;
 	}
 
@@ -279,7 +279,7 @@ bool DSZeroVoiceWritableBuffer(PVOICE Voice, uint32_t dwBufferSize)
 									(void*)pDSLockedBuffer1, dwDSLockedBufferSize1);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
+		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), (uint32_t)hr);
 		return false;
 	}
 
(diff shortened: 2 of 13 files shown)


source/Speaker.cpp
```diff
@@ -980,22 +980,22 @@ bool Spkr_DSInit()
 		SpeakerVoice.nVolume = DSBVOLUME_MAX;
 
 	hr = SpeakerVoice.lpDSBvoice->SetVolume(SpeakerVoice.nVolume);
-	LogFileOutput("Spkr_DSInit: SetVolume(%d) res = %08X\n", SpeakerVoice.nVolume, hr);
+	LogFileOutput("Spkr_DSInit: SetVolume(%d) res = %08X\n", SpeakerVoice.nVolume, (unsigned)hr);
```
tomcw, line 983: Can you change this cast "(unsigned)" to "(uint32_t)".

source/Speaker.cpp
```diff
@@ -980,22 +980,22 @@ bool Spkr_DSInit()
...
 
 	//
 
 	DWORD dwCurrentPlayCursor, dwCurrentWriteCursor;
 	hr = SpeakerVoice.lpDSBvoice->GetCurrentPosition(&dwCurrentPlayCursor, &dwCurrentWriteCursor);
 	if (FAILED(hr))
-		LogFileOutput("Spkr_DSInit: GetCurrentPosition failed (%08X)\n", hr);
+		LogFileOutput("Spkr_DSInit: GetCurrentPosition failed (%08X)\n", (unsigned)hr);
```
tomcw, line 990: Can you change this cast "(unsigned)" to "(uint32_t)".

source/Speaker.cpp
```diff
@@ -980,22 +980,22 @@ bool Spkr_DSInit()
...
 	if (SUCCEEDED(hr) && (dwCurrentPlayCursor == dwCurrentWriteCursor))
 	{
 		// KLUDGE: For my WinXP PC with "VIA AC'97 Enhanced Audio Controller"
 		// . Not required for my Win98SE/WinXP PC with PCI "Soundblaster Live!"
 		Sleep(200);
 
 		hr = SpeakerVoice.lpDSBvoice->GetCurrentPosition(&dwCurrentPlayCursor, &dwCurrentWriteCursor);
-		LogFileOutput("Spkr_DSInit: GetCurrentPosition kludge (%08X)\n", hr);
+		LogFileOutput("Spkr_DSInit: GetCurrentPosition kludge (%08X)\n", (unsigned)hr);
```
tomcw, line 998: Can you change this cast "(unsigned)" to "(uint32_t)".

source/SoundCore.cpp
```diff
@@ -280,7 +227,7 @@ bool DSZeroVoiceBuffer(PVOICE Voice, uint32_t dwBufferSize)
 	HRESULT hr = DSGetLock(Voice->lpDSBvoice, 0, 0, &pDSLockedBuffer, &dwDSLockedBufferSize, NULL, 0);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), hr);
+		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
```
tomcw, line 230: Can you change this cast "(unsigned)" to "(uint32_t)".

source/SoundCore.cpp
```diff
@@ -290,14 +237,14 @@ bool DSZeroVoiceBuffer(PVOICE Voice, uint32_t dwBufferSize)
 	hr = Voice->lpDSBvoice->Unlock((void*)pDSLockedBuffer, dwDSLockedBufferSize, NULL, 0);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), hr);
+		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
```
tomcw, line 240: Can you change this cast "(unsigned)" to "(uint32_t)".

source/SoundCore.cpp
```diff
@@ -290,14 +237,14 @@ bool DSZeroVoiceBuffer(PVOICE Voice, uint32_t dwBufferSize)
...
 		return false;
 	}
 
 	hr = Voice->lpDSBvoice->Play(0,0,DSBPLAY_LOOPING);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSPlay failed (%08X)\n", Voice->name.c_str(), hr);
+		if(g_fh) fprintf(g_fh, "%s: DSPlay failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
```
tomcw, line 247: Can you change this cast "(unsigned)" to "(uint32_t)".

source/SoundCore.cpp
```diff
@@ -332,7 +279,7 @@ bool DSZeroVoiceWritableBuffer(PVOICE Voice, uint32_t dwBufferSize)
 									(void*)pDSLockedBuffer1, dwDSLockedBufferSize1);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), hr);
+		if(g_fh) fprintf(g_fh, "%s: DSUnlock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
```
tomcw, line 282: Can you change this cast "(unsigned)" to "(uint32_t)".

source/SoundCore.cpp
```diff
@@ -320,7 +267,7 @@ bool DSZeroVoiceWritableBuffer(PVOICE Voice, uint32_t dwBufferSize)
 							&pDSLockedBuffer1, &dwDSLockedBufferSize1);
 	if(FAILED(hr))
 	{
-		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), hr);
+		if(g_fh) fprintf(g_fh, "%s: DSGetLock failed (%08X)\n", Voice->name.c_str(), (unsigned)hr);
```
tomcw, line 270: Can you change this cast "(unsigned)" to "(uint32_t)".

tomcw:
@Jamiras - can you change this cast "(unsigned)" to "(uint32_t)". I've add comments to some, but not all of them.

tomcw:
Overall this change looks OK to me.

@audetto can you take a look and raise any suggestions here.

audetto:
> @Jamiras - can you change this cast "(unsigned)" to "(uint32_t)". I've add comments to some, but not all of them.

I think `HRESULT` will require some cleanup like `DWORD`, but for now this is the most pragmatic way to handle it.

source/SoundCore.cpp
```diff
@@ -139,7 +105,7 @@ static const char *DirectSound_ErrorText (HRESULT error)
 
 //-----------------------------------------------------------------------------
 
-HRESULT DSGetLock(LPDIRECTSOUNDBUFFER pVoice, uint32_t dwOffset, uint32_t dwBytes,
+HRESULT DSGetLock(std::shared_ptr<SoundBuffer>& pVoice, uint32_t dwOffset, uint32_t dwBytes,
```
audetto, line 108: make this a `const std::shared_ptr<SoundBuffer>& pVoice` or it seems it is assigned inside this functon.

source/SoundCore.h
```diff
@@ -4,10 +4,11 @@
 
 #define SAFE_RELEASE(p)      { if(p) { (p)->Release(); (p)=NULL; } }
```
audetto, line 5: I am not sure this should exist or used any longer. It only makes sense in the Win32 implementation. For the others I suspect it will leak.

source/SoundBuffer.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+class SoundBuffer
+{
+public:
+	virtual ~SoundBuffer() = default;
+
+	virtual HRESULT Init(DWORD dwFlags, DWORD dwBufferSize, DWORD nSampleRate, int nChannels, LPCSTR pDevName) = 0;
+	virtual HRESULT Release() = 0;
```
audetto, line 9: Does this make sense in a C++ class? The Win32 `Relase` should only be called in the Win32 destructor.

source/SoundCore.h
```diff
@@ -20,7 +21,6 @@ struct VOICE
 	VOICE(void)
 	{
 		lpDSBvoice = NULL;
```
audetto, line 23: it is already a `shared_ptr` default initialised, maybe this can be removed.

source/Windows/DXSoundBuffer.h
```diff
@@ -0,0 +1,30 @@
...
+	virtual HRESULT Play(DWORD dwReserved1, DWORD dwReserved2, DWORD dwFlags);
+
+	virtual HRESULT SetVolume(LONG lVolume);
+	virtual HRESULT GetVolume(LONG* lplVolume);
+
+	virtual HRESULT GetStatus(LPDWORD lpdwStatus);
+	virtual HRESULT Restore();
+
+private:
+	LPDIRECTSOUNDBUFFER m_pBuffer;
```
audetto, line 29: to avoid confusion, this should be initialised to null either here or in a constructor.

audetto:
> @Jamiras - can you change this cast "(unsigned)" to "(uint32_t)". I've add comments to some, but not all of them.

Actually, since `HRESULT` is signed, should this not be `int32_t`?

source/SoundCore.h
audetto: I was wrong, it is fine.

source/SoundBuffer.h
audetto: Again, I was wrong, it is fine.

tomcw:
>Actually, since `HRESULT` is signed, should this not be `int32_t`?

Ah yes, so it is!
Sorry @Jamiras - can you switch to `int32_t` instead.

Jamiras:
> Actually, since HRESULT is signed, should this not be int32_t?

Even so, it's [always reported as](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/705fb797-2175-4a90-b5a3-3918024b10b8) an unsigned hex value. The [uppermost bit](https://learn.microsoft.com/en-us/office/client-developer/outlook/mapi/hresult) reflects the severity of the error.

image

And `%08X` expects an unsigned value, so I believe casting it to an unsigned value is appropriate for use in these log statements.

source/SoundBuffer.h
Jamiras: This is just an interface definition. The implementation can provide an empty function if no additional logic is required.

source/SoundCore.cpp
Jamiras: done.

source/SoundCore.h
Jamiras: done.

source/Windows/DXSoundBuffer.h
Jamiras: done.

audetto (approved):
OK

tomcw:
Great! Thanks again both... and merged into the main line.

PR #1355, opened 2024-12-31
A few miscellaneous cleanups.

audetto:
(I could split in 5 PR if needed)

1. `SoundBuffer`: hide Win32-only functions (`DSInit` and `DSuninit`)
2. `SoundBuffer`: remove `Release` from interface (it is only meaningful in the Win32 destructor) and use a non-COM pattern to release it
3. Make resource loading and names more uniform for other platforms
4. Fix 2019 solution


audetto:
BTW what is the minimum VS version still supported?

If we only offer 2019+ solutions, we could remove anything below `_MSC_VER` 1920.
Or at least 1900 (=2015) which is often checked against.

tomcw:
>BTW what is the minimum VS version still supported?

VS2019 - so yes, there's some old "cruft" that can be removed.

audetto:
I've added another commit to merge the 2-step process of creation of a `SoundBuffer` into a single step.

I realised that the 2-step process is more cumbersome to implement and would not permit a complete lock of the objects once created.

audetto:
If you want, I can remove the outdated `_MSC_VER` too, otherwise, I am happy with this one.

tomcw:
@audetto - I've merged it to the main line.

Feel free to raise a new PR for the outdated `_MSC_VER`.

Thanks.

PR #1359, opened 2025-01-05
DebugFont: use an ID which is a valid macro.

audetto:
Simplifies linux resource embedding.


tomcw:
Merged it to master.

Michaelangel007:
@audetto Thanks for catching this.

I've since come across [Every bug/quirk of the Windows resource compiler (rc.exe), probably](https://www.ryanliptak.com/blog/every-rc-exe-bug-quirk-probably/) so hopefully I won't introduce more of these.

PR #1368, opened 2025-01-15
Double dot in disk filename

hasseily:
Small typo, you have 2 dots in the extension for the filename when creating new disks.


Michaelangel007 (commented):
Thanks for catching this.

Issue #1370, opened 2025-01-26
Increase maximum floppy tracks from 40


Michaelangel007:
On the _Nox Archaist_ discord we were discussing the C64 (yes, heresy, I know. /s)  and we started talking about the maximum number of tracks supported by the hardware since the C64 supports 42 tracks (technically 82 tracks with half tracks.)  One Apple 2 user mentioned they regularly used a 42 (!) track disk images back in the day -- they have code to find the maximum number of _usable_ tracks and they were curious if AppleWin supported > 40 tracks.  I mentioned I'd have to get back to them.

Summary of my findings. (Sorry the order is not quite logical.)

1. I've updated my ProDS File System volume management utility (say that 5 times fast!) to support creating of disk images with a variable number of tracks:

 * https://github.com/Michaelangel007/apple2_prodos_utils

2. I added a batch file to create disks of various sizes:

 * https://github.com/Michaelangel007/apple2_prodos_utils/blob/master/bin/test_disk_sizes.bat

3. I tried mounting a 41 track `.po` disk image as a Floppy Disk image in AppleWin and it complains.  A user pointed out that max disk image size is ~7 MB from `CDiskImageHelper::GetMaxImageSize()` but the code that does the _actual_ checking is `IsValidImageSize()`, specifically line 591:
https://github.com/AppleWin/AppleWin/blob/9bd5f838b3d47ae271f3a16ca889abd1d6f8318c/source/DiskImageHelper.cpp#L591

```c
bool CImageBase::IsValidImageSize(const uint32_t uImageSize)
{
    m_uNumTracksInImage = 0;

    if ((TRACKS_MAX>TRACKS_STANDARD) && (uImageSize > TRACKS_MAX*TRACK_DENIBBLIZED_SIZE))
        return false;    // >160KB
```

4. I tried mounting this as a hard drive image and AppleWin like-wise complains. :confused:  **It looks like Hard Drive Images MUST be >= 200 tracks (3 1/2") and <= 8192 tracks (32 MB hard drive).**

I mentioned to the user that an alternative would be to compile AppleWin from source since it would be pretty trivial to change the max track size in [DiskImage.h](https://github.com/AppleWin/AppleWin/blob/master/source/DiskImage.h#L42-L44)
```c
    #define    TRACKS_STANDARD 35
    #define    TRACKS_EXTRA    5        // Allow up to a 40-track .dsk image (160KB)
[...]

PR #1375, opened 2025-02-06
Document the disk image size magic numbers.

Michaelangel007:
@tomcw I tracked down almost all of the magic numbers for disk sizes in DiskImageHelper.cpp mentioned in #1370.

Do we need update the AppleWin.chm and change this line?

> "Sometimes on the Internet you will see a disk image that is 143,488 or 143,616 bytes long;"

To something more like this?

"Sometimes on the Internet you will see a disk image that is not 143,360 bytes long;"

### DIFF
--- a/bin/A2_DOS33.SYM2
+++ b/bin/A2_DOS33.SYM2
@@ -1,0 +1,93 @@
+// DOS 3.3 Symbols for AppleWin
+// Min debugger version 2.9.2.4
+
+// = Functions =
+symdos33 DOS.COS0    = 9EEB
+symdos33 DOS.COS1    = 9F12
+symdos33 DOS.SCANCMD = 9FCD
+symdos33 DOS.ORTN    = 9FB3
+symdos33 DOS.MVCSW   = A851
+symdos33 DOS.EBLD    = A35D
+symdos33 DOS.EBLD1   = A382
+symdos33 DOS.EBLD2   = A360
+symdos33 DOS.EBLD3   = A36C
+symdos33 DOS.EBSV    = A331
+symdos33 DOS.EBRUN   = A38E
+symdos33 DOS.LD3     = A471
+symdos33 DOS.GNBC    = A1A4
+
+// = Data =
+dw DOS.CFTABA        = AA4F
+db DOS.ISTATE        = AA51
+db DOS.OSTATE        = AA52
+dw DOS.SVOUTS        = AA53
+dw DOS.SVINS         = AA55
+db DOS.CNFTBS        = AA57
+db DOS.DFNFB         = AA58
+db DOS.SVSTK         = AA59
+db DOS.SVX           = AA5A
+db DOS.SVY           = AA5B
+db DOS.SVA           = AA5C
+db DOS.LBUFD         = AA5D
+db DOS.MONMOD        = AA5E
+db DOS.CMDNO         = AA5F
+dw DOS.SVBL          = AA60
+db DOS.SVCMD         = AA62
+db DOS.TEMP1A        = AA63
+db DOS.TEMP2A        = AA64
+db DOS.INOPTS        = AA65
+dw DOS.CV            = AA66
+dw DOS.CD            = AA68
+dw DOS.CS            = AA6A
+dw DOS.CL            = AA6C
+dw DOS.CR            = AA6E
+dw DOS.CB            = AA70
+dw DOS.CA            = AA72
+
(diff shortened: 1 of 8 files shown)


tomcw:
>To something more like this?

Yes, that sounds better than what we currently have.

Michaelangel007:
@tomcw This is ready to be merged.  I also updated [source/DiskImageHelper.cpp](https://github.com/AppleWin/AppleWin/pull/1375/files#diff-8188671ea902ecb9d7796cdce3a8f932fc811b6e9033dc43c6aa382e186aaa35) with a reminder to update `help/ddi-formats.html`

This is a preview of the updated section.

---

<p>Apple floppy disks (normally) contain 35 tracks with
16 sectors per track, for a total of 560 sectors. Each of these
sectors contained 256 bytes of information, for a total of
143,360 bytes per disk. Therefore, DOS order disk images are
always at least 143,360 bytes long. Sometimes on the Internet you
will see a disk image that is not 143,360 bytes long; this
is probably a DOS order image with extra header information
before or after the image. In most cases, AppleWin can
automatically detect this and handle it. </p>

<p>Some disk drives were able to handle 40+ tracks.  AppleWin accept disk images up to 40 tracks -- 163,840 bytes -- as long as the disk image is a multiple of 4,096 bytes / track.</p>

<p>Due to some games being incorrectly converted to a disk image AppleWin also accepts disk images of sizes:
<ul>
  <li>143,403 bytes (<i>Castle Wolfenstein</i>)</li>
  <li>143,488 bytes (<i>Rescue Raiders</i>)</li>
</ul>

---

tomcw:
@Michaelangel007 - this mainly looks good.

Pedantically, your commented: (in source/DiskImageHelper.cpp) 

>AppleWin.chm (Disks and Disk Images > Disk Image Formats) mentions disk images of sizes 143,488 and 143,616 bytes
>		//     "Sometimes on the Internet you will see a disk image that is 143,488 or 143,616 bytes long;"

...but you also now changed the AppleWin.chm documentation to say:

>Sometimes on the Internet you will see a disk image that is not 143,360 bytes long;

So in DiskImageHelper.cpp you could change the comment to past tense:

>AppleWin.chm (Disks and Disk Images > Disk Image Formats) used to mention...

Michaelangel007:
> Pedantically, your commented: (in source/DiskImageHelper.cpp)

Thanks for the catch!  Fixed.

Michaelangel007:
> @Michaelangel007 - re. A2_DOS33.SYM2
> 
> What is the new "SYM2" extension? (Seems to imply the debugger has to RUN it?) 

**TL:DR;** Yes, that's why I left this comment.

```c
		// if ".sym2" extension then RUN since we need support for DB, DA, etc.
```

The long answer:

The very old `.sym` symbol file comes in two flavors as mentioned in `ParseSymbolTable()`:

```
			// Support 2 types of symbols files:
			// 1) AppleWin:
			//    . 0000 SYMBOL
			//    . FFFF SYMBOL
			// 2) ACME:
			//    . SYMBOL  =$0000; Comment
			//    . SYMBOL  =$FFFF; Comment```

The problem is that it doesn't convey data bytes only functions (entry points). I didn't want to break code (ours or external) that handles these files.

Debugger script files are plain text files -- it doesn't care what the extension is when `RUN`

On startup the debugger auto-pre-loads the following symbol table:

```c
	const char*     g_sFileNameSymbols[ NUM_SYMBOL_TABLES ] = {
		 "APPLE2E.SYM"
		,"A2_BASIC.SYM"
		,"A2_ASM.SYM"
		,"A2_USER1.SYM" // "A2_USER.SYM",
		,"A2_USER2.SYM"
		,"A2_SRC1.SYM" // "A2_SRC.SYM",
		,"A2_SRC2.SYM"
		,"A2_DOS33.SYM2"
		,"A2_PRODOS.SYM"
	};
```

Since we need to mark up data bytes `DB, DA, DW, etc` I've used `.sym2` as the new file extension.

> Should this be documented, eg. in help/dbg-symbols.html ?

Good catch! I completely forgot about it being mentioned there.

Yes.  I'll create a new issue to track this as there a few features I see that I need to add.

* [ ] Help
* [ ] Update preloading of symbol tables first trying `.sym2` extension and falling back to `.sym`
* [ ] Actually preload ALL the files mentioned in `g_sFileNameSymbols.`

tomcw:
Thanks @Michaelangel007 - now merged to master.

PR #1379, opened 2025-02-10
Remove TEXT() and _T().

audetto:
A long exercise in regular expressions...


sicklittlemonkey:
"Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems."
Actually, I use them quite a bit. Google's RE2 library using DFA was a revelation years ago.

xotmatrix:
I live and die by RegEx. It's a problem.

tomcw:
Thanks - merged it to the main line.

PR #1383, opened 2025-02-17
Fix initial volume of SSI263 and after a Load State.

audetto:
It was set at the maximum level rather than the MockingBoard volume.

At the same time, I have removed some redundant checks.


tomcw:
Thanks - merged to main.

PR #1400, opened 2025-05-04
Portable detection of 32/64 builds.

audetto:


tomcw:
Thanks @audetto - that was quick, and much better to be portable.

tomcw:
Merged.

PR #1406, opened 2025-06-08
Make ProDOS utils cross platform.

audetto:
1. move utilities to a separate file `ProDOS_Utils.cpp`
2. avoid leaks by using `std::vector`
3. `TCHAR` and `TEXT` are no longer in use anywhere else.

I wonder if we should make all the functions from `ProDOS_Utils.h` return a `bool` or check internally whether the file exists or not already.


tomcw:
LGTM - @Michaelangel007 ?

tomcw:
>I wonder if we should make all the functions from `ProDOS_Utils.h` return a `bool` or check internally whether the file exists or not already.

Sounds like a good idea - so more user friendly on a failure.

PR #1410, opened 2025-06-13
Fix compilation of SSI263.cpp

audetto:
LoadStringA does not exist.


tomcw:
Odd! VS gives me a tooltip showing:
`#define LoadString LoadStringA`

But I can't find anywhere in the source where this `#define` exists... so it must be from a system header.

tomcw:
Anyway, I'll merge it, as my original use of `LoadStringA()` is clearly wrong.

Issue #1392, opened 2025-03-24
Create, test and release a 64-bit AppleWin


tomcw:
In PR #1391, I said:
> On the subject of performance, we really need a x86-64 build, which will give us a big perf boost (mainly due to all the extra registers available).

And @Michaelangel007 said:

> Yes, a native x64 build would be very nice!
> 
> I was curious what this would entail so I tried:
> 
> 1. Right-click on the AppleWin project
> 2. Properties
> 3. [Configuration Manager]
> 4. Active Solution Platform -> <New...>
> 5. Type of select the new platform -> x64
> 6. [OK]
>
> It created Debug x64 and Release x64 for all projects which is nice.
> 
> Surprisingly it worked! I guess all the clean-up work everyone has done over the years has paid off -- I was expecting the usual 32-bit -> 64-bit tranistion problems. I would feel better though if we did a "full pass" of testing things out to make sure everything is good.

Let's use this issue to track progress and record any problems found during testing.

Michaelangel007:
> So how did the 64-bit version compare?

Oh yeah, that would be helpful ;-)

CPU: Ryzen 5600X (Sorry, won't have access to my Threadrippers for a while...)

# Release x64

```
---------------------------
Benchmarks
---------------------------
Pure Video FPS:	4200 hires, 4115 text
Pure CPU MHz:	53.5 (video update)
Pure CPU MHz:	339.6 (full-speed)

EXPECTED AVERAGE VIDEO GAME
PERFORMANCE: 2338 FPS
---------------------------
OK   
---------------------------
```

# Release Win32

```
---------------------------
Benchmarks
---------------------------
Pure Video FPS:	3723 hires, 3703 text
Pure CPU MHz:	17.0 (video update)
Pure CPU MHz:	319.0 (full-speed)

EXPECTED AVERAGE VIDEO GAME
PERFORMANCE: 890 FPS
---------------------------
OK   
---------------------------
```

Michaelangel007:
Long term I guess we might want to have a command line to specify a few different benchmarks.

* Benchmark game loop of alternating drawing sprites on HGR page 1, 2 
  * such as a Lode Runner man run across the screen being chased by 5 Bunglings,
  * Ms. Pac-man chasing Pac-man intermission
  * Conan intro screen of horse riding
  * Nox Archaist intro screen? (After we get permission from Mark of course)
[...]

PR #1430, opened 2025-09-13
Add x64 build configs

tomcw:
PR for #1392


tomcw:
For reference, the Benchmark dialog box:

This PR:
image image


Main line:
image

tomcw:
@audetto - can I check with you that this branch builds under Linux?
Specifically the __cpuid() and __cpuidex() intrinsic functions in `class InstructionSet` in Win32Frame.cpp.

Or perhaps for your Linux build, it's not important, since it is in this Win32Frame.cpp file.

audetto:
Correct, win32frame is non included in the linux build.

tomcw:
Final version:
Debug vs Release (both 32-bit):
image image

PR #1432, opened 2025-09-14
Synch linux branch

audetto:
Correct a few compilation errors from the linux branch.

### DIFF
--- a/source/Debugger/Debug.cpp
+++ b/source/Debugger/Debug.cpp
@@ -373,7 +373,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 	static	Update_t ExecuteCommand ( int nArgs );
 
 // Breakpoints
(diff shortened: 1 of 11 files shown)


source/Debugger/Debug.cpp
```diff
@@ -2112,13 +2112,13 @@ static std::string GetFullPrefixAddrForBreakpoint(const AddressPrefix_t& addrPre
 	{
 		if (addrPrefix.nBank < 0x100)
 		{
-			sprintf_s(sBank, "%02X", addrPrefix.nBank);
+			sprintf(sBank, "%02X", addrPrefix.nBank);
```
tomcw, line 2115: In CodingConventions.txt it says:
> Do not use sprintf(), StringCbPrintf(), wsprintf(), etc. - instead use StrFormat().

Can you switch to using `StrFormat()` here and below?

NB. I shouldn't have used `sprintf_s()` in the first place! I'll make this explicit in CodingConventions.txt.

tomcw:
Thanks & merged to main line.

PR #1434, opened 2025-09-21
Fix all warnings in the AppleWin proj

tomcw:
NB. 10 warnings still in libyaml proj (but there's no upstream update)

### DIFF
--- a/source/Debugger/Debugger_Display.cpp
+++ b/source/Debugger/Debugger_Display.cpp
@@ -831,11 +831,14 @@ bool CanDrawDebugger()
 int PrintText ( const char * pText, RECT & rRect )
 {
 #if _DEBUG
-	if (! pText)
-		GetFrame().FrameMessageBox("pText = NULL!", "DrawText()", MB_OK );
+	if (!pText)
+	{
+		GetFrame().FrameMessageBox("pText = NULL!", "DrawText()", MB_OK);
+		return 0;
+	}
 #endif
 
-	int nLen = strlen( pText );
+	int nLen = (int)strlen( pText );
 
 #if !DEBUG_FONT_NO_BACKGROUND_TEXT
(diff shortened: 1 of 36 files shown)


tomcw:
@Michaelangel007 - I've gone through and fixed all the x64 warnings.

Most are:
- `size_t` is now 64-bit: so eg. `size_t sz; int len = sz` causes a warning.
  - Sometimes, I can't just change the `int` to `size_t`, as this can cascade down to other `int` vars that depend on it. So in these cases I'll cast to `int`.
- pointer difference assigned/added to an `int`, eg. `char *p,*q; int len = q-p;` causes a warning.
- `WPARAM` and `LRESULT` are also now 64-bit.

Can you cast your eye over the debugger changes?

source/Debugger/Debugger_Display.cpp
```diff
@@ -1716,15 +1719,15 @@ WORD DrawDisassemblyLine ( int iLine, const WORD nBaseAddress )
 
 	if (bDisasmFormatFlags & DISASM_FORMAT_TARGET_POINTER)
 	{
-		nOverflow += strlen( line.sTargetPointer ); // '####'
+		nOverflow += (int)strlen( line.sTargetPointer ); // '####'
```
Michaelangel007, line 1722: The comments on the following lines also need to be indented 5 spaces:
```c
		nOverflow += (int)strlen( line.sTargetPointer ); // '####'
		nOverflow ++  ;                                  //     ':'
		nOverflow += 2;                                  //      '##'
		nOverflow ++  ;                                  //         ' '
```

Michaelangel007 (changes requested):
Looks good!

There are a couple minor nits with multi-column alignment in the ProDOS code but I'm not going to worry about that.

PR #1437, opened 2025-09-28
A coupl eo flinux from the linux branch.

audetto:
1. `size_t` and `printf`
2. skip `copy ... prn`


tomcw:
Thanks - merged it in.

Issue #1444, opened 2025-10-21
Write disk image to disk periodically


greg-kennedy:
Currently it appears that AppleWin writes a changed disk image back to a .dsk file only when
* the disk picker is opened (e.g. to swap what's in a virtual drive or eject it), or
* the program gracefully exits.

Two days into an Ultima I playthrough, my computer crashed unexpectedly.  Imagine my surprise when loading my (frequently "saved") character disk reverted my progress to basically nothing - because AppleWin never actually wrote any changes back during the entire time I had the game open.

It would be nice if AppleWin wrote changes to disk periodically - something like "5 seconds without an emulated write triggers a background write task".

tomcw:
Interesting...

In fact, AppleWin _also_ writes the floppy image when the track changes - which is the main way that the image gets written to.

It sounds like Ultima I is only writing to a single track (to save the character state), and is never switching track.
So we probably need to consider how to periodically force a save, for this edge case.

As a workaround, you could (manually) save the state (F11). This state file contains the current track's buffer, so can be used as a starting point in the event of a computer crash.

greg-kennedy:
I did a little more testing and I think the thing that surprised me most was this:
* Start playing and save the game
* Reset the game with the "running Apple" button top-right
* Load the game - looks like the save worked!
* Crash computer
* Load the game - save lost

Short of adding a periodic flush, it'd probably be good to at least do a writeback whenever a virtual reboot happens (clicking the boot button, settings change, etc)

tomcw:
"running Apple" button (F2) = power-cycle the Apple II machine.

Note that neither reset nor power-cycle affect the drive head (track position):
https://github.com/AppleWin/AppleWin/blob/66f172e77d715b6e0deeaa2d940aea845f3a2270/source/Disk.cpp#L1794-L1796

Since the track doesn't change (on reset / power-cycle) then nothing is written to the floppy image.

tomcw:
Developer notes:

`Disk2InterfaceCard::WriteTrack(int drive)` is only called by `FlushCurrentTrack(int drive)`.

`Disk2InterfaceCard::FlushCurrentTrack(int drive)` is called by:
- EjectDiskInternal()
  - EjectDisk()
    - Destroy() - ie. WM_DESTROY (when AppleWin is closed or there's h/w change)
    - UnplugDrive()
    - InsertDisk()
    - LoadSnapshot()
    - LoadLastDiskImage(), when there's an error with InsertDisk()
[...]

PR #1447, opened 2025-11-09
DiskII: motor off & stopped spinning: write back any dirty track

tomcw:
PR for #1444

### DIFF
--- a/source/Disk.cpp
+++ b/source/Disk.cpp
@@ -1956,9 +1956,6 @@ void Disk2InterfaceCard::Update(const ULONG cycles)
 		{
 			if (!(pDrive->m_spinning -= MIN(pDrive->m_spinning, cycles)))
 			{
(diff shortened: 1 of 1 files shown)


tomcw:
I gave it a quick test using a DOS 3.3 disk, and changed the `Update()` code to only call `FlushCurrentTrack()` when `m_trackimagedirty` = true. IE, so I could trap the case when it wrote the track:

```
	if (!m_floppyMotorOn && !pDrive->m_spinning)
	{
		FloppyDisk* pFloppy = &pDrive->m_disk;
		if (pFloppy->m_trackimage && pFloppy->m_trackimagedirty)
		{
			// Motor off and not spinning: so write back any dirty track (GH#1444)
			// . this also supports the power-cycle case (where m_floppyMotorOn & m_spinning are instantaneously 0)
			FlushCurrentTrack(loop);
		}
	}
```

Then I SAVE'd some empty BASIC files to the disk, and I could see this code path being taken.

PR #1454, opened 2025-12-29
Debugger: Fix right arrow follow branch address

Michaelangel007:
Simple fix for Fix right arrow follow branch address when target branch ad…dress was negative ($81 .. $FF)


tomcw:
Thanks! ...and merged.

PR #1463, opened 2026-01-11
Spkr_SubmitWaveBuffer_FullSpeed: fix size bug.

audetto:
The 2nd part of the buffer was using dwBufferSize0 rather than dwBufferSize1.

Reduced duplication with 2 new functions.
Moreover, frame padding with 1 channel was repeating the same DCFilter() value rather than recomputing it.

### DIFF
--- a/source/Speaker.cpp
+++ b/source/Speaker.cpp
@@ -182,6 +182,34 @@ inline short DCFilter(short sample_in)
 
 //=============================================================================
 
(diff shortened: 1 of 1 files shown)


tomcw:
Thanks - this looks good. I will do some testing first before merging.

tomcw:
Merged to main line.

PR #1467, opened 2026-02-27
Update to use zlib-1.3.2

kiyolee:
All the files are copied from official release tarball zlib-1.3.2.tar.xz unchanged except that they are converted to DOS format to avoid format change warnings from git.


tomcw:
Thanks @kiyolee - merged to main line.

Issue #1469, opened 2026-03-01
32-MB HD creation bug


blindphoneman:
When you use the option to create a new 32 MB HD image, CiderPress opens them read-only.  Looking at the properties for the image within CiderPress, for the default disk (with PRODOS and the other files), you get this warning:

- WARNING: ProDOS filesystem blockcount (0) too small, setting to 16.
- 9 blocks are marked used but not part of any file.
- WARNING: 57 blocks are used by files but not marked used.
- WARNING: Blocks past the end of the disk are marked 'in use' in the volume bitmap.
- WARNING: One or more files look suspicious.

A totally blank HD gives this message:

- WARNING: ProDOS filesystem blockcount (0) too small, setting to 16.
- 9 blocks are marked used but not part of any file.
- WARNING: Blocks past the end of the disk are marked 'in use' in the volume bitmap.

This only happens with the 32-MB images.  Others seem to be ok.

tomcw:
For me:

## "New ProDOS hard (32MB) disk image...":
- CP1:

<img width="752" height="250" alt="Image" src="https://github.com/user-attachments/assets/2042cc8f-7cbd-4f40-bcd5-0033e5588327" />

File -> Archive Info
```
- WARNING: ProDOS filesystem blockcount (0) too small, setting to 16.
- 9 blocks are marked used but not part of any file.
- WARNING: 57 blocks are used by files but not marked used.
- WARNING: Blocks past the end of the disk are marked 'in use' in the volume bitmap.
- WARNING: One or more files look suspicious.
```

- CP2: (1.0.7 & 1.1.1)

<img width="355" height="152" alt="Image" src="https://github.com/user-attachments/assets/f3e98394-50c4-4a29-b39d-5a6d7bd6130e" />

## "New blank hard (32MB) disk image...":
- CP1:

<img width="412" height="159" alt="Image" src="https://github.com/user-attachments/assets/a3fadd1f-5d3c-4ea9-8c4c-c6bdf165c67d" />

- CP2:  (1.0.7)

Opens image, but since it's complete empty, there's nothing shown in the archive contents and no directories.

tomcw:
Due to the CP1 warnings, I'm marking this as a bug.
I've also raised a CP2 issue, as I'd expect it to be able to open the image & offer similar warnings to CP1.

blindphoneman:
Hi:

               

Also note that CP1 doesn’t want to modify the image, but, seems to be able to read it.

 

[...]

PR #1472, opened 2026-03-09
Fix edge case when formatting a 32 MB ProDOS volume. Fixes: GH #1469 32-MB HD creation.

Michaelangel007:
Sorry for typo in commit message. Fixed and forced pushed (this branch.)

Details in #1469

### DIFF
--- a/source/ProDOS_FileSystem.h
+++ b/source/ProDOS_FileSystem.h
@@ -27,12 +27,13 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
 // --- ProDOS Consts ---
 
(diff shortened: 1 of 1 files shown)


Michaelangel007:
@tomcw Did we want to move the ProDOS magic number 0xFFFF to be a `const PRODOS_MAX_VOLUME_BLOCKS = 0xFFFF` ?

tomcw:
>Did we want to move the ProDOS magic number 0xFFFF to be a `const PRODOS_MAX_VOLUME_BLOCKS = 0xFFFF` ?

Yes, I think that's best.
Otherwise LGTM, so (after adding that `const`) go ahead and merge.

PR #1483, opened 2026-03-29
Fix linux build.

audetto:


tomcw:
Thanks @audetto - merged to main

PR #1484, opened 2026-03-29
CardManager: make GetCardChoicesFor... more x-platform.

audetto:
Move the WIndows UI string to a common separate function.

### DIFF
--- a/source/CardManager.cpp
+++ b/source/CardManager.cpp
@@ -374,7 +374,7 @@ bool CardManager::IsSingleInstanceCard(SS_CARDTYPE card)
 	return false;
 }
 
-void CardManager::GetCardChoicesForSlot(const UINT slot, const SS_CARDTYPE currConfig[NUM_SLOTS], std::string& choices, std::vector<SS_CARDTYPE>& choicesList)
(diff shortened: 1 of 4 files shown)


tomcw:
@audetto...

This code is (more or less) the same in 3 places:
```
		GetCardMgr().GetCardChoicesForSlot(SLOT0, currConfig, m_choicesList[SLOT0]);
		const std::string choices = GetCardNameChoices(m_choicesList[SLOT0]);
```

ie. `GetCardChoicesForSlot()` immediately followed by `GetCardNameChoices()`.

Can you not combine these? (either a single function, or the 1st function finishing with a call to the 2nd function)

audetto:
In the current code the pattern `+= '\0'` was repeated 6 times, now only 2.

Moreover, the string concatenation with `'\0'` only applies to the `Win32Frame` case, so it does not really belong to `CardManager`. Other UIs have completely different ways of creating the list of strings for the drop down.

I can create 2 new functions in `CPageSlots` with the same signature as the old ones, but I would like to keep in `CardManager` the UI-agnostic version.

tomcw:
>Moreover, the string concatenation with `'\0'` only applies to the `Win32Frame` case, so it does not really belong to `CardManager`.

Yes, good point.

tomcw:
Thanks... merged to main.

Issue #1491, opened 2026-04-11
Odd behaviour of inserting disks in empty slots


audetto:
`-s6 empty -d1 "demo.dsk"`

This generates a `bad_cast` in `DoDiskInsert` (373) (but could probably happen at line 421 too). nothing checks the card is inserted.

`-s7 empty -h1 "Total Replay v5.0.hdv"`

This one inserts the card, inserts the disk (I can see in the VS debugger), but does not boot from it, and the UI has no sign of the card, nor the disk. if the filename does not exist I get an error dialog box.

audetto:
Aside the meaning of inconsistent flags (probably I'd avoid a `bad_cast`), I think `InsertFloppyDisks` (which can be called from elsewhere), should do the same as `InsertHardDisks` and insert the card if missing.

And I am not sure the `slot` asserts are still meaningful.

tomcw:
FTR, I can repro both these behaviours using AppleWin 1.31.0.0 (so there's no regression from any later commits, eg. commits for the new Configuration GUI).

tomcw:
>I think `InsertFloppyDisks` (which can be called from elsewhere), should do the same as `InsertHardDisks` and insert the card if missing.

I agree they should do the same.

re. `-s6 empty` vs `-d1 <image>`
Maybe whatever was specified last on the command line should take precedence... which is consistent with if you try to insert different cards into the same slot eg. `-s3 saturn -s3 vidhd`.

tomcw:
Since `InsertFloppyDisks()` will now insert the Disk II card if the slot is empty, then the cmd line `-s6 empty -d1 <image>` results in: slot 6 = Disk II, and s6,d1 = \<image\>.

IE. this matches the behaviour of `InsertHardDisks()`.

audetto:
Looks good now.

tomcw:
Thanks, and closing.

PR #1496, opened 2026-05-02
Fixes for cmd-line support

tomcw:
PR for #1491


tomcw:
Fix bad_cast for inconsistent cmd-line options. (#1491, PR #1496)
Allow Parallel, SSC & VidHD card insertion into any slot.
Allow floppy/HDD image insertion into any slot.
. if inserting a floppy image, and the slot had no Disk II card, then insert one (consistent with HDD images).

PR #1498, opened 2026-05-17
Fix crash when a HD card is on a slot != 7

audetto:
The bug is in `source/SaveState.cpp`.

And tidy up casts and iteration bounds.


tomcw:
Thanks - C&P error (I guess). Merged to main line.

PR #1500, opened 2026-05-20
Replace some BOOL usages with proper bool

kiyolee:
1. Registry functions
  The set of functions are changed to consistently use bool arguments.
  The macros are changed to inline functions to improve type checking.
  `RegSaveValue()` are overloaded with help of templates to reduce typecasting at call sites.
  For instance, bool value would be converted consistently instead of relying on call sites to convert bool to int values.
2. Memory functions
  BOOL usages were unnecessary. Mostly internal usages which can simply be bool.
3. FrameBase
  Some member variables are changed to bool. BOOL was used because of those Registry functions.

Additional changes:
- Some expressions are forced to be bool when assigning to bool variables.

### DIFF
--- a/source/Registry.h
+++ b/source/Registry.h
@@ -1,17 +1,43 @@
 #pragma once
 #include "Card.h"
 #include "CopyProtectionDongles.h"
+#include <string>
+#include <type_traits>
 
-#define  REGLOAD(a, b)				RegLoadValue(REG_CONFIG, (a), TRUE, (b))
-#define  REGLOAD_DEFAULT(a, b, c)	RegLoadValue(REG_CONFIG, (a), TRUE, (b), (c))
-#define  REGSAVE(a, b)				RegSaveValue(REG_CONFIG, (a), TRUE, (b))
+bool RegLoadString (LPCTSTR section, LPCTSTR key, bool peruser, LPTSTR buffer, uint32_t chars);
+bool RegLoadString (LPCTSTR section, LPCTSTR key, bool peruser, LPTSTR buffer, uint32_t chars, LPCTSTR defaultValue);
+bool RegLoadValue (LPCTSTR section, LPCTSTR key, bool peruser, uint32_t* value);
+bool RegLoadValue (LPCTSTR section, LPCTSTR key, bool peruser, uint32_t* value, uint32_t defaultValue);
 
-BOOL RegLoadString (LPCTSTR section, LPCTSTR key, BOOL peruser, LPTSTR buffer, uint32_t chars);
-BOOL RegLoadString (LPCTSTR section, LPCTSTR key, BOOL peruser, LPTSTR buffer, uint32_t chars, LPCTSTR defaultValue);
-BOOL RegLoadValue (LPCTSTR section, LPCTSTR key, BOOL peruser, uint32_t* value);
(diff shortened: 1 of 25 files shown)


source/Registry.h
```diff
@@ -1,17 +1,43 @@
...
+    RegSaveString(section, key, peruser, value ? "1" : "0");
+}
+template<typename V, typename std::enable_if_t<std::is_signed_v<V> && !std::is_same_v<V, bool>, int> = 0>
+void RegSaveValue(LPCTSTR section, LPCTSTR key, bool peruser, V value) {
+    RegSaveString(section, key, peruser, std::to_string(value));
+}
+template<typename V, typename std::enable_if_t<std::is_unsigned_v<V> && !std::is_same_v<V, bool>, int> = 0>
+void RegSaveValue(LPCTSTR section, LPCTSTR key, bool peruser, V value) {
+    RegSaveValue<typename std::make_signed_t<V>>(section, key, peruser, value);
+}
```
kiyolee, line 31: Note: Following original code, unsigned values are saved as signed values. `"%d"` were used to format unsigned values.

source/Registry.h
```diff
@@ -1,17 +1,43 @@
...
+template<typename V, typename std::enable_if_t<std::is_enum_v<V>, int> = 0>
+void RegSaveValue(LPCTSTR section, LPCTSTR key, bool peruser, V value) {
+    RegSaveValue<typename std::underlying_type_t<V>>(section, key, peruser, value);
+}
```
kiyolee, line 35: Note: Play safe here. Instead of simply relying on `std::to_string(some_enum_value)` to format the enum value, here makes sure the enum value are converted to proper numeric values for formatting.

tomcw:
Thanks @kiyolee - this looks useful.
I'll need to have a deeper look at the `template` stuff in `Registry.h` to give me better confidence / understanding.

Note to self: maybe change `ResetPaging(BOOL initialize)` and `UpdatePaging(BOOL initialize)` to instead take an `enum` to express "initialize" (or not) - ie. to make the code more readable (and we won't need the comment then).

source/Registry.h
audetto: and is this a good or a bad thing?

kiyolee: As far as I can see, the code only loads/stores 32-bit values. So practically, it does not matter for now.
It becomes bad if one day the value stored need to be loaded into 64-bit variables.
To be accurate, I would prefer to use `"%d"` and `"%u"` instead for signeded and unsigned variables correspondingly. As least that is safer when the size of the underlying variables change. That can't deal with signed-ness change though but supposingly that'd be rarer and you'd need more code reviewing for that kind of change.
The code here is mainly as said to keep the origin behaviour unchanged, i.e. all values are stored as signed.
Btw, I haven't changed `RegLoadValue` to deduce what to do depending on variable types (that deserves a separate PR). When that's done, `RegSaveValue` could in theory just do `std::to_string(var)` to naturally convert the values to string as signed or unsigned; the set of templates can then be simplified. `RegLoadValue` can then do the right conversion accordingly.
Another possibility is to acctually store numeric values as REG_DWORD instead but that'd be a different kind of complication.

tomcw:
Thanks @kiyolee - I understand the template specialisations, and gave it some (quick) testing, including stepping through the new code. Now merged into the main line.

PR #1503, opened 2026-05-31
Speakr.cpp: fix a few sample / frame errors.

audetto:
`g_pSpeakerBuffer` contains interleaved data (or whatever `g_nSPKR_NumChannels` says), and `g_nBufferIdx` counts frames  not samples.


audetto:
After this pull request, which only addresses a bug, I have something to avoid duplications of code across Speaker, MB and SSI263 and repeated calculations.

And I will rename samples vs frames correctly.

tomcw:
Thanks - merged it to main

Issue #1507, opened 2026-06-09
"Unplug hard disk image" does not work if you click ok on the main configuration window before acknowlinging the unplug warning dialog


txgx42:
Start with a hard disk controller in slot 7 with a disk image mapped to HDD 1

Open the hard disk controller configuration and choose unplug for HDD1

Ignore the eject/unplug warning dialog and click OK on the controller configuration (which then closes)

Click OK on the main configuration (which does not close), then click Yes on the unplug warning dialog. The configuration window then closes, but the hard disk image remains plugged in.

Applewin 1.32.0.0. Hope this is clear enough, it was a bit hard to describe 🙂 Discovered this by mistake when I was doing things too quickly.

tomcw:
Thanks for the report, Tom - I'll take a look.

tomcw:
Thanks Tom, I can reproduce this.

It's also the same of the Disk II interface card.

PR #1508, opened 2026-06-21
Support MessageBox() for parent HWND

tomcw:
PR for #1507

### DIFF
--- a/source/Configuration/PageSlots.cpp
+++ b/source/Configuration/PageSlots.cpp
@@ -616,7 +616,7 @@ void CPageSlots::HandleFloppyDriveCombo(HWND hWnd, UINT driveSelected, UINT comb
 			m_PropertySheetHelper.GetConfigNew().m_slotInfoForFDC[ms_slot].pathname[driveSelected] = "";
 
 			std::string strText = StrFormat("%s already mounted in slot %d, drive %d.", pathname.c_str(), inUseSlot, inUseDrive + 1);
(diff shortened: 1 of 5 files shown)


tomcw:
@audetto,

I've had to add a new virtual function to `FrameBase.h` so that the `MessageBox()` blocks the parent window from continuing to respond to UI activity.

EG, In this example, I can keep selecting "Unplug Hard Disk Image" and get many "Eject/Unplug Warning" message boxes...

MsgBox

The reason is that `GetFrame().FrameMessageBox()` is using `g_hFrameWindow`, which in the above case is not the parent window. So I've added an overload to allow you to pass in the parent `HWND hWnd`.

Can you check this PR, and let me know if it's OK from your point of view?

audetto:
It is ok, but the only thing I can you is to ignore `HWND` altogether.

Anyway, the option UI is not used in linux.

tomcw:
>Anyway, the option UI is not used in linux.

Ah, so perhaps better to just use Win32's `MessageBox()` directly for the configuration UI property pages.
This may be a better approach, since it doesn't add another virtual-function to the `FrameBase` interface.

tomcw:
On the other hand, if we want all the code to go via a Platform Abstraction Layer, then removing the direct `MessageBox()` calls is the better.

audetto:
Yes. PageSlots.cpp is not included in the linux build. You can call the Win32 API directly.

PR #1509, opened 2026-07-05
Replace BOOL (almost all) with bool

kiyolee:
Except a few `BOOL` variables in `CPU` and a few Windows callback functions, all `BOOL` usages have been replaced by proper `bool`.

An inline function `_b2B()`, that converts `bool` to `BOOL`, is introduced for interacting with Windows APIs.

Note: The usages of `BOOL` in `CPU` are a bit weird and tricky to untangle. 2 of the `BOOL` variables are not actually boolean in the common sense that they require true being certain specific numeric values; they are more like flags. Anyhow, that deserves a separate PR.


tomcw:
>Note: The usages of `BOOL` in `CPU` are a bit weird and tricky to untangle. 2 of the `BOOL` variables are not actually boolean in the common sense that they require true being certain specific numeric values; they are more like flags. Anyhow, that deserves a separate PR.

I see... in `AF_TO_EF` and `EF_TO_AF` macros.
I agree that needs a separate PR.

tomcw:
In general, I agree that moving from `BOOL` to `bool` is a good thing to do.

I'm not so sure about the asymmetric conversion methods:
- BOOL to bool: using `!!`
- bool to BOOL: using `_b2B()`

Is the `!!`  idiom widely understood? Does it make for readable code? (eg. could instead use `to_bool()`?)
The `_b2B()` name: would it be better to use eg. `to_BOOL()`?

Does anyone else (@Michaelangel007, @sicklittlemonkey) have an opinion?

kiyolee:
`!!` is pretty common idiom. I submitted code using that before and use it a lot in real world.
This actually means negate (`!`) `operator!`. Any type has `operator!` (including all numeric types which has default `operator!`) would work.
You don't need `to_bool` or `to_BOOL`. The alternatives are just `bool(X)` or `BOOL(X)`. The subtle differenecs are these would call constructors instead. The compiler would work out if the expression can be used to construct `bool` or `BOOL`. If the expression value type has `operator!`, it would work like `!!`.
I need to double check, when I learn this, the main problem was for pointers. I think `bool(a_pointer)` would not compiler whereas `!!a_pointer` can and that is handy. For some types, you may need `bool(!!expr)` in the end. `!!` is kind of a shortcut for that.
I don't think being symmetric is necessary. The 2 ways I use are kind of meaning different things.
`!!` is a natural way to construct a boolean value. `_b2B` is a remarks to indicate that we are interacting with Windows APIs, where `BOOL` is still needed.
The concern of using `BOOL(expr)` is that that could become overused, even for cases there are no need.
You can't simply search `BOOL` to find all Windows APIs that need such conversion. `_b2B` would make that easy.

sicklittlemonkey:
The purist in me prefers to_bool() and to_BOOL(), but these require parentheses, which is really inconvenient.

So !! seems evil, but is convenient as it's a prefix-only replacement for to_bool().

Even worse, I see that some use +x as a convenient prefix-only replacement for to_BOOL().

tomcw:
I've been using [Compiler Explorer](https://godbolt.org/) to check what x86-64 code is actually generated for the `!!` idiom.

EG, Using this line as an example:
https://github.com/AppleWin/AppleWin/blob/8193cc1fd1b8afeb78d5321b517548f7702c3e22/source/Mockingboard.cpp#L789

NB. before PR #1500, this line was just:
``` c
bool CS_SSI263_A = nAddr & 0x40;
```

## x64, msvc v19.latest

Debug, /O0
https://godbolt.org/z/rTGsTY3Ez

Release, /O2
https://godbolt.org/z/8aWfeTYx7

TL;DR:
- the `!!` idiom is just ignored
- unsigned int to bool conversation happens implicitly, resulting in a 0 or 1 bool result

## clang-cl v22.1.6 (msvc v19.50, VS 18.2)

Debug, /O0
https://godbolt.org/z/TK3YExjYr

This does honour the `!!` with superfluous code:
```
        xor     al, -1
        xor     al, -1
```

Release, /O2
https://godbolt.org/z/GPGE7bb55

---

From the above, both debug & release builds ignore `!!`, so we can just rely on the compiler implicitly converting an (unsigned) int to a bool with just values 0 or 1.

tomcw:
There is another use for the `!!` idiom in AppleWin, eg:

``` c
	BOOL flagc;
	BOOL flagn;
	...
```

https://github.com/AppleWin/AppleWin/blob/8193cc1fd1b8afeb78d5321b517548f7702c3e22/source/CPU/cpu_instructions.inl#L245

In this case, the `BOOL` (underlying type is `int`) gets converted to 0 or 1.
(So `!!` is not ignored in this case)

## x64, msvc v19.latest

Release, /O2
https://godbolt.org/z/MqdzP5fe8

tomcw:
>I need to double check, when I learn this, the main problem was for pointers.

``` c
pcap_t * m_tfePcapFP;
```
https://github.com/AppleWin/AppleWin/blob/8193cc1fd1b8afeb78d5321b517548f7702c3e22/source/Tfe/PCapBackend.cpp#L68

If I remove the `!!` then this still compiles ok (and the generated x64 code is the same with or without it).

tomcw:
For `bool` to `BOOL`, again we can just rely on the compiler to do the right thing, without needing `_b2B()`.

Debug: https://godbolt.org/z/or35fbPb5
Release: https://godbolt.org/z/a4zT4G9nx

Michaelangel007:
> Is the `!!`  idiom widely understood? Does it make for readable code? (eg. could instead use `to_bool()`?)

It mostly comes from Javascript (and maybe early C code?)

It's not great (I prefer a C cast or C++ static cast over a function call) but I do like that it is very compact.

First choice, `!!`

Second choice, `to_bool()` or the latter `bool_to_BOOL()` is the better readable name for me. Not a fan of `_b2B()` as that is TOO compact / obtuse.

source/6522.cpp
```diff
@@ -296,15 +296,15 @@ int SY6522::OnTimer1Underflow(USHORT& counter)
 			timer += (m_regs.TIMER1_LATCH.w + kExtraTimerCycles);	// GH#651: account for underflowed cycles / GH#652: account for extra 2 cycles
 	}
 	counter = (USHORT)timer;
-	return (timer < 0) ? 1 : 0;			// timer1IrqDelay
+	return timer < 0;			// timer1IrqDelay
```
Michaelangel007: I prefer `return (timer < 0);`

@tomcw Do you have preference? Is this something we need to add to the coding standard?

source/Configuration/PageConfig.cpp
```diff
@@ -164,7 +164,7 @@ INT_PTR CPageConfig::DlgProcInternal(HWND hWnd, UINT message, WPARAM wparam, LPA
 			if(HIWORD(wparam) == CBN_SELCHANGE)
 			{
 				const VideoType_e newVideoType = (VideoType_e) SendDlgItemMessage(hWnd, IDC_VIDEOTYPE, CB_GETCURSEL, 0, 0);
-				EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), (newVideoType == VT_COLOR_IDEALIZED) ? TRUE : FALSE);
+				EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), _b2B(newVideoType == VT_COLOR_IDEALIZED));
```
Michaelangel007: Since this is in the middle of a function call I would rather prefer the verbose `bool_to_BOOL()` so that it sticks out.

Ideally, it should be moved out to a separate line IMHO:

```
    BOOL bEnabled = bool_to_BOOL(newVideoType == VT_COLOR_IDEALIZED);
    EnableWindow(GetDlgItem(hWnd, IDC_CHECK_VERTICAL_BLEND), bEnabled);

```

Not a fan of the slightly more compact `bool2BOOL()`

@tomcw Another use case for the coding standard that should be clarified IMHO.

source/Configuration/PageConfig.cpp
```diff
@@ -418,13 +418,13 @@ eApple2Type CPageConfig::GetApple2Type(uint32_t NewMenuItem)
...
 
-void CPageConfig::EnableTrackbar(HWND hWnd, BOOL enable)
+void CPageConfig::EnableTrackbar(HWND hWnd, bool enable)
 {
-	EnableWindow(GetDlgItem(hWnd,IDC_SLIDER_CPU_SPEED),enable);
-	EnableWindow(GetDlgItem(hWnd,IDC_0_5_MHz),enable);
-	EnableWindow(GetDlgItem(hWnd,IDC_1_0_MHz),enable);
-	EnableWindow(GetDlgItem(hWnd,IDC_2_0_MHz),enable);
-	EnableWindow(GetDlgItem(hWnd,IDC_MAX_MHz),enable);
+	EnableWindow(GetDlgItem(hWnd, IDC_SLIDER_CPU_SPEED), _b2B(enable));
```
Michaelangel007: Thanks for adding the spaces in the function arguments!

These 4 lines should be using a `bEnabled`:

```
    BOOL bEnabled = bool_to_BOOL(enable);
    Enable(GetDlgItem(hWnd, IDC_SLIDER_SPEED, bEnabled);
```

source/Configuration/PageSlots.cpp
```diff
@@ -673,11 +674,11 @@ void CPageSlots::HandleFloppyDriveCombo(HWND hWnd, UINT driveSelected, UINT comb
...
 
-void CPageSlots::EnableFloppyDrive(HWND hWnd, BOOL enable)
+void CPageSlots::EnableFloppyDrive(HWND hWnd, bool enable)
 {
-	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_DISK1), enable);
-	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_DISK2), enable);
-	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_DISK_SWAP), enable);
+	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_DISK1), _b2B(enable));
+	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_DISK2), _b2B(enable));
+	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_DISK_SWAP), _b2B(enable));
```
Michaelangel007: The 2nd last argument should multi-column aligned.

Replace with `BOOL bEnabled = `

source/Configuration/PageSlots.cpp
```diff
@@ -930,11 +931,11 @@ void CPageSlots::HandleHDDCombo(HWND hWnd, UINT driveSelected, UINT comboSelecte
...
 
-void CPageSlots::EnableHDD(HWND hWnd, BOOL enable)
+void CPageSlots::EnableHDD(HWND hWnd, bool enable)
 {
-	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_HDD1), enable);
-	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_HDD2), enable);
-	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_HDD_SWAP), enable);
+	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_HDD1), _b2B(enable));
+	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_COMBO_HDD2), _b2B(enable));
+	EnableWindow(GetDlgItem(hWnd, IDC_SLOT_OPT_HDD_SWAP), _b2B(enable));
```
Michaelangel007: 2nd last column aligned.

`BOOL bEnabled = `

source/Configuration/PageSlots.cpp
```diff
@@ -1041,11 +1042,9 @@ INT_PTR CPageSlots::DlgProcSSCInternal(HWND hWnd, UINT message, WPARAM wparam, L
 			const UINT serialPortItem = m_PropertySheetHelper.GetConfigNew().m_serialPortItem;
 			m_PropertySheetHelper.FillComboBox(hWnd, IDC_SERIALPORT, cardForConfig.GetSerialPortChoices().c_str(), serialPortItem);
 
-			BOOL enable = TRUE;
 			CSuperSerialCard* card = GetCardMgr().GetSSC();
-			if (card && card->IsActive())
-				enable = FALSE;
-			EnableWindow(GetDlgItem(hWnd, IDC_SERIALPORT), enable);
+			const bool enable = !card || !card->IsActive();
+			EnableWindow(GetDlgItem(hWnd, IDC_SERIALPORT), _b2B(enable));
```
Michaelangel007: Add:

```C
BOOL bEnabled = bool_to_BOOL(enable);
EnableWindow(..., bEnabled);

```

source/Debugger/Debug.h
```diff
@@ -71,7 +71,7 @@
 			bool operator() ( const Command_t & rLHS, const Command_t & rRHS ) const
 			{
 				// return true if lhs<rhs
-				return (strcmp( rLHS.m_sName, rRHS.m_sName ) <= 0) ? true : false;
+				return strcmp( rLHS.m_sName, rRHS.m_sName ) <= 0;
```
Michaelangel007: Minor edit:

```
    return (...);
```

source/FourPlay.cpp
```diff
@@ -53,19 +53,25 @@
 #include "Memory.h"
 #include "YamlHelper.h"
 
+inline static int _b2bit(bool b, int shift)
```
Michaelangel007: @tomcw I would rather prefer `bool2bits()` or even `bool2bit()`

We probably should avoid `_` prefix since they are reserved for C Lib.

source/FourPlay.cpp
```diff
@@ -53,19 +53,25 @@
...
+{
+	return b ? (1 << shift) : 0;
+}
+
+BYTE FourPlayCard::MakeByte(bool up, bool down, bool left, bool right, bool trigger1, bool trigger2)
+{
+	return _b2bit(up, 0) | _b2bit(down, 1) | _b2bit(left, 2) | _b2bit(right, 3)
```
Michaelangel007: This should be split into 8 or 9 lines IMHO:

```
    return 0
         | bool2bit(up      , 0)
         | bool2bit(down    , 1)
         | bool2bit(left    , 2)
         | bool2bit(right   , 3)
         | bool2bit(false   , 4) // not used, always low
         | bool2bit(true    , 5) // not used, always high
         | bool2bit(trigger2, 6)
         | bool2bit(trigger1, 7);
```

Better would probably be:

```c
    int bits = bool2bit(up      , 0)
             | bool2bit(down    , 1)
             | bool2bit(left    , 2)
             | bool2bit(right   , 3)
             | bool2bit(false   , 4) // not used, always low
             | bool2bit(true    , 5) // not used, always high
             | bool2bit(trigger2, 6)
             | bool2bit(trigger1, 7);
    return bits;
```

source/FourPlay.cpp
```diff
@@ -53,19 +53,25 @@
...
+         | _b2bit(false, 4) | _b2bit(true, 5) | _b2bit(trigger2, 6) | _b2bit(trigger1, 7);
```
Michaelangel007: See above.

source/Keyboard.cpp
```diff
@@ -103,9 +103,9 @@ bool KeybGetShiftStatus()
 //===========================================================================
 void KeybUpdateCtrlShiftStatus()
 {
-	g_bAltKey   = (GetKeyState( VK_MENU   ) < 0) ? true : false;	//  L or R alt
-	g_bCtrlKey  = (GetKeyState( VK_CONTROL) < 0) ? true : false;	//  L or R ctrl
-	g_bShiftKey = (GetKeyState( VK_SHIFT  ) < 0) ? true : false;	//  L or R shift
+	g_bAltKey   = GetKeyState( VK_MENU   ) < 0;	//  L or R alt
```
Michaelangel007: These should keep the parentheses

```
    g_bAltKey = (... < 0);

File: source/Mockingboard.h [hunk 12]
```diff
@@ -108,7 +108,7 @@ class MockingboardCard : public Card
 			nAYCurrentRegister[0] = nAYCurrentRegister[1] = 0;	// not valid
 			state[0] = state[1] = AY_INACTIVE;
 			isAYLatchedAddressValid[0] = isAYLatchedAddressValid[1] = false;	// after AY reset
-			isChipSelected[0] = type == CT_Phasor ? false : true;	// Only Phasor is false, all other MB variants are true
+			isChipSelected[0] = type != CT_Phasor;	// Only Phasor is false, all other MB variants are true
```
Michaelangel007: Should add parentheses here:

```
    isChipSelected[0] = (type != CT_Phasor);

File: source/StdAfx.h [hunk 13]
```diff
@@ -83,3 +83,6 @@
 #define SIZE_T_FMT "zu"
 
 #endif // _WIN32
+
+// Convert bool to BOOL for Windows API.
+inline BOOL _b2B(bool b) { return b ? TRUE : FALSE; }
```
Michaelangel007: See main discussion.

source/Tfe/tfearch.cpp
```diff
@@ -409,35 +409,37 @@ void tfe_arch_receive_remove_committed_frame()
...
 #if defined(TFE_DEBUG_ARCH) || defined(TFE_DEBUG_FRAMES)
 	if(g_fh) {
 		fprintf( g_fh, "tfe_arch_recv_ctl() called with the following parameters:" );
-		fprintf( g_fh, "\tbBroadcast   = %s", bBroadcast   ? "TRUE" : "FALSE" );
-		fprintf( g_fh, "\tbIA          = %s", bIA          ? "TRUE" : "FALSE" );
-		fprintf( g_fh, "\tbMulticast   = %s", bMulticast   ? "TRUE" : "FALSE" );
-		fprintf( g_fh, "\tbCorrect     = %s", bCorrect     ? "TRUE" : "FALSE" );
-		fprintf( g_fh, "\tbPromiscuous = %s", bPromiscuous ? "TRUE" : "FALSE" );
-		fprintf( g_fh, "\tbIAHash      = %s", bIAHash      ? "TRUE" : "FALSE" );
+		fprintf( g_fh, "\tbBroadcast   = %s", _b2psz(bBroadcast) );
```
Michaelangel007: `_b2psz()` is obtuse. We don't need to copy the bad Hungarian notation here IMHO.

Should be descriptive `bool2cstring()`

Hopefully this _won't_ be confusing with C++ std::string.

Michaelangel007 (changes requested):
Whew!

Overall looks great!

Couple of minor things here and there but easily addressable.

We should NOT be using `_b2B()` and `_b2psz()`, for two reasons:

* underscore prefix is reserved for C lib IIRC,
* these are little TOO terse.

We can hash this out in the main thread:

1. I'm good with using `!!`

2. Prefer `bool2bit()` or `bool_to_bit()`

3. Prefer `bool_to_BOOL()`. This verbosity is fine since it shouldn't be needed all that much.

4. Prefer `bool2cstring()` or `bool_to_cstring()` (for consistency). This verbosity is fine since it should rarely be needed.

I guess for consistency we should probably use `bool_to_X()`.

We should add `!!` and `return (var == state);` to the Coding Standard as a new issue once we have agreed on the few things above.

Lastly, minor nit, I would have liked to see ALL the new proposed conversion  functions mentioned in the initial PR. I didn't realize we had a `_b2psz()` in there. (Thank-you FOR calling out `_b2B()`.) No worries since that's why we do these code review, just something to consider for the future. (Normally, we _don't_ need to know about helper functions but since this one was kind of tied to Windows the 3 utilities should have been called out.) Hopefully we don't have too many Windowsisms shenanigans left. :-)

Again, fantastic work. Appreciate the cleanup!

tomcw:
@Michaelangel007 - thanks for all your comments.

I don't know if you read my comments above (from https://github.com/AppleWin/AppleWin/pull/1509#issuecomment-4951573353), but using Compiler Explorer, the majority of `!!` just collapse to nothing, so they just add extra noise to the code.

The only actual benefit of `!!` is "BOOL (unsigned int) to BOOL (0 or 1)", eg. here: https://github.com/AppleWin/AppleWin/pull/1509#issuecomment-4951621648

I'm advocating for
- not using `!!` at all (except for this special BOOL to BOOL case - and we can probably improve this code too)
- not using `_b2B()` at all

(And I hadn't noticed the `_b2psz()` until you pointed it out)

kiyolee:
Let me rework my commit taking in all the suggestions. I have no strong
objections to any of them.
I will remove all those `!!` and see how things go. I may rediscover when
it was actually needed or more convenient.

Side note: I think it is double underscore prefixes which are prohibited by
standard. Single underscore prefix is fine for general uses. However, it is
understood why some devs do not like it. It is just my habit to distinguish
short function names from variables.

Michaelangel007:
> Side note: I think it is double underscore prefixes which are prohibited by standard. Single underscore prefix is fine for general uses. 

Ah, I think you are right.

Here's the [C standard](https://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf) if anyone is interested.

Michaelangel007:
> @Michaelangel007 - thanks for all your comments.
> 
> I don't know if you read my comments above (from https://github.com/AppleWin/AppleWin/pull/1509#issuecomment-4951573353), but using Compiler Explorer, the majority of `!!` just collapse to nothing, so they just add extra noise to the code.

Thanks for verifying this Tom! I had a hunch that most wouldn't be needed.

tomcw:
>Let me rework my commit taking in all the suggestions. I have no strong objections to any of them.

Thanks.

source/6522.cpp
kiyolee: I guess this is the kind of preference which would split developers into 2 similar sized groups.
I am the one who prefer without brackets along the thinking to avoid `return` being looks like a function call.
However, I am happy to follow whichever way as long as it is written down as coding standard.

source/Configuration/PageConfig.cpp
kiyolee: Fixed in a new commit.

source/Configuration/PageConfig.cpp
kiyolee: Fixed in a new commit.

source/Configuration/PageSlots.cpp
kiyolee: Fixed in a new commit.

source/Configuration/PageSlots.cpp
kiyolee: Fixed in a new commit.

source/Configuration/PageSlots.cpp
kiyolee: Fixed in a new commit.

source/FourPlay.cpp
kiyolee: Fixed in a new commit.
Renamed `_b2bit` to `bool_to_bit` for consistence.

source/FourPlay.cpp
kiyolee: Fixed in a new commit.

source/FourPlay.cpp
kiyolee: Fixed in a new commit.

source/Tfe/tfearch.cpp
kiyolee: Fixed in a new commit.

source/StdAfx.h
kiyolee: Fixed in a new commit.

source/Keyboard.cpp
kiyolee: Fixed in a new commit.

source/Mockingboard.h
kiyolee: Fixed in a new commit.

source/Mockingboard.cpp
```diff
@@ -458,7 +458,7 @@ bool MockingboardCard::Is6522IRQ()
 	// . OR-sum of all active TIMER1, TIMER2 & SPEECH sources (from all 6522s)
 	bool irq = false;
 	for (UINT i = 0; i < NUM_SUBUNITS_PER_MB; i++)
-		irq |= !!(m_MBSubUnit[i].sy6522.GetReg(SY6522::rIFR) & 0x80);
+		irq |= ((m_MBSubUnit[i].sy6522.GetReg(SY6522::rIFR) & 0x80) != 0);
```
kiyolee, line 461: Note: This was the only case that `!!` was required because of `|=`. Explicitly using `!= 0` here.

source/6522.cpp
kiyolee: Fixed in a new commit.

source/Debugger/Debug.h
kiyolee: Fixed in a new commit.

kiyolee:
@tomcw @Michaelangel007 I have done updating the PR. Please check.

Michaelangel007:
> @tomcw @Michaelangel007 I have done updating the PR. Please check.

Looking good!  Just one minor case indentation and I think we're GTG.

@tomcw I've spawn off #1511 so we can update [CodingConventions.txt](https://github.com/AppleWin/AppleWin/blob/master/docs/CodingConventions.txt) to include the things that were discussed here.

tomcw:
@kiyolee, @Michaelangel007:
See my comment above: https://github.com/AppleWin/AppleWin/pull/1509#issuecomment-4974113937 where I said:

>I'm advocating for
>- not using _b2B() at all

Actually what I meant was that `_b2B()` or `bool_to_BOOL()` (or some other function) is not necessary - the compiler just does the right thing.
See the Compiler Explorer output here: https://github.com/AppleWin/AppleWin/pull/1509#issuecomment-4951699413

kiyolee:
Let me double check. I thought those Windows APIs might trigger warnings.

kiyolee:
Yeah, `bool_to_BOOL` is not needed. Added a commit to remove that.

tomcw:
This looks good to me. If others agree, then I'll run some tests before merging to main.

tomcw:
I've been testing the loading of save-states.

You can't replace `yamlLoadHelperLoadUint()` with `LoadBool()`.
Attempting to load _any_ save-state throws this exception:
```
---------------------------
Load State
---------------------------
Memory: Missing: IO_SELECT_InternalROM
---------------------------
OK   
---------------------------
```

And similarly, you can't replace: `yamlSaveHelper.SaveUint()` with `SaveBool()` - this breaks back-compatibility with older AppleWins.

(You could extend `yamlLoadHelperLoadBool()` to support "0" and "1" too, but you'd also need to bump all save-state yaml structs that switched from Uint to Bool to avoid breaking back-compatibility for older versions of AppleWin.)

IMO, it's simpler to just revert these state save/load functions so they don't save/load a bool type:
```
bool MemLoadSnapshot(YamlLoadHelper& yamlLoadHelper, UINT unitVersion)
- for SS_YAML_KEY_IOSELECT_INT, SS_YAML_KEY_LASTRAMWRITE

void SY6522::SaveSnapshot(YamlSaveHelper& yamlSaveHelper)
void SY6522::LoadSnapshot(YamlLoadHelper& yamlLoadHelper, UINT version)
- for SS_YAML_KEY_SY6522_TIMER1_IRQ_DELAY, SS_YAML_KEY_SY6522_TIMER2_IRQ_DELAY

void Disk2InterfaceCard::SaveSnapshotFloppy(YamlSaveHelper& yamlSaveHelper, UINT unit)
bool Disk2InterfaceCard::LoadSnapshotFloppy(YamlLoadHelper& yamlLoadHelper, UINT unit, UINT version, std::vector<BYTE>& track)
- for SS_YAML_KEY_TRACK_IMAGE_DATA, SS_YAML_KEY_TRACK_IMAGE_DIRTY

void Disk2InterfaceCard::SaveSnapshot(YamlSaveHelper& yamlSaveHelper)
bool Disk2InterfaceCard::LoadSnapshot(YamlLoadHelper& yamlLoadHelper, UINT version)
- for SS_YAML_KEY_FLOPPY_MOTOR_ON
```

kiyolee:
@tomcw Added a commit to revert those settings you identified except `SS_YAML_KEY_FLOPPY_MOTOR_ON`.
`SS_YAML_KEY_FLOPPY_MOTOR_ON` is already saved/loaded as Bool before this PR.
Can you confirm this was wrong?

tomcw:
>SS_YAML_KEY_FLOPPY_MOTOR_ON is already saved/loaded as Bool before this PR.
>Can you confirm this was wrong?

Yes - my mistake - this was wrong.

tomcw:
Thanks @kiyolee - I've now merged this to main.

tomcw:
btw, I ran the regression tests in the AppleWin-Test repo, and didn't find any issues.

PR #1519, opened 2026-09-14
Add an MCP server for driving the emulator from an AI client

### LINKED ISSUE #1518: Add an MCP server
Add an MCP server so agents can use AppleWin.

Pull request incoming. This is just a tracking issue.

XekriRedmane:
Closes #1518.

## Summary

This PR adds a [Model Context Protocol](https://modelcontextprotocol.io) server to AppleWin. When AppleWin starts with `-mcp` (or `-mcp=<port>`), it listens on `http://127.0.0.1:6502/mcp`. An MCP client such as Claude Code can then drive the emulator: read the text screen or take a screenshot, type on the keyboard, read and write memory, insert and eject disks, save and load machine state, and run debugger commands. The feature is off unless the switch is given, so there is no change in behaviour for existing users.

The server is a self-contained module in `source/MCP`. It has no dependency outside AppleWin. It carries its own small HTTP listener on winsock, a JSON reader and writer, and PNG and base64 encoders that use the zlib AppleWin already links.

## Changes to existing files

The only edits to existing source files are 8 added lines and 2 `#include`s:

- `CmdLine.cpp` calls `MCP_ParseCmdLineArg` in the `else if` chain, just before the "unsupported argument" branch.
- `AppleWin.cpp` calls `MCP_Initialize` after `RepeatInitialization` in `WinMain`, and `MCP_Destroy` at the top of `Shutdown`.

The `.vcxproj` and `.vcxproj.filters` changes add the new files under a `Source Files\MCP` filter.

## Design

**Threads.** The HTTP server runs on its own threads, but the emulator code is only safe to call from the main thread. `MCP_Initialize` creates a hidden message-only window on the main thread. The existing message loop calls `PeekMessage` with a null window handle, so it dispatches that window's messages between calls to `ContinueExecution`. A tool that must touch the machine posts a lambda to that window and waits on a condition variable, and the window procedure runs the lambda on the main thread while the CPU is between instructions. If the main thread does not run the lambda before a timeout, the caller marks the record finished under a shared mutex, so a lambda never runs after its caller's stack frame is gone. Tools that wait, such as `wait_for_text`, poll from the socket thread with short lambdas and let the machine keep running.

**Keyboard pacing.** The Apple II keyboard has one latch, so a key sent before the program clears the strobe is lost. `type_text` and `press_keys` check the strobe with `KeybReadData` before each key and wait for it to clear. If the program does not read the keyboard within `key_wait_ms`, the tool sends the key anyway and reports the count of such keys.

**Transport.** The server speaks the MCP streamable HTTP transport. It accepts `POST` with a JSON-RPC message and replies with `application/json`, and answers a notification with `202 Accepted`. It refuses `GET`, since it never sends a message on its own initiative. It binds to `127.0.0.1` only and refuses a request whose `Origin` header names a different host.

## Files

| File | Content |
| --- | --- |
| `MCP.h` | The three functions AppleWin calls |
| `MCPServer.cpp`, `MCPServer.h` | Server lifetime, the bridge to the main thread, JSON-RPC dispatch |
| `MCPTools.cpp`, `MCPTools.h` | The tools and their JSON schemas |
| `MCPHttp.cpp`, `MCPHttp.h` | The HTTP server, on winsock |
| `MCPJson.cpp`, `MCPJson.h` | A JSON parser and writer |
| `MCPEncode.cpp`, `MCPEncode.h` | PNG and base64 encoders, on zlib |
| `MCPHelpers.cpp`, `MCPHelpers.h` | String helpers shared by the above |
| `README.md` | Setup and the tool list |

## Testing

Build `AppleWin-VS2022.sln` (x64), then run `AppleWin-x64.exe -mcp`. Register the server with Claude Code:

```
claude mcp add --transport http --scope local applewin http://127.0.0.1:6502/mcp
```

Then ask the client to insert a disk, boot it, read the screen, and type into a running program. `source/MCP/README.md` lists every tool. Running without `-mcp` should behave exactly as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


XekriRedmane:
Although this was coded with the help of Claude, I have gone through the files myself (I am a C++ dev AND an Apple //e user) and made sure that it isn't "weird". I've definitely modified some things that Claude thought were ok. I'm open to changes!

I've also tested this using Claude and AppleWin to work with Ultima I, so it works well!

